PR #1, opened 2019-12-03
Add a location icon component

evansiroky:
This PR is sort of a proof-of-concept for adding in a new package. This is done for one of the simplest components within otp-react-redux which merely renders an icon representing either a from or to location.

Also, I have added a few interesting items to storybook to help display a few more items.


.storybook/webpack.config.js
```diff
@@ -1,9 +1,15 @@
...
+
 module.exports = function({ config }) {
     config.module.rules.push({
       test: /.story.js$/,
       loaders: [require.resolve('@storybook/addon-storysource/loader')],
       enforce: 'pre',
     });
-  
+
+    config.resolve.alias = {
```
fpurcell: @c-harper says this construct might not be something we do long term, but good enough for now.

PR #2, opened 2019-12-03
Import trimet.org previous component work

landonreed:
Adds itinerary components.


landonreed:
Closing in favor of #7.

PR #4, opened 2019-12-05
Location field

evansiroky:
This PR contains all the necessary items to get the location field working.

See the IBI storybook deployment here: https://ibi-group.github.io/otp-ui/


packages/location-field/src/index.js
```diff
@@ -0,0 +1,831 @@
...
+}
+
+function createTransitStopOption(stop, onClick, isActive) {
+  return (
+    <Styled.MenuItem onClick={onClick} key={itemKey++} active={isActive}>
+      <div>
+        <div style={{ float: "left", paddingTop: "3px" }}>
+          <Bus size={13} />
+          <div style={{ fontSize: "8px" }}>
+            {humainzeDistanceStringImperial(stop.dist, true)}
```
landonreed: typo in method name `humainzeDistanceStringImperial`

packages/location-field/src/index.js
```diff
@@ -0,0 +1,831 @@
...
+  formatStoredPlaceName
+} from "@opentripplanner/core-utils/lib/map";
+import { isIE } from "@opentripplanner/core-utils/lib/ui";
+import getGeocoder from "@opentripplanner/geocoder";
+import { humainzeDistanceStringImperial } from "@opentripplanner/humanize-distance";
+import LocationIcon from "@opentripplanner/location-icon";
+
+import * as Styled from "./styled";
+
+// helper functions for dropdown options
```
landonreed: Since we're working in this micro-packages now, this might be a good opportunity to move these "helper functions" to a separate file in the package (similar to how the styled components are imported).

packages/location-field/src/index.js
```diff
@@ -0,0 +1,831 @@
...
+      locationType
+    } = this.props;
+    const location = currentPositionToLocation(currentPosition);
+    if (location) {
+      // If geolocation is successful (i.e., user has granted app geolocation
+      // permission and coords exist), set location.
+      onLocationSelected({ locationType, location });
+    } else {
+      // Call geolocation.getCurrentPosition and set as from/to locationType
+      getCurrentPosition(locationType);
```
landonreed: Does `getCurrentPosition` not take a `onLocationSelected` callback? See https://github.com/opentripplanner/otp-react-redux/compare/component-refactor#diff-b47b2554cff82708da15605b764f0e6bR296-R297

landonreed: Also, when I select 'Use Current Location in storybook: https://ibi-group.github.io/otp-ui/?path=/story/locationfield--locationfield-with-selected-location-in-mobile-context

I don't see the behavior function as it does in otp-rr (i.e., it doesn't seem to trigger a getCurrentPosition action). Perhaps this is because `currentPosition` is always provided? Please add a storybook entry for when currentPosition is not available.

packages/location-field/src/index.js
```diff
@@ -0,0 +1,831 @@
...
+  location: null,
+  nearbyStops: [],
+  onTextInputClick: null,
+  sessionSearches: [],
+  showClearButton: true,
+  showUserSettings: false,
+  static: false,
+  stopsIndex: null,
+  suppressNearby: false,
+  userLocationsAndRecentPlaces: []
```
landonreed: Perhaps we should remove some of these `defaultProps` values. Someone may try to use LocationField expecting it to work out of the box with no props defined for certain actions and then spend time trying to figure out which props they need to provide. Know what I mean?

packages/location-field/src/index.js
```diff
@@ -0,0 +1,831 @@
...
+    const target =
+      e.relatedTarget !== null ? e.relatedTarget : document.activeElement;
+    if (!location && (!target || target.getAttribute("role") !== "menuitem")) {
+      this.setState({ menuVisible: false, value: "", geocodedFeatures: [] });
+    }
+  };
+
+  onTextInputChange = evt => {
+    this.setState({ value: evt.target.value, menuVisible: true });
+    this.geocodeAutocomplete(evt.target.value);
```
landonreed: We should explicitly ask which of these low-level text input listeners AltSource needs access to.

packages/location-field/src/index.js
evansiroky: I kinda think it's better to turn off everything by default. If users want to be more advanced they can explore the other options as they see fit.

packages/location-field/src/index.js
evansiroky: We can ask. Or they can tell us. Or they can make a PR. :)

PR #5, opened 2019-12-12
From/To location picker

binh-dam-ibigroup:
This PR is for the From/To location selector found on map popups, stop viewer, etc.
image

Full storybook now at: https://ibi-group.github.io/otp-ui/?path=/story/from-to-picker--from-to


packages/from-to-location-picker/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@opentripplanner/from-to-location-picker",
+  "version": "0.0.1",
+  "description": "Location Picker Component",
+  "keywords": [
+    "sample",
+    "component"
+  ],
+  "author": "c-harper <charper@mapyours.com>",
```
evansiroky: Please change to your name @binh-dam-ibigroup.

packages/from-to-location-picker/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "homepage": "https://github.com/TriMetPDX/otp-ui#readme",
+  "license": "MIT",
+  "main": "lib/sample.js",
+  "module": "src/sample.js",
+  "peerDependencies": {
+    "icons": "^0.0.1",
```
evansiroky: What is this? I don't think it's needed.

packages/from-to-location-picker/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "react": "^16.8.6",
+    "react-dom": "^16.8.6",
+    "styled-components": "^4.4.1"
+  },
+  "publishConfig": {
+    "registry": "https://registry.yarnpkg.com"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/TriMetPDX/otp-ui.git"
```
evansiroky: Should be ...github.com/trimet-sandbox/otp-ui/

packages/from-to-location-picker/package.json
```diff
@@ -0,0 +1,37 @@
...
+  },
+  "scripts": {
+    "test": "echo \"Error: run tests from root\" && exit 1"
+  },
+  "bugs": {
+    "url": "https://github.com/TriMetPDX/otp-ui/issues"
```
evansiroky: Should be ...github.com/trimet-sandbox/otp-ui/

packages/from-to-location-picker/package.json
```diff
@@ -0,0 +1,37 @@
...
+  },
+  "dependencies": {
+    "@opentripplanner/location-icon": "0.0.1",
+    "@storybook/addon-actions": "^5.2.8"
```
evansiroky: Please remove, this is available in the root package.json

packages/from-to-location-picker/src/index.js
```diff
@@ -0,0 +1,59 @@
...
+  fromText = "From here",
+  toText = "To here",
+  showIcons = true,
+  onFromClick,
+  onToClick
+}) => {
+  return (
+    <FromToPickerSpan>
+      <LocationPickerSpan>
+        {showIcons ? <LocationIcon type="from" size={iconSize} /> : null}
```
evansiroky: This can be shortened to the following:

```js
{showIcons && <LocationIcon ... }
```

packages/from-to-location-picker/src/index.js
```diff
@@ -0,0 +1,59 @@
...
+        <Button onClick={onFromClick}>{fromText}</Button>
+      </LocationPickerSpan>
+      <LocationPickerSpan>
+        {showIcons ? <LocationIcon type="to" size={iconSize} /> : null}
```
evansiroky: This can be shortened to the following:

```js
{showIcons && <LocationIcon ... }
```

packages/from-to-location-picker/src/index.js
```diff
@@ -0,0 +1,59 @@
...
+
+import LocationIcon from "@opentripplanner/location-icon";
+import { LocationPickerSpan, Button, FromToPickerSpan } from "./styled";
+
+const iconSize = "0.9em";
+
+const FromToLocationPicker = ({
+  fromText = "From here",
+  toText = "To here",
+  showIcons = true,
```
evansiroky: These defaults should not be included here. The defaultProps are already sufficient.

packages/from-to-location-picker/mocks/trimet-mock.css
```diff
@@ -0,0 +1,22 @@
+.trimet-ambient {
```
evansiroky: We should determine if there is a different way to apply different themes and styling. If so, this file might not be needed.

packages/from-to-location-picker/package.json
binh-dam-ibigroup: Will change for now.

packages/from-to-location-picker/mocks/trimet-mock.css
binh-dam-ibigroup: This was resolved last week with the decision to use rebass.js (separate PR(s)).

fpurcell: Hey Binh,

Cole and I just talked, and he also mentioned that the .css file is a bit funky in how styled-components work.  Rather than doing things in a .css way with class names, he pointed out how the Pulse-Dot is done, and how one can override a default setting: https://github.com/opentripplanner/otp-ui/blob/master/packages/pulse-dot/src/index.js#L4  

Anyway, I'm going to accept the PR, and note that we'll work these things out later.

evansiroky:
Since everything in this branch is included in the branch for #6, I propose to close this PR.

PR #6, opened 2019-12-13
Base Map Component

binh-dam-ibigroup:
This branch has the code for the basic map functionality.


packages/base-map/src/index.js
```diff
@@ -0,0 +1,248 @@
...
+    lmap.on("singleclick", this.onLeftClick);
+  }
+
+  componentDidUpdate() {}
+
+  // remove custom overlays on unmount
+  // TODO: Is this needed? It may have something to do with mobile vs desktop views
+  componentWillUnmount() {
+    const lmap = this.refs.map.leafletElement;
+    lmap.eachLayer(lmap.removeLayer);
```
fpurcell: Hey Binh, 

I'm getting a "this is undefined" error here.  Stepping into Leaflet's eachLayer -> removeLayer functions, I can see that the 'this' in removeLayer is indeed null (see image).

Screen Shot 2019-12-15 at 1 31 12 PM

BTW, yum install pulled in the following:
Leaflet 1.5.1 (2019-05-08)
React-Leaflet v2.6.0 (2019-11-18) (note that the changelog mentions "- Updated Leaflet dependency to v1.6.").

Take care,
Frank

binh-dam-ibigroup: Hi Frank, I reverted an inlined function call (see my latest commit). Let me know if this error is still happening on your side.

packages/base-map/mocks/AllVehicles.js
```diff
@@ -0,0 +1,48 @@
+import React from "react";
```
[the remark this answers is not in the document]
binh-dam-ibigroup: The mocks directory was renamed in a commit today per prior discussion with @evansiroky.

packages/base-map/src/index.story.js
```diff
@@ -0,0 +1,128 @@
...
+  <BaseMap
+    mapConfig={mapConfig}
+    popupLocation={center}
+    popupContent={samplePopup}
+    onPopupClosed={onPopupClosed}
+  />
+);
+
+export const customLocationPopupNoContent = () => (
+  <BaseMap mapConfig={mapConfig} popupLocation={center} />
```
evansiroky: This story does not appear to demonstrate anything that the `customLocationPopupContent` doesn't do.  Might be worth removing to avoid confusion.

packages/base-map/src/index.story.js
```diff
@@ -0,0 +1,128 @@
...
+);
+
+export const customPopupContentNoLocation = () => (
+  <BaseMap mapConfig={mapConfig} popupContent={samplePopup} />
```
evansiroky: This story does not appear to demonstrate anything that the `customLocationPopupContent` doesn't do.  Might be worth removing to avoid confusion.

packages/base-map/src/index.js
```diff
@@ -0,0 +1,259 @@
...
+            {popupContent}
+          </Popup>
+        )}
+      </Map>
+    );
+  }
+}
+
+BaseMap.propTypes = {
+  children: PropTypes.oneOfType([
```
evansiroky: Needs documentation of how to define a layer. Also, are these propTypes right?

packages/base-map/src/index.js
```diff
@@ -0,0 +1,259 @@
...
+   * The coordinates of the popup object to display, as a [lat, lng] array.
+   */
+  popupLocation: PropTypes.arrayOf(PropTypes.number)
+};
+
+BaseMap.defaultProps = {
+  children: null,
+  mapConfig: {
+    initLat: 45.52,
+    initLon: -122.682,
```
evansiroky: The initial lat/lon should probably not be defaults. I'm somewhat inclined to not have any defaults (except maybe for the osm tile layer) to keep this to a bare minimum.

packages/base-map/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@opentripplanner/base-map",
+  "version": "0.0.1",
+  "description": "Base Map Component",
+  "author": "@binh-dam-ibigroup",
+  "homepage": "https://github.com/opentripplanner/otp-ui/#readme",
+  "license": "MIT",
+  "main": "index.js",
```
evansiroky: This should be `lib/index.js`.

packages/base-map/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "module": "src/index.js",
```
evansiroky: These module entries should be removed or renamed to `lib/index.js`.

packages/base-map/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "private": false,
+  "dependencies": {
+    "leaflet": "^1.6.0",
+    "react-leaflet": "^2.6.0"
```
evansiroky: Needs an entry for `prop-types`.

packages/base-map/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/opentripplanner/otp-ui.git"
+  },
+  "scripts": {
+    "test": "echo \"Error: run tests from root\" && exit 1"
+  },
+  "bugs": {
+    "url": "https://github.com/opentripplanner/otp-ui/issues"
+  }
```
evansiroky: This needs a peerDependencies entry for `react`.

packages/from-to-location-picker/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/opentripplanner/otp-ui.git"
+  },
+  "scripts": {
+    "test": "echo \"Error: run tests from root\" && exit 1"
+  },
+  "bugs": {
+    "url": "https://github.com/opentripplanner/otp-ui/issues"
+  }
```
evansiroky: This needs a peerDependencies entry for `react` and `styled-components`.

packages/from-to-location-picker/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "version": "0.0.1",
+  "description": "Location Picker Component",
+  "author": "@binh-dam-ibigroup",
+  "homepage": "https://github.com/opentripplanner/otp-ui/#readme",
+  "license": "MIT",
+  "main": "lib/index.js",
+  "module": "src/index.js",
+  "private": false,
+  "dependencies": {
+    "@opentripplanner/location-icon": "0.0.1"
```
evansiroky: This should be updated to `^0.0.2`. Also, an entry for `prop-types` needs to be added here.

packages/base-map/src/index.js
landonreed: I wonder if we should not include default values for lat/lng/zoom and instead make these required props. @evansiroky?

packages/base-map/src/index.story.js
binh-dam-ibigroup: I included this story and the next, so that if you pass the content but omit the location, or vice-versa, the popup is not shown at all.

packages/base-map/src/index.js
binh-dam-ibigroup: The children prop types do not result in validation errors at least. That takes care of zero, one, or multiple children. The overlays and map contents seem to resolve to type Object as opposed to a more precise type.

packages/base-map/src/index.js
binh-dam-ibigroup: So, make it required?

packages/from-to-location-picker/package.json
binh-dam-ibigroup: Does this require merging with add-itineary-eas?

packages/base-map/assets/map.css
```diff
@@ -0,0 +1,5 @@
+@import url(https://unpkg.com/leaflet@1.0.3/dist/leaflet.css);
```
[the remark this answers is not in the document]
binh-dam-ibigroup: It is the leaflet css version that is used in otp-react-redux.

packages/base-map/src/index.story.js
```diff
@@ -0,0 +1,128 @@
...
+);
```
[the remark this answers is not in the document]
binh-dam-ibigroup: These two demonstrate that is you omit the popup location or the popup content, then the popup should not be displayed at all.

packages/base-map/src/index.js
binh-dam-ibigroup: initZoom should remain optional.

packages/base-map/src/index.js
```diff
@@ -0,0 +1,259 @@
...
+    const userControlledOverlays = [];
+    const fixedOverlays = [];
+    React.Children.toArray(children).forEach(child => {
+      if (child.props.name) userControlledOverlays.push(child);
+      else fixedOverlays.push(child);
+    });
+
+    const center = mapConfig &&
+      mapConfig.initLat &&
+      mapConfig.initLon && [mapConfig.initLat, mapConfig.initLon];
```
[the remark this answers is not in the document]
binh-dam-ibigroup: @evansiroky and @fpurcell, how would you like to handle that?

packages/base-map/src/index.js
```diff
@@ -0,0 +1,259 @@
...
+
+    return (
```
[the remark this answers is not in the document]
binh-dam-ibigroup: I disagree, but if others feel otherwise I will add a mention to that code.

packages/base-map/src/index.js
fpurcell: Agree w/ Landon about lat & lon being required.  Could be strange for an Atlanta map to zoom into Portland by default.  Zoom should have a default value.

packages/base-map/src/index.js
```diff
@@ -0,0 +1,259 @@
...
+
+  clearSingleClickTimeout() {
+    if (this.singleClickTimeout !== null) {
+      clearTimeout(this.singleClickTimeout);
+      this.singleClickTimeout = null;
+    }
+  }
+});
+
+class BaseMap extends Component {
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Aren't we were putting the documentation to Storybook instead?

packages/base-map/src/index.js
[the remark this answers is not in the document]
fpurcell: Not sure of the implementation, Binh ... but I like where @landonreed  is going with having a center (or lat lon) prop that can then be updated in an outside state controller, and thus move the map around. 

FWIW, @c-harper  mentioned a few times in looking at otp-rr, that he'd thought having the components pull data directly from the config object (as opposed to using individual props) was an area ripe for refactoring.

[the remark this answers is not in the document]
binh-dam-ibigroup: How about writing:
<BaseMap center={[lat, lon]} zoom={13} ...> 
and let center and the optional zoom prop be managed  by an external state controller?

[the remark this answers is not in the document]
landonreed: Yes, my thought is basically we break out everything in mapConfig into their own BaseMap props and remove `init` from the name of those props (lat, lon can become a single `center` prop with an array as you've described). I'm OK with making those controlled. We can update otp-rr to derive those values from the application state (I don't think that should be too big of a lift).

packages/base-map/src/index.js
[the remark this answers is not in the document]
landonreed: @evansiroky, want to break the tie?

packages/from-to-location-picker/package.json
evansiroky: It appears that that branch has been merged into master and there are now conflicts. So merging master into this branch should take care of all that.

packages/base-map/src/index.js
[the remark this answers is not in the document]
evansiroky: I think it's good practice to always have jsdoc for class comments to describe the basics of what the class is supposed to do.

packages/base-map/src/index.js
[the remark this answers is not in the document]
evansiroky: I agree with @binh-dam-ibigroup. Seems like that should be a separate component that should be either in the transitive overlay or a separate component altogether.

packages/base-map/src/index.js
```diff
@@ -0,0 +1,284 @@
...
+          <Popup position={popupLocation} onClose={onPopupClosed}>
+            {popupContent}
+          </Popup>
+        )}
+      </Map>
+    );
+  }
+}
+
+const twoNumberArrayChecker = (
```
evansiroky: Since we're going all out with a custom PropType here, we should also include checks for whether the values are valid GPS coordinates. IE -90 <= lat <= 90 and -180 <= lon <= 180.

binh-dam-ibigroup: Good point

binh-dam-ibigroup: Might have gone too far in the latest commit with the props validation... Let me know if you want to go back.

packages/base-map/package.json
```diff
@@ -10,7 +10,8 @@
   "dependencies": {
     "leaflet": "^1.6.0",
     "react-leaflet": "^2.6.0",
-    "prop-types": "^15.7.2"
+    "prop-types": "^15.7.2",
+    "core-utils": "0.0.1"
```
evansiroky: Pretty sure this should be `@opentripplanner/core-utils`.

packages/base-map/src/index.js
```diff
@@ -250,7 +233,7 @@ BaseMap.propTypes = {
   /**
    * The coordinates of the popup object to display, as a [lat, lng] array.
    */
-  popupLocation: PropTypes.arrayOf(twoNumberArrayChecker),
+  popupLocation: LatlngPropType,
```
evansiroky: This looks very clean. Well done.

packages/base-map/src/index.js
evansiroky: I like it. One thing tho, I'd move the props validation into this new file I created: `core-utils/src/types.js`.

binh-dam-ibigroup: Should I merge with the itinerary branch to get that file?

packages/from-to-location-picker/package.json
binh-dam-ibigroup: I merged with add-itinerary-eas which has everyone's most recent files and versions.

packages/base-map/src/index.js
binh-dam-ibigroup: I merged with add-itinerary-eas to get the most current outside of master.

packages/base-map/src/index.js
```diff
@@ -0,0 +1,271 @@
...
+    const userControlledOverlays = [];
+    const fixedOverlays = [];
+    React.Children.toArray(children).forEach(child => {
+      if (child.props.name) userControlledOverlays.push(child);
+      else fixedOverlays.push(child);
+    });
+
+    return (
+      <Map
+        ref="map"
```
landonreed: @binh-dam-ibigroup, were you running into issues with the use of `this.refs.map`? I believe the latest react-leaflet docs say that we should use `this.leafletElement` to access the map, so we can probably remove this `ref`. https://react-leaflet.js.org/docs/en/components

packages/base-map/src/index.js
```diff
@@ -0,0 +1,271 @@
...
+    }
+  }
+});
+
+/**
+ * The base OpenTripPlanner map on which everything else is rendered.
+ */
+class BaseMap extends Component {
+  componentDidMount() {
+    const lmap = this.refs.map.leafletElement;
```
landonreed: See https://github.com/opentripplanner/otp-ui/pull/6/files#r359934885 about replacing `this.refs.map.leafletElement` with `this.leafletElement`.

packages/base-map/src/index.js
```diff
@@ -0,0 +1,271 @@
...
+    lmap.options.singleClickTimeout = 250;
+    lmap.on("singleclick", this.onLeftClick);
+  }
+
+  componentDidUpdate() {}
+
+  // remove custom overlays on unmount
+  // TODO: Is this needed? It may have something to do with mobile vs desktop views
+  componentWillUnmount() {
+    const lmap = this.refs.map.leafletElement;
```
landonreed: See https://github.com/opentripplanner/otp-ui/pull/6/files#r359934885 about replacing `this.refs.map.leafletElement` with `this.leafletElement`.

packages/base-map/src/index.js
```diff
@@ -0,0 +1,271 @@
...
+                  {L.Browser.retina && layer.hasRetinaSupport ? (
+                    <TileLayer
+                      url={layer.url}
+                      attribution={layer.attribution}
+                      maxZoom={layer.maxZoom}
+                      tileSize={512}
+                      zoomOffset={-1}
+                      detectRetina
+                    />
+                  ) : (
```
landonreed: I prefer for multi-line ternaries to be styled like:
```
{L.Browser.retina && layer.hasRetinaSupport
  ? x
  : y
}
```
If what you have is what airbnb specifies, it's fine for now.

packages/base-map/src/index.story.js
landonreed: @evansiroky and I both came to the determination that these final two stories are confusing. Please remove them and perhaps just update the docs for the props to indicate that `popupContent` and `popupLocation` are both needed in order to function. Or perhaps it is more clear to simply combine them into a single prop called `popup` that looks like:
```
popup: {
  location: latlngType,
  content: PropTypes.node
}
```

binh-dam-ibigroup: I like this suggestion.

packages/base-map/src/index.js
binh-dam-ibigroup: The original format is enforced by es-lint prior to commit.

packages/base-map/src/index.js
binh-dam-ibigroup: I will make the change; we will have to see how nice it plays outside of Storybook.

binh-dam-ibigroup: No, these lines don't work with the suggested changes:
`    const lmap = this.leafletElement; // this.refs.map.leafletElement;
    lmap.options.singleClickTimeout = 250;
`
and 
`    const lmap = this.leafletElement; // this.refs.map.leafletElement;
    lmap.eachLayer(layer => { ...

`

landonreed:
I'm going to go ahead and merge to make reviews of other PRs a little easier (e.g., #11). Nice work, @binh-dam-ibigroup!

PR #7, opened 2019-12-17
Add itinerary component

evansiroky:
This PR adds the ability to render itineraries returned from the IBI Fork of OpenTripPlanner.

To accomplish this, the following was added:

### The Icons package

All of the icons produced for the trimet.org project were added here in this package. Also, some custom functions for rendering the proper icon for a mode or leg were added.

### The Itinerary-Body package

This package allows the ability to render itineraries returned from the IBI Fork of OpenTripPlanner. The work so far is a blend of some work for the trimet.org and then other stuff copied and modified from otp-react-redux for displaying the leg details. A big difference in the design of the itinerary component for trimet.org and in otp-react-redux is the display of the line icon on the left side. In otp-react-redux, the design mimicks a typical transit diagram with nodes and lines representing a stops on a transit line. This scheme is also reflected in almost the exact same way on the map. The trimet.org rendering moves the mode icon into the node of the transit line. I went ahead and followed this scheme, but perhaps some items from otp-react-redux such as the interlined stop and intermediate stop stylings may not make much sense anymore.

Also, as of this time, the PropTypes are not comprehensive in type definition and documentation. I hope to revisit this after catching up on the other PRs and work that's been done for this project.

### Removal of pulse-dot package

Might as well do it now.


packages/itinerary-body/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@opentripplanner/itinerary-body",
+  "version": "0.0.1",
+  "description": "A component for displaying an itinerary body of a trip planning result",
+  "keywords": [
+    "otp",
+    "itinerary"
+  ],
+  "author": "Evan Siroky",
+  "homepage": "https://github.com/TriMetPDX/otp-ui#readme",
```
binh-dam-ibigroup: You will have to update to opentripplanner/otp-ui.

packages/itinerary-body/src/AccessLegBody/index.js
```diff
@@ -0,0 +1,302 @@
...
+  timeOptions: timeOptionsType.isRequired
+};
+
+AccessLegBody.defaultProps = {
+  diagramVisible: null,
+  followsTransit: false,
+  showElevationProfile: false
+};
+
+function TNCLeg({
```
binh-dam-ibigroup: Should this class be in its own file? (same question for AccessLegSummary)

packages/itinerary-body/src/TransitLegBody/index.js
```diff
@@ -0,0 +1,259 @@
...
+                  src={logoUrl}
+                  height={25}
+                  style={{ marginLeft: "5px" }}
+                />
+              )}
+            </a>
+          </Styled.AgencyInfo>
+        }
+
+        {/* Alerts toggle */}
```
binh-dam-ibigroup: For a future iteration, extracting alerts to their own components.

packages/itinerary-body/src/place-row.js
```diff
@@ -0,0 +1,291 @@
...
+                routingType={routingType}
+                setActiveLeg={setActiveLeg}
+                setLegDiagram={setLegDiagram}
+                showElevationProfile={showElevationProfile}
+                timeOptions={timeOptions}
+              />
+            ))}
+        </Styled.PlaceDetails>
+      </Styled.DetailsColumn>
+      <Styled.MapButtonColumn hideBorder={hideBorder.toString()}>
```
binh-dam-ibigroup: Is this 
image
 supposed to reproduce this portion 
image
 of the UI?

packages/icons/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "lib"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/opentripplanner/otp-ui.git"
+  },
+  "bugs": {
+    "url": "https://github.com/opentripplanner/otp-ui/issues"
+  },
+  "dependencies": {
```
c-harper: I think core-utils needs to be a peer dependency or else its going to get bundled into each of the individual components. Right as each component ie core-utils is transpiled into es5 I don't think that it efficiently is bundling just the "leg" type with icons for example. I might be wrong about this though, up for discussion.

package.json
```diff
@@ -55,7 +55,7 @@
   },
   "scripts": {
     "bootstrap": "lerna bootstrap --use-workspaces",
-    "build": "lerna exec --parallel -- babel --root-mode upward src -d lib --ignore **/*.story.js,**/*.spec.js",
+    "prepublish": "lerna exec --parallel -- babel --root-mode upward src -d lib --ignore **/*.story.js,**/*.spec.js",
```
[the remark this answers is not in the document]
evansiroky: I actually like this because this is triggered every time `yarn` is ran. As a result, one can be sure that all the packages have been built after installing the node_modules in the development environment.

packages/icons/package.json
evansiroky: There are currently a lot of helper functions in the core-utils package that this package depends on in order to work. These were extracted from the utils in otp-react-redux, so not a lot of thought or analysis was done to see if these utils are needed just for this package or are needed for other ones as well.

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "root",
+  "name": "@opentripplanner/kitchen-sink",
```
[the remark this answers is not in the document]
evansiroky: Heh, I believe @landonreed did this one. I'll rename it to `@opentripplanner/otp-ui`.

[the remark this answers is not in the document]
c-harper: I don't think there are plans to publish the monorepo as an npm module based on other monorepos I have seen it should just be "otp-ui"

packages/itinerary-body/src/AccessLegBody/index.js
evansiroky: In the past, we haven't been too strict about having one component per file, but maybe it's more readable to do so.

packages/icons/package.json
c-harper: Fair enough, but I think it should be a peer dependency shouldn't it?

packages/itinerary-body/src/place-row.js
evansiroky: The trimet.org mockup moved some things around. I am not exactly sure what the intent is with putting the StopID and kind of omitting the place sub-header is some areas, but I left it as is. I assume this will need more iteration to get it looked a little nicer.

packages/itinerary-body/src/styled.js
```diff
@@ -0,0 +1,190 @@
...
+  /* special messaging, not sure yet */
+`;
+
+export const StopIdSpan = styled.span`
+  font-weight: 200;
+  font-size: 0.9em;
+  margin-left: 10px;
+`;
+
+export const MapButton = styled(ClearButton)`
```
c-harper: Do we need to extend ClearButton here? Can't we just rename ClearButton to MapButton and set these styles on it?

packages/itinerary-body/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "registry": "https://registry.yarnpkg.com"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/opentripplanner/otp-ui.git"
+  },
+  "bugs": {
+    "url": "https://github.com/opentripplanner/otp-ui/issues"
+  },
+  "dependencies": {
```
c-harper: Lets figure out which ones should be peer dependencies

evansiroky:
I just added a few commits that implement requested changes in the package.json files. Also, I removed the pulse-dot package.

packages/itinerary-body/src/place-row.js
evansiroky: I don't quite know what it is supposed to do. It just accepts a function and whatever is done as a result of calling that function is out of the scope of this component.

PR #17, opened 2020-01-06
Route viewer overlay

evansiroky:
This package allows the display of a route overlay on the base map.


packages/route-viewer-overlay/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/opentripplanner/otp-ui/issues"
+  },
+  "homepage": "https://github.com/opentripplanner/otp-ui#readme",
+  "dependencies": {
+    "@mapbox/polyline": "^1.1.0",
+    "@opentripplanner/core-utils": "^0.0.8",
+    "prop-types": "^15.7.2",
+    "react-leaflet": "^2.6.1"
```
fpurcell: shouldn't prop-types and react-leaflet be peerDependencies, and base-map be a direct dependency?

evansiroky:
> Any issues working with BaseMap so far?

@binh-dam-ibigroup, it's working just fine.

packages/route-viewer-overlay/package.json
evansiroky: I think being too reliant on other packages to include these is unnecessary. I've only been doing this for packages that require one and only one version of a package to be installed. Otherwise, this should work fine.

PR #20, opened 2020-01-07
Settings selector

binh-dam-ibigroup:
This pull request covers the controls under the "Trip Options & Preferences" tab in the OTP UI, including:

* Mode and submode selection,
* Company/provider selection
* Trip optimization parameters


binh-dam-ibigroup:
I'm good with renaming the top package.
For individual components, how about I combine the stories for the UI sub-components?

> Hey Bihn,
> 
> 2 thoughts:
> a) can we rename the top level directory from settings-selector to trip-form, or trip-form-components?
> b) should we have 1 storybook as opposed to multiple stories? (debate can be made either way...I like that all the separate components are under a single package/ , but no loving the multiple stories).

fpurcell:
+1

binh-dam-ibigroup:
@fpurcell It is the name that was used in the old code...

> This is great, Binh. I like the rename and the storybook layout and the functionality of the SettingsSelectorPanel (not crazy about the name "SettingsSelector", but that's a minor quibble since the code is under the trip-form bucket).

packages/trip-form/src/ModeIcon/index.js
```diff
@@ -0,0 +1,65 @@
+import React from "react";
+import PropTypes from "prop-types";
+import * as Icons from "@opentripplanner/icons";
+
+/**
+ * A generic icon component that displays an icon for the specified transportation mode.
+ */
```
landonreed: Is this duplicative of https://github.com/opentripplanner/otp-ui/blob/master/packages/icons/src/trimet-mode-icon.js?

packages/trip-form/src/ModeSelector/index.js
```diff
@@ -0,0 +1,75 @@
...
+
+  const handleClick = option => {
+    if (!option.selected && typeof onChange === "function") {
+      onChange(option.id);
+    }
+  };
+
+  const makeButton = option => (
+    <ModeButton
+      /* If key is missing, a warning in React Dev Tools is triggered. */
```
landonreed: This comment probably isn't needed. That's a relatively well known prop/warning.

packages/trip-form/src/SubmodeSelector/index.js
```diff
@@ -0,0 +1,76 @@
...
+  const RowType = inline ? InlineSubmodeRow : SubmodeRow;
+
+  return (
+    <SettingsSection className={className} style={style}>
+      {label && <LabelType>{label}</LabelType>}
+      <RowType>
+        {modes &&
+          modes.map(option => (
+            <ModeButton
+              /* If key is missing, a warning in React Dev Tools is triggered. */
```
landonreed: Again, comment probably not needed

packages/trip-form/src/ModeIcon/index.js
binh-dam-ibigroup: I didn't see that one... the original code I based this work was referring to this file instead: https://github.com/opentripplanner/otp-react-redux/blob/dev/lib/components/icons/mode-icon.js

If creating an issue, is it against this repo or against otp-react-redux?

landonreed: Probably this repo is OK.

packages/trip-form/src/GeneralSettingsPanel/index.js
```diff
@@ -0,0 +1,107 @@
...
+import queryParams from "@opentripplanner/core-utils/lib/query-params";
+import {
+  defaultParams,
+  getQueryParamProperty
+} from "@opentripplanner/core-utils/lib/query";
+import { queryType } from "@opentripplanner/core-utils/lib/types";
+
+import CheckboxSelector from "../CheckboxSelector";
+import DropdownSelector from "../DropdownSelector";
+import commonModes from "../__mocks__/modes"; // FIXME: Replace with ref to configuration.
```
landonreed: I don't think we should be importing from `__mocks__` in the files that make it to the library (i.e., not test or story files). My recommendation would be to make this a constant that is imported from a `consts` file and set this as a default value for a new prop `modes`. I'm not crazy about the `configWrapper` below, especially if all that's needed is the modes field.

binh-dam-ibigroup: Oh good catch, I forgot about getting rid of this one.

packages/trip-form/src/__mocks__/modes.js
```diff
@@ -0,0 +1,69 @@
...
+      mode: "RAIL",
+      label: "WES"
+    },
+    {
+      mode: "GONDOLA",
+      label: "Aerial Tram"
+    }
+  ],
+  accessModes: [
+    // Biketown and Uber should be displayed the same despite Biketown company set in companies.
```
landonreed: This comment reads a bit cryptic, but seems important. Could you elaborate here?

packages/core-utils/src/state.js
```diff
@@ -0,0 +1,151 @@
+import isEqual from "lodash.isequal";
```
landonreed: Is this `state.js` file used anywhere?

binh-dam-ibigroup: Yes, query.js has this:
`import { getActiveSearch } from "./state";`

packages/trip-form/src/__mocks__/modes.js
binh-dam-ibigroup: Yep. Fixed in lastest commit.

packages/core-utils/src/state.js
landonreed: Oh, I was just confused because I didn't see any imports in this PR.

PR #19, opened 2020-01-07
Trip viewer overlay

evansiroky:
This component renders a trip on the base map.

This PR depends on #17.

PR #21, opened 2020-01-10
Transitive and endpoints overlays

evansiroky:
This PR includes both the transitive and endpoints overlays. They are both included because it doesn't make a whole lot of sense to display one without the other.

This PR depends on #17 and #19.

### DIFF
--- a/packages/transitive-overlay/src/index.js
+++ b/packages/transitive-overlay/src/index.js
@@ -1,7 +1,6 @@
 import L from "leaflet";
 import isEqual from "lodash.isequal";
-import { encodedPolylineType } from "@opentripplanner/core-utils/lib/types";
-import PropTypes from "prop-types";
+import { transitiveDataType } from "@opentripplanner/core-utils/lib/types";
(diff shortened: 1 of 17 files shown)


landonreed:
Can you rebase this off of the latest master so that it's easier to read, @evansiroky?

packages/transitive-overlay/src/index.js
```diff
@@ -0,0 +1,240 @@
...
+});
+
+TransitiveCanvasOverlay.propTypes = {
+  /**
+   * The transitiveData object is assumed to be the result of converting an
+   * OpenTripPlanner itinerary result into a transitive-readable format. This is
+   * typically done using the @opentripplanner/core-utils/map#itineraryToTransitive
+   * function.
+   */
+  transitiveData: PropTypes.shape({
```
binh-dam-ibigroup: These should probably be moved to types.js in core-utils.

evansiroky:
I merged master into this branch and move the transitive prop type definition to core-utils. Should be ready for review again.

PR #22, opened 2020-01-10
Add stop viewer overlay

evansiroky:
This PR adds the stop viewer overlay. It basically just shows a highlighted stop.


packages/stop-viewer-overlay/src/index.js
```diff
@@ -0,0 +1,71 @@
...
+import React from "react";
+import {
+  FeatureGroup,
+  MapLayer,
+  Popup,
+  CircleMarker,
+  withLeaflet
+} from "react-leaflet";
+
+class StopViewerOverlay extends MapLayer {
```
landonreed: I agree with @fpurcell's comment about usage of this component. It would be good to add JSdoc regarding how this works.

packages/stop-viewer-overlay/src/index.js
```diff
@@ -0,0 +1,71 @@
...
+    return (
+      <FeatureGroup>
+        <CircleMarker
+          key={stopData.id}
+          center={[stopData.lat, stopData.lon]}
+          radius={9}
+          fillOpacity={1}
+          fillColor="cyan"
+          color="#000"
+          weight={3}
```
landonreed: Re: @fpurcell's comment, might be good to have `markerProps` similar to how you added `polylineProps` (or whatever they were called) on the trip/route viewer overlays

binh-dam-ibigroup:
Is the intent to generate random stops each time you pan the map?

evansiroky:
I just improved the leaflet path styling and added jsDoc. This change applies to more than just this stop viewer overlay now.

evansiroky:
Ok, just one more change: I got rid of the randomized mock stops for the stops overlay.

PR #23, opened 2020-01-11
Add vehicle rental overlay

evansiroky:
This adds the ability to display rental vehicles as an overlay on the map.


packages/vehicle-rental-overlay/src/VehicleRentalOverlay.story.js
```diff
@@ -0,0 +1,259 @@
...
+];
+const randomBikeRentalStations = Array(20)
+  .fill(null)
+  .map(makeRandomBikeRentalStation);
+const randomCarRentalStations = Array(20)
+  .fill(null)
+  .map(makeRandomCarRentalStation);
+const randomEScooterStations = Array(20)
+  .fill(null)
+  .map(makeRandomEScooterRentalStation);
```
landonreed: I'd like to see a lot of this mock data perhaps live in a `__mocks__` folder for this package. Isn't that the pattern we've been generally following?

packages/vehicle-rental-overlay/src/styled.js
```diff
@@ -0,0 +1,62 @@
...
+const BaseBikeRentalIcon = styled.div`
+  background-size: contain;
+  background-position: center;
+  background-repeat: no-repeat;
+  margin: auto;
+  width: 24px;
+  height: 24px;
+`;
+
+export const hubIcons = [
```
landonreed: Maybe this could use a comment describing what the icon is.

packages/vehicle-rental-overlay/src/index.js
```diff
@@ -0,0 +1,251 @@
...
+    className: "",
+    iconSize: [11, 16],
+    popupAnchor: [0, -6],
+    html: ReactDOMServer.renderToStaticMarkup(
+      <Styled.StationMarker color={color} />
+    )
+  })
+);
+
+class VehicleRentalOverlay extends MapLayer {
```
landonreed: Definitely deserves JSdoc given that it serves as the overlay for a number of different modes.

packages/vehicle-rental-overlay/src/index.js
```diff
@@ -0,0 +1,251 @@
...
+  createLeafletElement() {}
+
+  updateLeafletElement() {}
+
+  startRefreshing() {
+    // ititial station retrieval
```
landonreed: typo: ititial

packages/vehicle-rental-overlay/src/index.js
```diff
@@ -0,0 +1,251 @@
...
+      configCompanies
+    );
+    let stationName = station.name || station.id;
+    if (station.isFloatingBike) {
+      stationName = `Free-floating bike: ${stationName}`;
+    } else if (station.isFloatingCar) {
+      stationName = `${stationNetworks} ${stationName}`;
+    } else if (station.isFloatingVehicle) {
+      // assumes that all floating vehicles are E-scooters
+      stationName = `${stationNetworks} E-scooter`;
```
landonreed: Since we're touching this code, it might be worth extracting the `E-scooter` string into a prop that can be overridden if need be.

packages/vehicle-rental-overlay/src/index.js
```diff
@@ -0,0 +1,251 @@
...
+    let filteredStations = stations;
+    if (companies) {
+      filteredStations = stations.filter(
+        station =>
+          station.networks.filter(value => companies.includes(value)).length > 0
+      );
+    }
+
+    if (!filteredStations || filteredStations.length === 0)
+      return <FeatureGroup />;
```
landonreed: Does the airbnb style guide really recommend not encapsulating multi-line conditionals in curly braces? Not sure I like that at all.

binh-dam-ibigroup: On this style rule, it seems more like when you write a long 
`if (condition) thenStatement();` line, then es-lint recommends to put the thenStatement() into its own line. In that case would also add braces around thenStatement() to avoid confusion.

packages/core-utils/src/types.js
```diff
@@ -20,23 +27,65 @@ export const languageConfigType = PropTypes.shape({
...
+    pixels: PropTypes.number,
+    type: PropTypes.string.isRequired
+  }).isRequired
+);
+
 /**
- * Represents the expected configuration of the webapp
+ * Represents the expected configuration of the webapp.
+ *
+ * Note: this is an incomplete type mapping.
```
fpurcell: What all is missing? ... maybe list here or the line comment below.

packages/vehicle-rental-overlay/src/VehicleRentalOverlay.story.js
```diff
@@ -0,0 +1,259 @@
...
+    ...makeBaseRandomRentalStation({ id, networks }),
+    bikesAvailable: 1,
+    isFloatingBike: true,
+    spacesAvailable: 0
+  };
+}
+
+function makeRandomCarRentalStation(mapVal, idx) {
+  // make a free-floating style car rental station
+  const id = `Bike ${idx}`;
```
fpurcell: Minor: storybook popup says car2go BIKE :-)

packages/core-utils/src/types.js
```diff
@@ -20,23 +27,65 @@ export const languageConfigType = PropTypes.shape({
   stopViewer: PropTypes.string
 });
 
+export const mapSymbolsType = PropTypes.arrayOf(
```
binh-dam-ibigroup: This name seems too generic and IMO should instead suggest why the dockStrokeColor is a relevant attribute.

packages/core-utils/src/types.js
```diff
@@ -243,6 +306,25 @@ export const transitIndexStopWithRoutes = PropTypes.shape({
   )
 });
 
+/**
+ * This models data about vehicle rental stations as obtained from various
+ * vehicle rental API endpoints from OTP.
+ */
+export const stationsType = PropTypes.arrayOf(
```
binh-dam-ibigroup: Did you mean `stationType`? (It seems to be about one station as opposed to multiple.)

packages/from-to-location-picker/src/index.js
```diff
@@ -1,59 +1,100 @@
...
-  );
-};
+class FromToLocationPicker extends Component {
+  onFromClick = () => {
+    const { location, onFromClick, setLocation } = this.props;
+    if (onFromClick) {
+      onFromClick();
+      return;
+    }
+    setLocation({
```
binh-dam-ibigroup: Does the code that handles setLocation split into two different paths for 'from' and 'to'?

packages/vehicle-rental-overlay/src/VehicleRentalOverlay.story.js
fpurcell: Honestly, I would rather we get away from generating random data, and used canned data for the Storybooks (unless there's a storybook that absolutely needs a current datestamp on the mock data), @evansiroky.

Three reasons:
  a) when folks see a stop or vehicle in the Willamette (on storybook), they wonder if the app is broken (as @c-harper can attest :-).
  b) the format of the mock'd data might not fully reflect a snapshot of the real data the component might be fetching in the future  ... so having a static __mocks__/blah.json file pulled from the service, and rendering that seems to get one closer to the eventual truth
  c) I've heard that some recent changes to Storybook makes for a good unit testing framework (via a plugin or something -- haven't found out the how yet), and was thinking that if that does exist, it would probably need static data.

packages/core-utils/src/types.js
evansiroky: No, it's specifically for multiple as indicated by the arrayOf propType.

packages/from-to-location-picker/src/index.js
evansiroky: This is somewhat of an artifact of extracting stuff from otp-react-redux. Here is the relevant action: https://github.com/opentripplanner/otp-react-redux/blob/906d9226553cbbdd83cf455986f199c5a6964f72/lib/actions/map.js#L38-L63

packages/core-utils/src/types.js
evansiroky: There are probably too many things missing and the config itself is sometimes a moving target that may not be worth the time to make a detailed inventory. Ideally this type shouldn't even be used directly in these components that we're extracting. Only parts of the config should be used in each component.

evansiroky:
All right, I just did some refactoring to address all of these comments as follows:

- Used a mocked static list of vehicle rentals
- Added more comments
- Fixed a typo
- Made a slot for a custom vehicle station name function
- Renamed the mapSymbols type to a more meaningful name
- Extracted each styled hub icon styled component to allow for further individualized customization
- Made a story demonstrating use of the custom vehicle station name function

packages/core-utils/src/types.js
binh-dam-ibigroup: Oh I missed the inner array. Then do you want to do a type for one station, and have the stations props to be PropTypes.arrayOf(stationType)?

evansiroky:
@binh-dam-ibigroup, I just updated the station type as you requested.

binh-dam-ibigroup:
Your call if you want to wait for Landon or not.

PR #24, opened 2020-01-11
Add park and ride overlay

evansiroky:
This adds the park and ride map overlay component.

This PR depends on #23.


packages/vehicle-rental-overlay/src/VehicleRentalOverlay.story.js
```diff
@@ -0,0 +1,259 @@
...
+  .map(makeRandomBikeRentalStation);
+const randomCarRentalStations = Array(20)
+  .fill(null)
+  .map(makeRandomCarRentalStation);
+const randomEScooterStations = Array(20)
+  .fill(null)
+  .map(makeRandomEScooterRentalStation);
+const setLocation = action("setLocation");
+
+class ZoomControlledMapWithVehicleRentalOverlay extends Component {
```
binh-dam-ibigroup: Could this be achieved with just a story showing a different initial zoom level for the layer?

binh-dam-ibigroup: Alternatively, you could set up a knob for the storybook user to change the zoom interactively.

PR #29, opened 2020-01-16
Location field improvements

### LINKED ISSUE #12: location field: styling improvements
- easier to extend styling if className prop is passed down to root element 
location-3

the link in the image above is: https://www.styled-components.com/docs/basics#styling-any-component

  - export distinct className ... help overcome hacks needed to control nested elements
location-4

  - menu items in location field ... possible not to have menu display stuff?
location-5

  - aria-label supplied by user of component
location-6

### LINKED ISSUE #13: location field: more data returned from geocoder
AltSource's comments:
location-9

Frank's $0.02 (see SEE #9 for more...):
We're going to need a geosearch form to only search on the 'stops' layers in Pelias.  
https://ws-st.trimet.org/pelias/v1/search?text=6&layers=stops

We're also going to need that form to return the stop_id and agency_id fields, which are encoded in the response via the id field -- e.g., {stop_id}:{agency_id}:{layer_name} is how the id field is composed in Pelias for transit (stops from GTFS).

SEE #9 for more...

### LINKED ISSUE #14: location field: override icons (possibly using Slots pattern)
Our components will eventually need to allow the caller to override icons and other styling.  

location-8

evansiroky:
This PR has the following improvements:

- Adds a result type to the onLocationSelected callback (potentially resolves #13)
- Improve styling by (potentially resolves #12)
  - converting some plain jsx/html to styled components
  - adding a className to the root element to allow for styled-component styling
  - adding a story demonstrating how the LocationField component can be used as a styled component
- Allow customization of all the icons (potentially resolves #14)

This PR is built on top of #28


packages/core-utils/src/types.js
```diff
@@ -338,3 +338,38 @@ export const configuredCompanyType = PropTypes.shape({
...
+  }).isRequired,
+  properties: PropTypes.shape({
+    label: PropTypes.string.isRequired
+  }).isRequired
+});
+
+export const userLocationType = PropTypes.shape({
+  id: PropTypes.string,
+  /**
+   * Can be either 'home', 'work', or null
```
binh-dam-ibigroup: Are these "suggested" icon ids? In that case it should state so.

PR #30, opened 2020-01-16
Add layer add/remove, viewport callbacks inside overlays.

binh-dam-ibigroup:
This PR adds the ability to listen to the maps onOverlayAdded, onOverlayRemoved, onViewportChanged from inside each overlay (provided the overlay implements componentDidMount()).


packages/base-map/__mocks__/SelectVehicles.js
```diff
@@ -27,7 +29,24 @@ class SelectVehicles extends MapLayer {
 
   farZoom = 10;
 
-  componentDidMount() {}
+  onOverlayAdded() {
+    console.log("SelectVehicles => onOverlayAdded");
```
evansiroky: Could we use storybook actions for these so they can be seen without opening the debugger?

packages/base-map/src/funcs.js
```diff
@@ -0,0 +1,6 @@
+export default function callIfValid(fn) {
```
evansiroky: I'd prefer calling this file `utils.js`.

evansiroky: Also, please add some jsDoc for this function.

binh-dam-ibigroup: Do you think this could live in core-utils?

fpurcell: Yeah, I think it makes sense to live in core-utils ... pretty generic (check if param is a function, and return that as a closure or return a closure that does nothing -- that's what it's doing, right?).

evansiroky: Actually I was just thinking a little more about this. I think maybe a better way to go about this is to make an abstract overlay class with methods that can be overridden that each overlay can implement.

binh-dam-ibigroup: Yep that is the next level, that would remove the need for the derived class to register itself.

fpurcell: BTW, I was trying to use this with the vehicles code that I've been working on.  I refactored that code to use functional components, and couldn't get it to work.  So switched back to a class, and that kinda worked.

But there might be a pretty big performance issue here.  With ~500 markers (ala vehicles), turning a layer on & off is really slow.  From what I can tell, React-Leaflet makes every marker a separate layer.  Researching...

packages/base-map/__mocks__/SelectVehicles.js
```diff
@@ -1,9 +1,12 @@
 import React from "react";
 import { FeatureGroup, MapLayer, withLeaflet } from "react-leaflet";
 
+import callIfValid from "../src/funcs";
```
fpurcell: As mentioned below, moving this function to core-utils would be a good thing!

packages/base-map/__mocks__/SelectVehicles.js
```diff
@@ -20,14 +22,39 @@ class SelectVehicles extends MapLayer {
     vehicles: vehicleData
   };
 
+  constructor(props) {
+    super(props);
+
+    this.onOverlayAdded = this.onOverlayAdded.bind(this);
+    this.onOverlayRemoved = this.onOverlayRemoved.bind(this);
+    this.onViewportChanged = this.onViewportChanged.bind(this);
```
evansiroky: These bindings can be done by rewriting the class methods as follows:

```js
onOverlayAdded = e => {
  ...
}
```

packages/base-map/src/index.js
```diff
@@ -47,6 +49,15 @@ L.Evented.include({
  * The base OpenTripPlanner map on which everything else is rendered.
  */
 class BaseMap extends Component {
+  overlays = [];
+
+  constructor() {
+    super();
+    this.handleOverlayAdded = this.handleOverlayAdded.bind(this);
+    this.handleOverlayRemoved = this.handleOverlayRemoved.bind(this);
+    this.handleViewportChanged = this.handleViewportChanged.bind(this);
```
evansiroky: These bindings can be done by rewriting the class methods as follows:

```js
handleOverlayAdded = e => {
  ...
}
```

packages/base-map/src/index.js
```diff
@@ -67,7 +78,42 @@ class BaseMap extends Component {
...
     const { onClick } = this.props;
-    if (typeof onClick === "function") onClick(e);
+    callIfValid(onClick)(e);
+  };
+
+  forwardOne = (eventName, e) => {
+    const layer = this.overlays.find(child => child.props.name === e.name);
+    if (layer) callIfValid(layer[eventName])(e);
+
+    // Raise onOverlayAdded for this control.
```
evansiroky: Comment should be rewritten since more than just onOverlayAdded could be dispatched here.

packages/base-map/src/index.js
```diff
@@ -67,7 +78,42 @@ class BaseMap extends Component {
...
+    // eslint-disable-next-line react/destructuring-assignment
+    callIfValid(this.props[eventName])(e);
+  };
+
+  forwardAll = (eventName, e) => {
+    this.overlays.forEach(layer => {
+      callIfValid(layer[eventName])(e);
+    });
+
+    // Raise onOverlayAdded for this control.
```
evansiroky: Comment should be rewritten since more than just onOverlayAdded could be dispatched here.

packages/base-map/src/index.js
binh-dam-ibigroup: I clarified the comment.

packages/base-map/src/index.js
binh-dam-ibigroup: I clarified the comment.

packages/base-map/src/index.js
binh-dam-ibigroup: Good catch, the handlers were already written this way...

packages/base-map/__mocks__/SelectVehicles.js
binh-dam-ibigroup: I made the change and added a comment in that example layer stating that both syntaxes are supported.

packages/base-map/__mocks__/SelectVehicles.js
```diff
@@ -22,44 +22,46 @@ class SelectVehicles extends MapLayer {
...
 
   farZoom = 10;
 
-  onOverlayAdded(e) {
+  /**
+   * The preferred way to write the `onOverlay...` events is shown below.
+   * If you wish to write `onOverlayAdded(e) {...}` instead,
+   * then you must bind that function in the `constructor` using:
+   *     this.onOverlayAdded = this.onOverlayAdded.bind(this);
+   */
```
evansiroky: This should be common knowledge for react developers. This comment could be deleted or better yet replaced with something descriptive of how this handler is activated.

PR #34, opened 2020-01-22
Raw geocode response

### LINKED ISSUE #33: location field: expose record type and id from geocode response in the onLayerSelected callback
See #13 

Would like calling app to have access to data from the geocoder beyond name and coordinate.  In Pelias there are 2 properties that carry needed information.  The 'layers' property describes what type of record (e.g., a stop vs. an address vs. a venue, etc..).  The 'id' property, for certain records, contains valuable information, like stop_id and agency_id (for layers=stops).  Eventually there will be a 'routes' layer, which will contain a route_id, etc...  The use of those 2 properties (type and id) might work well, since all records in Pelias have those fields, plus guessing other geocoders have similar values (e.g., should make things universal).  That said, open to suggestions on other solutions and ideas to communicate needed information back to the calling app.

Screen Shot 2020-01-21 at 10 27 28 AM

evansiroky:
This PR does the following:

- Includes the raw geocoder result in geocode responses (fixes #33)
- Adds a project-wide test script
- Corrects some grammar in a LocationField story

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "lint:styles": "stylelint \"packages/**/*.js{,x}\"",
     "lint": "yarn lint:js && yarn lint:styles",
     "prettier": "prettier --write \"**/*.{json,md,yml}\"",
+    "test": "yarn lint:js && yarn lint:styles && yarn unit",
     "unit": "jest"
   },
   "eslintConfig": {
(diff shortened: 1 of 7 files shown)


evansiroky:
> BTW, I had to make changes to location-field's package.json file (along with other component's package.json files). Don't think we have to rebase w/master here, but thought I'd mention it.

Yeah, whenever I develop locally I often have to restart storybook and rebuild the packages. So I appreciate you updating the package.json files for us.

fpurcell:
> appreciate you updating the package.json files for us.

No problem...happy to help (or as Cole says, doing what could be automated :-)).  This is the commit: https://github.com/opentripplanner/otp-ui/commit/51e4bd4ece1f8005fde5725fc93225d2d9ca82f1

Was mostly the version of base-map not getting updated, due to peerDeps.  But for location-field, it had a tricky (I missed it the first time thru ... like 9.0.1, where other uses were 9.1.0) version number mis-match on styled-icons, which resulted in it's own node_modules junk.

PR #35, opened 2020-01-22
Trip form styling

binh-dam-ibigroup:
This PR enables custom styling for all of the trip form components.
You can customize classes that are exposed in `styled.js` at the root of the package.
File `styled.js` contains CSS that only handles the minimal layout. An example custom styling is provided in `__mocks__/trimet.styled.js`.


packages/trip-form/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@opentripplanner/trip-form",
-  "version": "0.0.11",
+  "version": "0.0.12",
```
fpurcell: Hey Binh ... think we might want to bump this back to 0.0.11.  When I execute the lerna stuff, it will increment this to 0.0.12 for the release (that said, I'll probably for a version to that number for all components, so it might not matter).

packages/trip-form/src/styled.js
```diff
@@ -0,0 +1,135 @@
...
+
+SubmodeSelector.Row = styled.div``;
+
+SubmodeSelector.InlineRow = styled(SubmodeSelector.Row)`
+  text-align: right;
+`;
+
+// ////////////////////////////////////////////////
+// ///////////// Generic components ///////////////
+// ////////////////////////////////////////////////
```
evansiroky: All the components below seem to also be components directly related to the app. I'd remove this comment. Eventually, we should figure out a better approach so that we're at least reusing some common button components or being better about making components that'd work with styled-system or whatever.

binh-dam-ibigroup: I removed those carry over comments from one of the styled.js you wrote that I copied.

packages/trip-form/package.json
binh-dam-ibigroup: I reverted back to 0.0.11.

PR #43, opened 2020-01-24
Allow creation of custom endpoint icons

### LINKED ISSUE #37: Allow custom icons for EndpointsOverlay component
In order to match our mocks, we'll need to customize the icons used in the `EndpointsOverlay` component. The way we supply custom icons to the `LocationField` component seems to work well. I'm good with following the same pattern if it can be used. If there is an existing prop that already allows me to do that then just let me know. Thanks a lot.

evansiroky:
Fixes #37. Allows passing a component as a prop to the endpoints overlay which is then used to render the map icons for the endpoints.

### DIFF
--- a/packages/endpoints-overlay/src/endpoint.js
+++ b/packages/endpoints-overlay/src/endpoint.js
@@ -18,30 +18,10 @@ import {
 
 import * as Styled from "./styled";
 
-function DefaultMapMarkerIcon({ location, type }) {
-  return (
-    <Styled.StackedIconContainer title={location.name}>
-      {type === "from" ? (
-        // From icon should have white circle background
-        <>
-          <Styled.StackedCircle size={24} />
-          <Styled.StackedLocationIcon size={24} type={type} />
-        </>
-      ) : (
-        <>
-          <Styled.StackedToIcon size={24} type="to" />
-          <Styled.ToIcon size={20} type={type} />
-        </>
-      )}
-    </Styled.StackedIconContainer>
-  );
-}
-
-DefaultMapMarkerIcon.propTypes = {
-  location: locationType.isRequired,
-  type: PropTypes.string.isRequired
-};
-
+/**
+ * These icons are used to render common icons for user locations. These will
+ * only show up in applications that allow saving user locations.
+ */
 function UserLocationIcon({ type }) {
   switch (type) {
     case "briefcase":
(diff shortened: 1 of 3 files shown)


packages/endpoints-overlay/src/endpoint.js
```diff
@@ -18,7 +18,31 @@ import {
...
+function DefaultMapMarkerIcon({ location, type }) {
+  return (
+    <Styled.StackedIconContainer title={location.name}>
+      {type === "from" ? (
+        // From icon should have white circle background
+        <>
+          <Styled.StackedCircle size={24} />
+          <Styled.StackedLocationIcon size={24} type={type} />
+        </>
+      ) : (
```
binh-dam-ibigroup: These are really two different icons.

evansiroky: I don't think this deserves a refactor as suggested. I think the single custom component prop as currently designed provides an additional benefit of leaving more things open to the implementing application by providing both the `location` and `type` properties. In this way the implementing app is unconstrained and could implement more than just 2 icon types perhaps based on location attributes or even use a single icon type.

binh-dam-ibigroup: If multiple icons can be returned by the implementer, then we should rename `MapMarkerIcon` to `mapMarkerIconProvider`.

packages/endpoints-overlay/src/endpoint.js
```diff
@@ -18,7 +18,31 @@ import {
...
+    </Styled.StackedIconContainer>
+  );
+}
+
+DefaultMapMarkerIcon.propTypes = {
+  location: locationType.isRequired,
+  type: PropTypes.string.isRequired
+};
+
+function UserLocationIcon({ type }) {
```
landonreed: This could use jsdoc. I had trouble remembering what these icons were for at first.

packages/endpoints-overlay/src/index.js
```diff
@@ -96,6 +111,7 @@ EndpointsOverlay.defaultProps = {
   forgetPlace: noop,
   rememberPlace: noop,
   locations: [],
+  MapMarkerIcon: undefined,
```
landonreed: Is it worth pulling the `DefaultMapMarkerIcon` out of the `endpoint.js` default props and into this file? Perhaps not since users of the library may want to use endpoint components independently from this overlay. But I think it's worth posing the question.

fpurcell: Landon's idea sounds like a good one. But maybe better to ticket this for a future version. My preference at the moment is to get AltSource all the stuff they need (they demo'd their work yesterday, and this PR came up, due to the icons not matching the mockups).

packages/endpoints-overlay/src/endpoint.js
binh-dam-ibigroup: At least, rename MapMarkerIcon to something that suggests that different icons can/will be rendered.

Issue #18, opened 2020-01-06
Itinerary Components Feedback from AltSource


PaulIvanov:
We primarily have styling feedback for items that we cant override using Styled-Components. There is stuff that is related to component ordering and actual content within the components. 
Most of the requests are also related to getting the view to look more like the UI mocks the Trimet marketing team provided for us. Let me know if you have any questions or comments, thanks. 

OTP UI Feedback:
- Change verbage on Transit Legs to have "Board" verb in it (Old: 52 Farmington/185th to Beaverton TC) (NEW: Board 52-Farmington/185th to Beaverton TC)
- Internals of "LegClickable" under the transit component causes centered multi-line styling.
- Allow for real-time arrival information to be passed into the itinerary body/Trip Legs.
- Change Trip Viewer Details info from "Ride 10 min / 15 stops" to "10-minute ride (15 stops)".
- Drop the "Service operated by Trimet" section on a transit leg. Or make it optional via a prop/config
- Add the service alerts section below the Trip viewer details section on a leg, instead of being on top.
- If The Service Alert section only has 1 alert, show it by default with a brief one-line display of the error.
- Some PropTypes could get some improvements. It was difficult to know what all the functions did and what they should take without diving into the actual library code.
- Is there a Handler for the "View leg on Map"/Map button on the right-side of the trip leg?

PaulIvanov:
@fpurcell

PaulIvanov:
The time stamp on the itinerary leg doesn't display the meridian. We dug into this and it looks like we are observing the time format set it the `timeOptions` object that is a prop of `ItineraryBody`; it conforms to this shape `{format: string}` OTP-UI does contain a module that contains the most complicated types, but we also pass in the timeFormat and dateFormat within the config. Should we be using the timeFormat from the config to at least allow us to override time options from the config? It's currently confusing which is the preferred way to get at the functionality we want.

evansiroky:
Something that may still be needed is a custom renderer for the route name.

PR #44, opened 2020-01-24
Add slot for custom route description component

evansiroky:
This PR adds a slot to the ItineraryBody component for a custom component that can render the route description. See a new story titled `ItineraryBody with E-scooter rental + transit itinerary with OTP-RR styling and customizations` for an example implementation.

Refs #18.

### DIFF
--- a/packages/itinerary-body/src/TransitLegBody/index.js
+++ b/packages/itinerary-body/src/TransitLegBody/index.js
@@ -54,10 +54,9 @@ function DefaultRouteDescription({ leg }) {
         {routeLongName}
         {headsign && (
           <span>
-            {" "}
             <Styled.LegDescriptionHeadsignPrefix>
-              to
-            </Styled.LegDescriptionHeadsignPrefix>{" "}
+              {" to "}
+            </Styled.LegDescriptionHeadsignPrefix>
             {headsign}
           </span>
         )}
--- a/packages/itinerary-body/src/TransitLegBody/index.js
+++ b/packages/itinerary-body/src/TransitLegBody/index.js
@@ -39,6 +39,37 @@ DefaultTransitLegSummary.propTypes = {
   stopsExpanded: PropTypes.bool.isRequired
 };
 
+function DefaultRouteDescription({ leg }) {
+  const { headsign, routeLongName, routeShortName } = leg;
+  return (
+    <Styled.LegDescriptionForTransit>
+      {routeShortName && (
+        <div>
+          <Styled.LegDescriptionRouteShortName>
+            {routeShortName}
+          </Styled.LegDescriptionRouteShortName>
+        </div>
+      )}
+      <Styled.LegDescriptionRouteLongName>
+        {routeLongName}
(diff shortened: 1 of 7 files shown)


packages/itinerary-body/src/TransitLegBody/index.js
```diff
@@ -39,6 +39,37 @@ DefaultTransitLegSummary.propTypes = {
   stopsExpanded: PropTypes.bool.isRequired
 };
 
+function DefaultRouteDescription({ leg }) {
+  const { headsign, routeLongName, routeShortName } = leg;
+  return (
+    <Styled.LegDescriptionForTransit>
+      {routeShortName && (
+        <div>
```
binh-dam-ibigroup: Does the `<div>` element accomplish anything that `<ItineraryBodyClasses.LegDescriptionRouteShortName>` doesn't already do?

packages/itinerary-body/src/TransitLegBody/index.js
```diff
@@ -39,6 +39,37 @@ DefaultTransitLegSummary.propTypes = {
...
+          <Styled.LegDescriptionRouteShortName>
+            {routeShortName}
+          </Styled.LegDescriptionRouteShortName>
+        </div>
+      )}
+      <Styled.LegDescriptionRouteLongName>
+        {routeLongName}
+        {headsign && (
+          <span>
+            {" "}
```
binh-dam-ibigroup: Roll these spaces into the CSS for `<Styled.LegDescriptionHeadsignPrefix>`, either as :before/:after content or as margin/padding? Then remove the surrounding `<span>` if it doesn't accomplish anything that `<Styled.LegDescriptionHeadsignPrefix>` can't already do.

packages/itinerary-body/src/TransitLegBody/index.js
evansiroky: Yes, it has some styles applied from the `LegDescriptionForTransit` styling rules. Then, the `LegDescriptionRouteShortName` contains a margin property that is able to be applied since this div in question has a display type of `table-cell` which disables the effect of margin settings.

packages/itinerary-body/src/TransitLegBody/index.js
evansiroky: That sounds a little too fancy. This approach results in the content being all in one place which seems a little simpler and easier to maintain.

binh-dam-ibigroup: Another way to do it is to roll the spaces like this: `<Styled.LegDescriptionHeadsignPrefix>{" to "}</Styled.LegDescriptionHeadsignPrefix>`.

evansiroky: That looks the cleanest of all.

packages/itinerary-body/src/TransitLegBody/index.js
binh-dam-ibigroup: A table??

Issue #47, opened 2020-01-25
Transitive Overlay rendering is inaccurate without intermediate stops


randolphjones:
When rendering an itinerary without intermediate stops, the transitive overlay looks to be inaccurate. It looks like it only renders the geometry between the last two stops on each leg. 

With intermediate stops:
![image](https://user-images.githubusercontent.com/5219049/73116058-045f2a80-3ee5-11ea-81bb-151ee1a7a023.png)
Without:
![image](https://user-images.githubusercontent.com/5219049/73116080-5ef88680-3ee5-11ea-80fc-c73faba28faf.png)

fpurcell:
See #42 and #36

randolphjones:
To clarify, I am seeing this issue with both `core-utils` and `transitive-overlay` on v `0.0.12`

evansiroky:
This bug report is what I deserve for not testing with data that actually came from OTP.

evansiroky:
FYI, the otp-react-redux application will always make requests for the intermediate stops as they are used in the ItineraryBody component to show how many stops a user has to ride. These stops are also displayed when a user clicks the DefaultTransitLegSummary component to toggle the view of stops. In otp-react-redux these intermediate stops are also rendered as a schematic element within the transit line display.

<img width="541" alt="Screen Shot 2020-01-26 at 10 29 38 AM" src="https://user-images.githubusercontent.com/3112493/73139865-55773780-4027-11ea-9b54-fc8a88de8820.png">

The design of otp-react-redux attempts to make the line/stops display in the itinerary body consistent with what is shown on the map in the transitive overlay. I'm not sure if this current design really needs those intermediate stops showing up in the line display so thought I'd bring this up since this functionality might be more suited for just the otp-react-redux implementation.

randolphjones:
Thanks. That all makes sense. I'm ok with us always making the request with intermediate stops as well. @fpurcell may have a strong preference here when it comes to how he wants the app we're building to function. At a glance it looks like we want it to operate the same way as OTP-RR.

PR #49, opened 2020-01-26
Properly render transitive geometry when no intermediate stops exist

evansiroky:
Fixes https://github.com/opentripplanner/otp-ui/issues/47


packages/core-utils/src/map.js
```diff
@@ -213,7 +213,9 @@ export function itineraryToTransitive(itin, companies) {
         stop_id: leg.to.stopId,
         geometry:
           hasInterStopGeometry &&
-          leg.interStopGeometry[leg.interStopGeometry.length - 1].points
+          (hasIntermediateStopGeomerty
```
landonreed: Just noticed this typo in `Geomerty`.

PR #57, opened 2020-02-04
isNaN bugfix

### LINKED ISSUE #50: NaN bug with `query/planParamsToQuery` util
We found a bug with the `planParamsToQuery` function inside of `v0.0.12` of `core-utils`
https://github.com/opentripplanner/otp-ui/blob/c6e6225749b8cdece619154cdc5c4e0b32c931dc/packages/core-utils/src/query.js#L213

Strings passed to the default case in this function are being turned into NaN. Having compared this to the OTP-RR code it looks like the global `isNan` was swapped with `Number.isNaN` when being added to OTP-UI's core utils. `Number.isNaN` will not implicitly coerse strings.

Worth noting that empty string values passed to the default condition still return NaN even with the global version of isNaN being used. We should be able to gracefully handle this condition in url query params

There are a couple of ways to address this. Feel free to reach out if you need some ideas or guidance. I think the best solution here will be to coerce the value with to a number before checking it with `Number.isNaN` 

If this is the way it's intended to work then we can double-check this after the utility that constructs the OTP routing params has been released.

Thanks!

evansiroky:
Fixes #50 and adds a test to prove fix.

### DIFF
--- a/packages/core-utils/src/query.js
+++ b/packages/core-utils/src/query.js
@@ -209,9 +209,11 @@ export function planParamsToQuery(params, config) {
       case "time":
         query.time = params.time || getCurrentTime(config);
         break;
-      default:
-        if (!Number.isNaN(params[key])) query[key] = parseFloat(params[key]);
-        else query[key] = params[key];
+      default: {
+        const maybeNumber = parseFloat(params[key]);
+        query[key] = Number.isNaN(maybeNumber) ? params[key] : maybeNumber;
+        break;
+      }
     }
   });
   return query;
--- a/packages/core-utils/src/__tests__/__snapshots__/query.js.snap
+++ b/packages/core-utils/src/__tests__/__snapshots__/query.js.snap
@@ -1,0 +1,28 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`query planParamsToQuery should parse a depart at query 1`] = `
+Object {
+  "companies": "",
+  "date": "2019-10-31",
+  "departArrive": "DEPART",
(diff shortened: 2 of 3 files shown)


packages/core-utils/src/query.js
```diff
@@ -209,9 +209,11 @@ export function planParamsToQuery(params, config) {
...
         query.time = params.time || getCurrentTime(config);
         break;
-      default:
-        if (!Number.isNaN(params[key])) query[key] = parseFloat(params[key]);
-        else query[key] = params[key];
+      default: {
+        const maybeNumber = parseFloat(params[key]);
+        query[key] = Number.isNaN(maybeNumber) ? params[key] : maybeNumber;
+        break;
+      }
```
landonreed: I'm worried about how this will handle strings that begin with a number character. E.g., let's say we're handling bannedRoutes, which is a comma-separated list of route IDs consisting of one route `897ABC`:
```
parseFloat('897ABC') // yields the number 897
Number.isNaN(897) // false
query[key] = 897 // instead of the correct string value 897ABC 
```
I think the original issue stated that OTP-RR used the global `isNaN` here, and I would perhaps advocate going back to that:
```
if (!isNaN('897ABC')) query[key] = parseFloat(params[key]) // conditional is true with global isNaN and float is parsed
else query[key] = params[key]
```
Also, we should add comments here about what's going on here because this is a sneaky issue.

evansiroky:
Good catch, @landonreed. I merged in your PR here.

landonreed:
Alright, @fpurcell, I think this is good to go!

Issue #38, opened 2020-01-23
Print Itinerary: add new component to itinerary related components


fpurcell:
![image](https://user-images.githubusercontent.com/64323/73016919-adb5fb80-3dd3-11ea-81ea-b248b54dfecb.png)

Have alternate trip itin rendering for print.

fpurcell:
PR #60 addresses this issue.

PR #63, opened 2020-02-12
Create Trip Details package

evansiroky:
Refs https://github.com/opentripplanner/otp-ui/issues/38

### DIFF
--- a/packages/trip-details/src/TripDetails.story.js
+++ b/packages/trip-details/src/TripDetails.story.js
@@ -32,79 +32,79 @@ const longDateFormat = "MMMM D, YYYY";
 storiesOf("TripDetails", module)
   .addDecorator(withA11y)
   .addDecorator(withInfo)
-  .add("ItineraryBody with walk-only itinerary", () => (
+  .add("TripDetails with walk-only itinerary", () => (
     <TripDetails
       itinerary={walkOnlyItinerary}
       longDateFormat={longDateFormat}
     />
   ))
-  .add("ItineraryBody with bike-only itinerary", () => (
+  .add("TripDetails with bike-only itinerary", () => (
     <TripDetails
       itinerary={bikeOnlyItinerary}
       longDateFormat={longDateFormat}
     />
   ))
-  .add("ItineraryBody with walk-transit-walk itinerary", () => (
+  .add("TripDetails with walk-transit-walk itinerary", () => (
     <TripDetails
       itinerary={walkTransitWalkItinerary}
       longDateFormat={longDateFormat}
     />
   ))
-  .add("Styled ItineraryBody with walk-transit-walk itinerary", () => (
+  .add("Styled TripDetails with walk-transit-walk itinerary", () => (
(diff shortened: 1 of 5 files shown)


packages/trip-details/src/TripDetails.story.js
```diff
@@ -0,0 +1,112 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import { withA11y } from "@storybook/addon-a11y";
+import { withInfo } from "@storybook/addon-info";
+import styled from "styled-components";
```
binh-dam-ibigroup: Suggestion: Move `styled` import below React's.

packages/trip-details/src/index.js
```diff
@@ -0,0 +1,215 @@
...
+};
+
+TripDetails.defaultProps = {
+  className: null,
+  longDateFormat: null,
+  routingType: "ITINERARY",
+  timeOptions: null
+};
+
+class TripDetail extends Component {
```
binh-dam-ibigroup: Move this to its own file, e.g. `/TripDetails/index.js`

packages/trip-details/src/TripDetails.story.js
evansiroky: I'm declining to do this since it would no longer be an alphabetic organization of imports.

evansiroky:
Just pushed a commit that implements most of @binh-dam-ibigroup's suggestions.

packages/trip-details/src/index.js
```diff
@@ -0,0 +1,145 @@
+import {
+  calculateFares,
+  calculatePhysicalActivity
+} from "@opentripplanner/core-utils/lib/itinerary";
```
fpurcell: A note about using .../lib/... paths on import statements:

In a future PR, I'm going to go thru all the files in otp-ui and switch all the .../lib/... import paths to .../src/...  Cole pointed out that by us using transpiled .../lib/... code, we're preventing webpack from tree-shaking the otp-ui code, resulting in larger footprints than might be necessary.  We might do more, based on Cole's continued research in making the code performant.

We're not there yet tho, so no action necessary just yet. More just a heads-up.
We're at the 'get it running phase' right now.
Next, we'll move to 'get it running correctly'.
Then when (if) we get to the 'get it running fast' phase, expect said changes...

fpurcell: BTW, I've tried out the .../src/... paths in a dev bundle on npm (https://www.npmjs.com/package/@ott_test/transit-vehicle-overlay), and I was able to pull in base-map, core-utils and transit-vehicle-overlay, which used the .../src/... paths, and my simple_app (https://github.com/trimet-sandbox/simple_app), as well as dev version of trimet.org were able to pull in the code and do a latter transpile in building the app.  More testing necessary, but...

landonreed: Do you want to copy this comment to an issue so that we can track it outside of this PR, @fpurcell?

Issue #64, opened 2020-02-13
Long press functionality in BaseMap component


randolphjones:
We have a request on the TORA project to implement long-press functionality in the app. The user will be using this for specifying a current location manually from the map instead of using geolocation API.  `onMapLongPress` prop added to basemap (or something equivalent, the name can be up to you). 

Alternately if you wanted to provide more granularity to the 'onMapClick' callback argument then we can handle things that way. We noticed that it passed an event name of `singleclick` so we can look at that event to switch behavior. This would keep your prop list a little shorter if you want.

Reach out if you have any questions or need any clarification. Thanks as always.

fpurcell:
#70 addresses this issue

PR #70, opened 2020-02-20
Basemap right click

binh-dam-ibigroup:
Adds context menu (long-press on mobile devices) support. Addresses #64.


binh-dam-ibigroup:
1. @fpurcell I noticed the popup in the storybook was gone for good after closing... Have you found an workaround with your other devs?
2. We'll have to find out more.
3. I got the popup appear on long-press on my Galaxy S9.

binh-dam-ibigroup:
Should this be converted to an issue?

> Can't help with making a popup repeatable to be opened and closed. The only react / leaflet popup things I have experience with, there is a marker that owns and controls the popup. Repeated open/close of a marker popup works differently I guess than a detached popup.

PR #78, opened 2020-03-04
Support missing OTP configuration data.

binh-dam-ibigroup:
There are some instances where OTP configuration is incomplete (e.g. no micromobility entries), and
we want to handle that gracefully.

### DIFF
--- a/packages/trip-form/src/SettingsSelectorPanel/index.js
+++ b/packages/trip-form/src/SettingsSelectorPanel/index.js
@@ -93,9 +93,13 @@ export default class SettingsSelectorPanel extends Component {
       // If there are multiple (scooter | bikeshare) providers,
       // then if one is specified by the mode button, select it,
       // othewise select all providers.
+      // selectedCompanies is at least an empty array.
       const selectedCompanies =
         defaultCompany ||
-        getCompanies(supportedCompanies, nonTransitModes).map(comp => comp.id);
+        getCompanies(supportedCompanies, nonTransitModes).map(
+          comp => comp.id
+        ) ||
+        [];
 
       this.handleQueryParamChange({
         mode: finalModes.join(","),
@@ -169,9 +173,10 @@ export default class SettingsSelectorPanel extends Component {
     );
     const nonTransitModes = selectedModes.filter(m => !isTransit(m));
     const companies = getCompaniesOptions(
-      supportedCompanies.filter(comp =>
-        defaultCompany ? comp.id === defaultCompany : true
-      ),
+      supportedCompanies &&
+        supportedCompanies.filter(comp =>
+          defaultCompany ? comp.id === defaultCompany : true
+        ),
       nonTransitModes,
       this.getSelectedCompanies()
     );
@@ -226,7 +231,7 @@ export default class SettingsSelectorPanel extends Component {
           )}
 
         {/* This order is probably better. */}
-        {companies.length >= 2 && (
+        {companies && companies.length >= 2 && (
           <SubmodeSelector
             label="Use companies"
             modes={companies}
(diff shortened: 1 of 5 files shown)


packages/trip-form/src/SettingsSelectorPanel/index.js
```diff
@@ -93,9 +93,13 @@ export default class SettingsSelectorPanel extends Component {
       // If there are multiple (scooter | bikeshare) providers,
       // then if one is specified by the mode button, select it,
       // othewise select all providers.
+      // selectedCompanies is at least an empty array.
       const selectedCompanies =
```
fpurcell: Hey Binh ... seeing a console error here:

TypeError: Object(...)(...) is null
2 index.js:97
    SettingsSelectorPanel index.js:97
    handleClick index.js:22
    onClick index.js:32
    React 17
    bind_applyFunctionN self-hosted:1054
    dispatchEvent self-hosted:1018
    unstable_runWithPriority scheduler.development.js:697
    React 4
    dispatchDiscreteEvent self-hosted:977
Source map error: Error: request failed with status 404

SEE IT: click the button on this story:
http://localhost:5555/?path=/story/settingsselectorpanel--settings-selector-panel-undefined-param

binh-dam-ibigroup: Thanks, good catch!

packages/trip-form/src/SettingsSelectorPanel/index.js
```diff
@@ -93,9 +93,12 @@ export default class SettingsSelectorPanel extends Component {
...
         defaultCompany ||
-        getCompanies(supportedCompanies, nonTransitModes).map(comp => comp.id);
+        (getCompanies(supportedCompanies, nonTransitModes) || []).map(
```
landonreed: Why not just change the default props value for supportedCompanies to an empty array?

binh-dam-ibigroup: supportedCompanies is passed from config.companies, which can be null/undefined.

binh-dam-ibigroup: But that seems to work too!

landonreed: Yea, my thinking here is that either `companies` should be defined in the config with an array OR if it's missing then it would default to this empty array. I don't think we necessarily need to tolerate the case where it is null. So anywhere you check that it exists `companies && companies.length > ...` you may be able to just remove that first check.

packages/trip-form/src/SettingsSelectorPanel/index.js
```diff
@@ -82,7 +82,8 @@ export default class SettingsSelectorPanel extends Component {
       }
 
       const nonTransitModes = newModes.length > 1 ? [newModes[1]] : ["WALK"]; // TODO: also accommodate WALK+DRIVE, WALK+e-scooter?? They already seem to work without WALK right now.
-      const defaultCompany = newModes.length > 2 ? [newModes[2]] : null; // To accommodate companies defined under accessModes.
+      const defaultAccessModeCompany =
+        newModes.length > 2 ? [newModes[2]] : null; // To accommodate companies defined under accessModes.
```
landonreed: This is just a minor style thing, so not necessary to change, but I generally prefer comments to be above the line not next to the line. Just improves legibility in a code editor.

packages/trip-form/src/settings-selector-panel.story.js
```diff
@@ -29,3 +30,12 @@ export const settingsSelectorPanel = () => (
 );
 export const settingsSelectorPanelStyled = () =>
   trimet(settingsSelectorPanel());
+
+export const settingsSelectorPanelUndefinedParams = () => (
+  <SettingsSelectorPanel
+    queryParams={queryParams}
+    supportedModes={commonModesEmpty}
+    supportedCompanies={undefined}
```
landonreed: Not sure it's necessary to set supportedCompanies to `undefined`, but I don't have a strong feeling on this.

binh-dam-ibigroup:
> Hey Binh. Visiting this story, I get a console dump of the entire story: http://localhost:5555/?path=/story/settingsselectorpanel--settings-selector-panel-styled

It seems to happen on the master branch too...

PR #83, opened 2020-03-06
refactor(callIfValid): Move callIfValid to core-utils

binh-dam-ibigroup:
This PR moves `callIfValid` function (was duplicate at one point) to core-utils and applies it in simple cases.


landonreed:
OK, my preference would be to remove it from the library. @fpurcell, is that OK with you?

binh-dam-ibigroup:
Closing this pull request, there is not much support to generalize this abstraction.

PR #89, opened 2020-03-15
Make from-to-location-picker and endpoints-overlay return consistent data structures and one other thing

evansiroky:
To be consistent with the `location-field` and the `stops-overlay` packages, this changes the `setLocation` callback in the `from-to-location-picker`and `endpoints-overlay` packages to have the same data structure.

### DIFF
--- a/packages/base-map/src/index.js
+++ b/packages/base-map/src/index.js
@@ -1,7 +1,7 @@
 import React, { Component } from "react";
 import PropTypes from "prop-types";
 import { LayersControl, Map, Popup, TileLayer } from "react-leaflet";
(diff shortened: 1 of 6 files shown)


evansiroky:
I've also added an unrelated item that removes a direct import from the `src` directory which was causing build problems in otp-react-redux.

evansiroky:
Just fixed the prop-type errors @fpurcell was seeing. 

Also, I improved the location-field handling of blur events. It'll now properly collapse the menu in a desktop context when clicking outside the menu and when pressing tab. This was noticed in https://github.com/opentripplanner/otp-react-redux/pull/137.

PR #92, opened 2020-03-20
Styling ModeButton title when component is selected

binh-dam-ibigroup:
This is a minor PR to allow the title of ModeButton to have a different style when the button is selected, as illustrated in the screenshot below.

image

### DIFF
--- a/packages/trip-form/src/ModeButton/index.js
+++ b/packages/trip-form/src/ModeButton/index.js
@@ -36,7 +36,10 @@ const ModeButton = props => {
       </Styled.ModeButton.Button>
 
       {title && showTitle && (
-        <Styled.ModeButton.Title className={disabledClassName} title={title}>
+        <Styled.ModeButton.Title
+          className={`${activeClassName} ${disabledClassName}`}
+          title={title}
+        >
           {title}
         </Styled.ModeButton.Title>
       )}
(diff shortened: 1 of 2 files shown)


packages/trip-form/src/__mocks__/trimet.styled.js
```diff
@@ -50,6 +50,10 @@ const TriMetStyled = styled.div`
     padding: 4px 0px 0px;
     font-size: 10px;
     line-height: 12px;
+
+    &.active {
```
evansiroky: We should not use classes like this with styled-components. Instead, the styled component should provide dynamic values based on the propsthat are sent. So this should be rewritten as:

```css
text-decoration: ${props => props.active && 'underline'}
```

packages/trip-form/src/ModeButton/index.js
```diff
@@ -36,7 +36,10 @@ const ModeButton = props => {
       </Styled.ModeButton.Button>
 
       {title && showTitle && (
-        <Styled.ModeButton.Title className={disabledClassName} title={title}>
+        <Styled.ModeButton.Title
+          className={`${activeClassName} ${disabledClassName}`}
```
evansiroky: We should not use classes like this with styled-components. Instead, the styled component should provide dynamic values based on the props that are sent. So this should be rewritten as:

```css
<Styled.ModeButton.Title
  selected={selected}
  title={title}
>
```

packages/trip-form/src/__mocks__/trimet.styled.js
binh-dam-ibigroup: Fair point. Except that props refers to the top-level styled component, in which case is SettingsSelectorPanel, and props contains the props of that component only.

PR #97, opened 2020-03-25
Refactor ItineraryBody so it can be formatted like the original ItineraryBody in otp-react-redux 

### LINKED ISSUE #94: Pass the leg index into frameLeg callback
We have done the necessary integration work to frame a trip leg within the trip tools map when the associated "map" icon button is clicked next to the associated leg. We are blocked from completing this work though because the `frameLeg` callback does not include the leg index when invoked inside of the OTP `ItineraryBody` component (to be more specific it's called inside of `PlaceRow`) I think this should be a simple matter of wrapping the onClick handler with an arrow function and supplying the `legIndex` as an argument to `frameLeg` however, we're open to any other things you may want to do (i.e. passing the entire leg if you think it meets your own needs).

At the very least, we need some reference to the leg that's just been clicked by the user. If I can offer up any more detail or clarification then please let me know. I think this should be fairly straightforward.

@fpurcell since we are so close to wrapping this feature, what are the changes of us getting this request expedited? I think this can be wrapped with a very small update. Let me know. Thanks.

evansiroky:
This PR adds and refactors a bunch of stuff so that the ItineraryBody component can be styled and rendered as close as possible to the original ItineraryBody component in otp-react-redux. This also introduces some breaking changes to improve code maintenance and avoid loading unnecessary components.

- Creates stories for each of the mock itineraries with otp-rr styling
- Extracts some individual components into single files to better organize things
- Adds new props to the root component:
  - LineColumnContent
  - showLegIcon
  - showMapButtonColumn
  - TransitLegSubheader
- Places all default components for slots into the defaults folder
- BREAKING: does not automatically use the default components for each slot. It is now expected for the developer to include these on their own. Doing this allows the application to not have to load potentially unnecessary components and code.
- Adds meaningful arguments when calling the `frameLeg` function (fixes #94).

### DIFF
--- a/packages/itinerary-body/src/AccessLegBody/access-leg-steps.js
+++ b/packages/itinerary-body/src/AccessLegBody/access-leg-steps.js
@@ -1,0 +1,39 @@
+import {
+  getStepDirection,
+  getStepStreetName
+} from "@opentripplanner/core-utils/lib/itinerary";
+import { stepsType } from "@opentripplanner/core-utils/lib/types";
+import { DirectionIcon } from "@opentripplanner/icons/lib/directions";
+import React from "react";
+
+import * as Styled from "../styled";
+
+export default function AccessLegSteps({ steps }) {
+  return (
+    <Styled.Steps>
+      {steps.map((step, k) => {
+        return (
+          <Styled.StepRow key={k}>
+            <Styled.StepIconContainer>
+              <DirectionIcon relativeDirection={step.relativeDirection} />
+            </Styled.StepIconContainer>
+
+            <Styled.StepDescriptionContainer>
+              {getStepDirection(step)}
+              <span>
+                {step.relativeDirection === "ELEVATOR" ? " to " : " on "}
+              </span>
+              <Styled.StepStreetName>
+                {getStepStreetName(step)}
+              </Styled.StepStreetName>
+            </Styled.StepDescriptionContainer>
+          </Styled.StepRow>
+        );
+      })}
+    </Styled.Steps>
+  );
+}
+
+AccessLegSteps.propTypes = {
+  steps: stepsType.isRequired
+};
--- a/packages/itinerary-body/src/AccessLegBody/rented-vehicle-subheader.js
+++ b/packages/itinerary-body/src/AccessLegBody/rented-vehicle-subheader.js
@@ -3,9 +3,10 @@ import {
   getCompaniesLabelFromNetworks,
   getModeForPlace
 } from "@opentripplanner/core-utils/lib/itinerary";
+import PropTypes from "prop-types";
 import React from "react";
 
-import * as Styled from "./styled";
+import * as Styled from "../styled";
 
 /**
  * A component to display vehicle rental data. The word "Vehicle" has been used
@@ -14,7 +15,12 @@ import * as Styled from "./styled";
  * assuming that the leg.rentedCar and leg.rentedBike response elements from OTP
  * will eventually be merged into the leg.rentedVehicle element.
  */
-export default function RentedVehicleLeg({ config, leg }) {
+export default function RentedVehicleSubheader({
+  config,
+  leg,
+  LegIcon,
+  showLegIcon
+}) {
   const configCompanies = config.companies || [];
 
   // Sometimes rented vehicles can be walked over things like stairs or other
@@ -22,7 +28,8 @@ export default function RentedVehicleLeg({ config, leg }) {
   if (leg.mode === "WALK") {
     return (
       <Styled.PlaceSubheader>
-        Walk vehicle along {leg.from.name}
+        {showLegIcon && <LegIcon leg={leg} />}
+        <span>Walk vehicle along {leg.from.name}</span>
       </Styled.PlaceSubheader>
     );
   }
@@ -67,10 +74,19 @@ export default function RentedVehicleLeg({ config, leg }) {
   // e.g., Pick up REACHNOW rented car XYZNDB OR
   //       Pick up SPIN E-scooter
   //       Pick up shared bike
-  return <Styled.PlaceSubheader>{rentalDescription}</Styled.PlaceSubheader>;
+  return (
+    <Styled.PlaceSubheader>
+      {showLegIcon && <LegIcon leg={leg} />}
+      <span>{rentalDescription}</span>
+    </Styled.PlaceSubheader>
+  );
 }
 
-RentedVehicleLeg.propTypes = {
+RentedVehicleSubheader.propTypes = {
   config: configType.isRequired,
-  leg: legType.isRequired
+  leg: legType.isRequired,
+  /** A component class used to render the icon for a leg */
+  LegIcon: PropTypes.elementType.isRequired,
+  /** Whether or not to show the leg icon next to the */
+  showLegIcon: PropTypes.bool.isRequired
 };
--- a/packages/itinerary-body/src/AccessLegBody/rented-vehicle-subheader.js
+++ b/packages/itinerary-body/src/AccessLegBody/rented-vehicle-subheader.js
@@ -27,12 +27,13 @@ export default function RentedVehicleSubheader({ config, leg }) {
     );
   }
 
-  let rentalDescription = "Pick up";
+  let rentalDescription;
   if (leg.rentedBike) {
     // TODO: Special case for TriMet may need to be refactored.
-    rentalDescription += ` shared bike`;
+    rentalDescription = "Pick up shared bike";
   } else {
     // Add company and vehicle labels.
+    let companyName = "";
     let vehicleName = "";
     // TODO allow more flexibility in customizing these mode strings
     let modeString = leg.rentedVehicle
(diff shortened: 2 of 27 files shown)


evansiroky:
@fpurcell, the null values occurring while clicking on the destination happen because there isn't another leg that continues after the destination. Maybe I'll include another parameter that indicates that the item clicked is the destination.

binh-dam-ibigroup:
Would it be possible to remove the custom font from the unstyled itinerary bodies?

packages/itinerary-body/src/AccessLegBody/access-leg-steps.js
```diff
@@ -0,0 +1,39 @@
...
+import { stepsType } from "@opentripplanner/core-utils/lib/types";
+import { DirectionIcon } from "@opentripplanner/icons/lib/directions";
+import React from "react";
+
+import * as Styled from "../styled";
+
+export default function AccessLegSteps({ steps }) {
+  return (
+    <Styled.Steps>
+      {steps.map((step, k) => {
```
binh-dam-ibigroup: Can steps be null/undefined?

fpurcell:
> @fpurcell, the null values occurring while clicking on the destination happen because there isn't another leg that continues after the destination. Maybe I'll include another parameter that indicates that the item clicked is the destination.

Yeah something.  The clicking on that map icon in the itinerary is going to then be used to move the map to that location. So having a record with coords available in the frameLeg() callback would be necessary.

packages/itinerary-body/src/AccessLegBody/rented-vehicle-subheader.js
```diff
@@ -0,0 +1,76 @@
...
+  getCompaniesLabelFromNetworks,
+  getModeForPlace
+} from "@opentripplanner/core-utils/lib/itinerary";
+import React from "react";
+
+import * as Styled from "../styled";
+
+/**
+ * A component to display vehicle rental data. The word "Vehicle" has been used
+ * because a future refactor is intended to combine car rental, bike rental
```
binh-dam-ibigroup: "because we plan to..."

packages/itinerary-body/src/AccessLegBody/rented-vehicle-subheader.js
```diff
@@ -0,0 +1,76 @@
...
+ * and micromobility rental all within this component. The future refactor is
```
binh-dam-ibigroup: "We are assuming that..."
If it is a new feature, I would avoid using the word refactor.

packages/itinerary-body/src/AccessLegBody/rented-vehicle-subheader.js
```diff
@@ -0,0 +1,76 @@
...
+    } else {
+      rentalDescription = "Continue using rental";
+    }
+
+    rentalDescription += ` ${modeString} ${vehicleName}`;
+  }
+  // e.g., Pick up REACHNOW rented car XYZNDB OR
+  //       Pick up SPIN E-scooter
+  //       Pick up shared bike
+  return <Styled.PlaceSubheader>{rentalDescription}</Styled.PlaceSubheader>;
```
binh-dam-ibigroup: Move this before the big if block.

packages/itinerary-body/src/AccessLegBody/rented-vehicle-subheader.js
```diff
@@ -0,0 +1,76 @@
...
+      <Styled.PlaceSubheader>
+        <span>Walk vehicle along {leg.from.name}</span>
+      </Styled.PlaceSubheader>
+    );
+  }
+
+  let rentalDescription = "Pick up";
+  if (leg.rentedBike) {
+    // TODO: Special case for TriMet may need to be refactored.
+    rentalDescription += ` shared bike`;
```
binh-dam-ibigroup: I would rather not concatenate strings (for future translations, if it gets there).

packages/itinerary-body/src/AccessLegBody/rented-vehicle-subheader.js
```diff
@@ -0,0 +1,76 @@
...
+    // beginning of the rental. It is possible that there will be some forced
+    // walking that occurs in the middle of the rental, so once the main mode
+    // resumes there won't be any network info. In that case we simply return
+    // that the rental is continuing.
+    if (leg.from.networks) {
+      const companiesLabel = getCompaniesLabelFromNetworks(
+        leg.from.networks,
+        configCompanies
+      );
+      rentalDescription += ` ${companiesLabel}`;
```
binh-dam-ibigroup: Same here

packages/itinerary-body/src/AccessLegBody/rented-vehicle-subheader.js
```diff
@@ -0,0 +1,76 @@
...
+      // IDs/names tend to be less relevant (or entirely useless) in this context.
+      if (leg.rentedCar && leg.from.name) {
+        vehicleName = leg.from.name;
+      }
+      modeString = getModeForPlace(leg.from);
+    } else {
+      rentalDescription = "Continue using rental";
+    }
+
+    rentalDescription += ` ${modeString} ${vehicleName}`;
```
binh-dam-ibigroup: Same here.

packages/itinerary-body/src/TransitLegBody/alerts-body.js
```diff
@@ -0,0 +1,54 @@
...
+        .map((alert, i) => {
+          // If alert is effective as of +/- one day, use today, tomorrow, or
+          // yesterday with time. Otherwise, use long date format.
+          const dateTimeString = moment(alert.effectiveStartDate).calendar(
+            null,
+            {
+              sameDay: `${timeFormat}, [Today]`,
+              nextDay: `${timeFormat}, [Tomorrow]`,
+              lastDay: `${timeFormat}, [Yesterday]`,
+              lastWeek: `${longDateFormat}`,
```
binh-dam-ibigroup: Do we need brackets around Today, Tomorrow, Yesterday?

packages/itinerary-body/src/defaults/line-column-content.js
```diff
@@ -0,0 +1,58 @@
...
+            <LocationIcon size={25} type="to" />
+          </Styled.Destination>
+        )}
+      </Styled.LineBadgeContainer>
+    </Styled.LegLine>
+  );
+}
+
+LineColumnContent.propTypes = {
+  /** whether this leg is an interlined-transit leg */
```
binh-dam-ibigroup: Capital W

randolphjones:
@fpurcell Thanks for pointing that out. Yes, we are trying to frame the selected leg on the map.
Two ways we can implement this: We can get the entire leg object as an argument and frame the view based on that or we can get the index and use that to select the leg from the user's active itinerary. Either way, we use the existing bound utilities with a leg object and frame up the view that way. Whether we get the entire leg from the get-go or the index of the leg... we can be flexible on that.

Taking a glance at the code in Evan's earlier commits, looks like we get both? I think we can definitely work w that 👍

packages/itinerary-body/src/defaults/route-description.js
```diff
@@ -0,0 +1,34 @@
...
+            {routeShortName}
+          </Styled.LegDescriptionRouteShortName>
+        </div>
+      )}
+      <Styled.LegDescriptionRouteLongName>
+        {routeLongName}
+        {headsign && (
+          <span>
+            <Styled.LegDescriptionHeadsignPrefix>
+              {" to "}
```
binh-dam-ibigroup: I think you can remove `{"`

packages/itinerary-body/src/index.js
```diff
@@ -36,46 +40,24 @@ const ItineraryBody = ({
   */
   const rows = [];
   let followsTransit = false;
+  let lastLeg;
   itinerary.legs.forEach((leg, i) => {
```
binh-dam-ibigroup: Can you use array.map to achieve this?

binh-dam-ibigroup:
What is the`otp-react-redux` subfolder for? Is it a mock folder?

evansiroky:
> Would it be possible to remove the custom font from the unstyled steps?
> 
> image

@binh-dam-ibigroup, there is no custom font present. The differences in font are due to default browser styling rules. It is generally expected that implementers will define fonts at a higher level than just this component.

packages/itinerary-body/src/AccessLegBody/rented-vehicle-subheader.js
evansiroky: Let's just wait until this refactor actually happens to change this comment.

packages/itinerary-body/src/AccessLegBody/rented-vehicle-subheader.js
evansiroky: This rentalDescription variable is modified in the immediately preceding if block, so I don't understand what you are asking me to do.

packages/itinerary-body/src/defaults/route-description.js
evansiroky: The spaces are needed, otherwise, the words will get squished together.

evansiroky:
I believe I've addressed most of the comments here. @fpurcell, regarding your requests, here is what I have done:

> a) rewrite 'Ride 3 min / 2 stops ▼ | Trip Viewer' as '3-minute ride (2 stops) ▼'
(or 'Ride 3 min (2 stops) ▼' ... btw, bold used here to differentiate desired to be state)

See story titled `ItineraryBody with walk-transit-walk itinerary with custom view trip button activated and custom route abbreviation` and corresponding code.

> b) remove the MA from the bubble

See story titled `ItineraryBody with walk-transit-walk itinerary with custom view trip button activated and custom route abbreviation` and corresponding code.

> c) align the time near the leg circle

See story titled `Styled ItineraryBody with walk-transit-walk itinerary` and corresponding code.

> frameLeg() callback

The frameLeg callback now also sends the `isDestination` and `place` values in callbacks.

packages/itinerary-body/src/AccessLegBody/rented-vehicle-subheader.js
binh-dam-ibigroup: I was suggesting to move the comment, so it describes the big if block does.
Alternatively, you can state in the existing comment that the examples are produced by the if block above.

binh-dam-ibigroup: Well, never mind, that is probably not necessary anymore as the block above has shrunk...

binh-dam-ibigroup:
This case producing a validation error. Is that intended?
image

evansiroky:
@binh-dam-ibigroup, see latest commit.

PR #106, opened 2020-04-06
Make the toRouteAbbreviation property optional

evansiroky:
The `toRouteAbbreviation` property isn't needed in some implementations that use a custom `LineColumnContent` component. Therefore, this PR makes it optional.


evansiroky:
I went ahead and just declared a function.

PR #113, opened 2020-04-23
Fix types

### LINKED ISSUE #96: `fareType.details` entries should be of type `arrayOf({ fareId, price, routes })`
The `fareType` type definition is not consistent with data returned by OTP and causes browser errors log inside the `TripDetails` and `PrintableItinerary` components. The errors read as follows:
image
image

OTP returns the following structure for fares:
image

For reference, the current definition of `fareType` is, per 
https://github.com/opentripplanner/otp-ui/blob/master/packages/core-utils/src/types.js#L267:
```js
export const fareType = PropTypes.shape({
  details: PropTypes.objectOf(
    PropTypes.shape({
      fareId: feedScopedIdType.isRequired,
      price: moneyType.isRequired,
      routes: PropTypes.arrayOf(feedScopedIdType).isRequired
    }).isRequired
   ...
```

### LINKED ISSUE #103: Query prop validation errors in OTP-RR
It looks like the `queryType` we defined in the core-utils package and used in the `SettingsSelectorPanel` has type information that is not consistent with the query data that is passed from OTP-RR.

Affected fields: (they are not used inside `SettingsSelectorPanel`)
* `queryType.from`
* `queryType.to`

Validation error:
* Expected: `String`
* Actual: `{name: String, lat: Number, lon: Number}`

image

binh-dam-ibigroup:
This PR fixed these non-blocking issues that cause unnecessary error messages in the browser console:
* fix #103 
* fix #96 

To test on your machine until next release: in the affected components, replace `import ... from "@opentripplanner/core-utils/lib/types"` with `import ... from "../../core-utils/src/types"`.

EDIT: Affected components: `GeneralSettingsPanel` and `SettingsSelectorPanel` in package `trip-form`.

### DIFF
--- a/packages/core-utils/src/types.js
+++ b/packages/core-utils/src/types.js
@@ -498,8 +498,8 @@ export const modeSelectorOptionsType = PropTypes.shape({
 });
 
 export const queryType = PropTypes.shape({
-  from: PropTypes.string,
-  to: PropTypes.string,
+  from: locationType,
+  to: locationType,
   date: PropTypes.string,
   time: PropTypes.string,
   departArrive: PropTypes.string,
--- a/packages/core-utils/src/types.js
+++ b/packages/core-utils/src/types.js
@@ -266,11 +266,16 @@ const moneyType = PropTypes.shape({
  */
 export const fareType = PropTypes.shape({
   details: PropTypes.objectOf(
-    PropTypes.shape({
-      fareId: feedScopedIdType.isRequired,
-      price: moneyType.isRequired,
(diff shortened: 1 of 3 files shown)


landonreed:
Perhaps we should add these instructions to the docs for testing out changes to sibling packages:

```
1. Find and replace:
"@opentripplanner/package-to-test": "current-version"
with
"@opentripplanner/core-utils": "file:../package-to-test"
2. Find and replace:
opentripplanner/package-to-test/lib/types
with
opentripplanner/package-to-test/src/types
3. Run:
yarn && yarn dev
```

PR #117, opened 2020-04-27
docs(README): add part about test sibling pkg

landonreed:
I was struggling to test a sibling package and this seems like a good way to handle. If someone else has a better way, I'm happy to modify this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,13 +22,13 @@ See:
 
 Some packages in otp-ui depend on sibling packages (e.g., `@opentripplanner/core-utils` is used by many of its siblings). In order to test a package with local changes you have made to its sibling, you can run the following find/replace operations to make sure you're depending on your latest work (and not the released version):
 
-1. Find and replace (package-to-test being the package with local changes -- make sure these are committed to first to avoid the find/replace operations below polluting your work):
+1. In the `package.json` files for packages in which you want to test the sibling, find and replace (package-to-test being the package with local changes -- make sure these are committed to first to avoid the find/replace operations below polluting your work):
 
     > "@opentripplanner/package-to-test": **"current-version"**
     
     with
     
-    > "@opentripplanner/core-utils": **"file:../package-to-test"**
+    > "@opentripplanner/package-to-test": **"file:../package-to-test"**
 
 2. Find and replace:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,6 +18,29 @@ See:
  yarn dev # (opens storybook to running component library on localhost:5555)
 ```
 
+## Development
+
+Some packages in otp-ui depend on sibling packages (e.g., `@opentripplanner/core-utils` is used by many of its siblings). In order to test a package with local changes you have made to its sibling, you can run the following find/replace operations to make sure you're depending on your latest work (and not the released version):
+
+1. Find and replace (package-to-test being the package with local changes -- make sure these are committed to first to avoid the find/replace operations below polluting your work):
```
binh-dam-ibigroup, line 25: This needs to say it is for `package.json`

README.md
```diff
@@ -18,6 +18,29 @@ See:
...
+
+    > "@opentripplanner/package-to-test": **"current-version"**
+    
+    with
+    
+    > "@opentripplanner/core-utils": **"file:../package-to-test"**
+
+2. Find and replace:
```
binh-dam-ibigroup, line 33: "In your source files, find and replace:"

README.md
```diff
@@ -18,6 +18,29 @@ See:
...
+
+Some packages in otp-ui depend on sibling packages (e.g., `@opentripplanner/core-utils` is used by many of its siblings). In order to test a package with local changes you have made to its sibling, you can run the following find/replace operations to make sure you're depending on your latest work (and not the released version):
+
+1. Find and replace (package-to-test being the package with local changes -- make sure these are committed to first to avoid the find/replace operations below polluting your work):
+
+    > "@opentripplanner/package-to-test": **"current-version"**
+    
+    with
+    
+    > "@opentripplanner/core-utils": **"file:../package-to-test"**
```
binh-dam-ibigroup, line 31: /package-to-test (not /core-utils)

landonreed:
Thanks @binh-dam-ibigroup. How is that?

README.md
binh-dam-ibigroup: Second bullet point still needs a clarification.

landonreed:
@binh-dam-ibigroup, you may have been filtering by only the first commit. I think the second bullet reflects your requested change.

binh-dam-ibigroup:
Sorry, you were right @landonreed.

PR #119, opened 2020-04-29
Add classic icons

### LINKED ISSUE #118: Icons package: Add and use a non-branded icon set
Some implementers of OTP-UI library may want to use a non-branded default icon set without having to provide their own icons. Currently, many UI components use `TriMetModeIcon` and `TriMetLegIcon` which contain a number of branded items. There should be an unbranded equivalent `ModeIcon` and `LegIcon`. In addition, the unbranded sets should contain some of the 'classic' OTP-rr icons (rail, etc.)

binh-dam-ibigroup:
This PR covers the following:
- Add the icon set from OTP-react-redux into the icons package (icons/classic) (EDIT: fix #118),
- Add the icon set from the existing TriMet MOD OTP into the icons package (icons/mod), 
- Publish icon sets `ClassicModeIcon`, `ClassicLegIcon`, `TriMetModModeIcon`, `TriMetModLegIcon`,
- [Breaking] Rename prop `TriMetModeIcon.label` to `TriMetModeIcon.mode`
- Add props to allow `PrintableItinerary` and components in `itinerary-body` to use icon sets other than `TriMetModeIcon`,
- Remove `ModeIcon` from package trip-from.
- [Breaking] Replace prop `SettingsSelectorPanel.icons` with `SettingsSelectorPanel.ModeIcon` which can be one of the icon sets above.
- Fix issue where selecting one rental mode selected all similar options.


packages/icons/src/classic-mode-icon.js
```diff
@@ -0,0 +1,51 @@
...
+      return <ClassicCar {...props} />;
+    case "ferry":
+      return <ClassicFerry {...props} />;
+    case "gondola":
+      return <ClassicGondola {...props} />;
+    case "micromobility":
+    case "micromobility_rent":
+      return <ClassicMicromobility {...props} />;
+    case "rail":
+    case "subway":
```
landonreed, line 39: I think it might be useful to add separate rail and subway icons at some point. This can be a separate issue for later though.

packages/icons/src/leg-icon.js
```diff
@@ -0,0 +1,77 @@
...
+import PropTypes from "prop-types";
+import React from "react";
+
+import Biketown from "./companies/biketown-icon";
+import Bird from "./companies/bird-icon";
+// import Bolt from "./companies/bolt-icon";
+// import Car2go from "./companies/car2go-icon";
+import Gruv from "./companies/gruv-icon";
+import Lime from "./companies/lime-icon";
+// import Lyft from "./companies/lyft-icon";
```
landonreed, line 11: Why are these icons commented out?

packages/icons/src/mod/index.js
```diff
@@ -0,0 +1,19 @@
+import ModBike from "./mod-bike";
```
landonreed, line 1: I think the `mod/` directory (and `Mod*` icon names) should be called `trimet/` (TriMet)

packages/icons/src/leg-icon.js
binh-dam-ibigroup: I don't know. They were that way initially. We can add them back.

packages/icons/src/mod/index.js
landonreed: OK, never mind. I guess there already is a TriMet directory. This is confusing to me. Can you add to the package's README about different use cases and distinguish between the icon sets? Before you do that though, let's hammer out this mod vs. trimet distinction.

binh-dam-ibigroup: There is already a trimet/ folder.

packages/icons/src/trimet-mod-leg-icon.js
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+
+import LegIcon from "./leg-icon";
+import TriMetModModeIcon from "./trimet-mod-mode-icon";
+
+const TriMetModLegIcon = props => (
+  <LegIcon ModeIcon={TriMetModModeIcon} {...props} />
+);
```
landonreed, line 8: Let's consider deleting in favor of `trimet-leg-icon`. It seems silly to have multiple TriMet leg icons.

packages/icons/src/trimet-mod-mode-icon.js
```diff
@@ -0,0 +1,34 @@
+import React from "react";
```
landonreed, line 1: Let's consider deleting.

packages/icons/src/trimet/Bus copy.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
```
landonreed, line 1: Is this icon supposed to be here?

packages/itinerary-body/src/TransitLegBody/index.js
```diff
@@ -158,6 +163,7 @@ export default class TransitLegBody extends Component {
 TransitLegBody.propTypes = {
   config: configType.isRequired,
   leg: legType.isRequired,
+  LegIcon: PropTypes.elementType.isRequired,
```
landonreed, line 166: Shouldn't this be optional (with a default value provided)?

packages/icons/src/trimet/Bus copy.js
binh-dam-ibigroup: No... thanks for catching that.

packages/itinerary-body/src/otp-react-redux/route-description.js
```diff
@@ -43,5 +43,6 @@ export default function RouteDescription({ leg }) {
 }
 
 RouteDescription.propTypes = {
-  leg: legType.isRequired
+  leg: legType.isRequired,
+  LegIcon: PropTypes.elementType.isRequired
```
landonreed, line 47: Again, it seems like making this optional and using a default value is probably the right approach.

packages/trip-form/src/util.js
```diff
@@ -107,52 +206,60 @@ function getTransitCombinedModeOptions(icons, modes, selectedModes) {
...
-      selected:
-        !selectedModes.some(isTransit) && selectedModes.some(modeObj.isActive),
+    .filter(({ mode }) => exclusiveModes && exclusiveModes.includes(mode))
+    .map(({ isActive, label, mode }) => ({
+      id: mode,
+      selected: !selectedModes.some(isTransit) && selectedModes.some(isActive),
       showTitle: false,
       text: (
         <span>
-          <ModeIcon icons={icons} mode={modeObj.mode} /> {modeObj.label}
```
landonreed, line 122: Is this whitespace preserved?
`<ModeIcon icons={icons} mode={modeObj.mode} /> {modeObj.label}`

packages/trip-form/src/util.js
```diff
@@ -107,52 +206,60 @@ function getTransitCombinedModeOptions(icons, modes, selectedModes) {
...
 }
 
 /**
  * Generates the options (primary, secondary, tertiary) for the mode selector based on the modes read from config.yaml.
  * @param {*} modes The modes defined in config.yaml.
+ * @param {*} icons The icon set to use.
+ * @param {*} modes The available modes to choose from.
  * @param {*} selectedModes An array of string that lists the modes selected for a trip query.
+ * @param {*} selectedCompanies The companies to show as selected.
+ * @param {*} supportedCompanies The supported companies for certain modes.
```
landonreed, line 241: This jsdoc doesn't seem right.

packages/trip-form/src/util.js
binh-dam-ibigroup: Good catch. No it is not.

packages/trip-form/src/util.js
```diff
@@ -49,7 +47,56 @@ export function getModeString(modeObj) {
...
+ * @param {*} modes The desired modes for which to get the operating companies.
+ * @returns An array of companies that operate the specified modes (should not be undefined as companies is an array).
+ */
+function getCompanies(companies, modes) {
+  return companies
+    .filter(
+      comp => comp.modes.split(",").filter(m => modes.includes(m)).length > 0
+    )
+    .filter(comp => hasRental(comp.modes) || hasHail(comp.modes));
+}
```
landonreed, line 62: I think this code `getCompanies` was just moved up in the file, which generally I would tend to avoid in a large PR like this... it just makes it seem like new code when it's not actually.

packages/trip-form/src/util.js
```diff
@@ -49,7 +47,56 @@ export function getModeString(modeObj) {
...
+
+/**
+ * Returns an array containing the company ids, in upper case for MOD UI URLs, for the specified mode id.
+ * The mode id scheme is set and used by function getTransitCombinedModeOptions().
+ * @param {*} id The mode id to process.
+ * @param {*} supportedCompanies The list of supported companies (see structure in __mocks__/companies.js).
+ */
+export function getCompaniesForModeId(id, supportedCompanies) {
+  const newModes = id.split("+"); // Duplicate logic.
+  const nonTransitModes = newModes.length > 1 ? [newModes[1]] : ["WALK"]; // Duplicate logic.
```
landonreed, line 72: What are these mentions of duplicate logic?

packages/trip-form/src/util.js
```diff
@@ -49,7 +47,56 @@ export function getModeString(modeObj) {
...
+
+  // Accommodate companies defined under accessModes.
+  // Convert company ID to upper case for passing to MOD UI URL.
+  const defaultAccessModeCompany =
+    newModes.length > 2 ? [newModes[2].toUpperCase()] : null;
+
+  // If there are multiple (scooter | bikeshare | etc.) providers,
+  // then if one is specified by the mode button, select it,
+  // othewise select all providers.
```
landonreed, line 81: This language about `selecting` is confusing here.

packages/trip-form/src/util.js
```diff
@@ -69,36 +116,88 @@ export function getTransitSubmodeOptions(icons, modes, selectedModes) {
...
     accessModes &&
     accessModes.map(modeObj => {
       const modeStr = getModeString(modeObj);
+      const modeCompany = modeObj.company
+        ? modeObj.company.toUpperCase()
+        : null;
+
+      const id = `TRANSIT+${modeStr}${
+        modeObj.company ? `+${modeObj.company}` : ""
+      }`;
```
landonreed, line 167: I think id would be easier to read if you put the result of the company ternary into a variable above:

```
const company = modeObj.company ? `+${modeObj.company}` : ""
const id = `TRANSIT+${modeStr}${company}`;
```

packages/trip-form/src/util.js
binh-dam-ibigroup: That was enforced by es-lint I'm afraid.

packages/trip-form/src/util.js
```diff
@@ -69,36 +116,88 @@ export function getTransitSubmodeOptions(icons, modes, selectedModes) {
...
+ */
+function getPrimaryModeOption(ModeIcon, selectedModes) {
   return {
     id: "TRANSIT",
     selected: selectedModes.some(isTransit) && selectedModes.includes("WALK"),
     showTitle: false,
     text: (
       <span>
-        <ModeIcon icons={icons} mode="TRANSIT" />
+        <ModeIcon mode="TRANSIT" />
```
landonreed, line 131: In general, I'm not too sure about the idea of removing the ability to supply custom icons for the `ModeIcon` component. It seems like that was an important part of handling custom stuff in trimet-mod-otp (e.g., https://github.com/ibi-group/trimet-mod-otp/blob/septa/lib/config.js#L24-L36), but it's been a while since I thought about how all of the pieces fit together, so I may be off there.

binh-dam-ibigroup: It looks @evansiroky has ported that logic to `TriMetLegIcon` in the early days at https://github.com/opentripplanner/otp-ui/blob/master/packages/icons/src/trimet-leg-icon.js. The idea is if an agency has custom logic they would supply their own ModeIcon and/or or LegIcon components. Is this used anywhere else beside legs?

packages/itinerary-body/src/otp-react-redux/route-description.js
binh-dam-ibigroup: If we do that, there should be a package-wide default value.

packages/itinerary-body/src/TransitLegBody/index.js
binh-dam-ibigroup: If we do that, there should be a package-wide default value. That should be handled separately IMO.

packages/icons/src/classic-mode-icon.js
binh-dam-ibigroup: I don't have a separate icon on hand, so that should be dealt with separately.

packages/icons/src/trimet-mod-leg-icon.js
binh-dam-ibigroup: I have renamed the `mod` set to `standard`.

binh-dam-ibigroup:
I have addressed PR comments changes. Please review. Thanks in advance!

packages/icons/src/classic-mode-icon.js
landonreed: Agreed.

packages/trip-form/src/util.js
landonreed: I'm not sure... is it used in the SettingsSelector, @binh-dam-ibigroup?

binh-dam-ibigroup: I these methods, I replaced parameter `icons` with `ModeIcon` and removed the component of that name. The call highlighted is using the parameter.

packages/trip-form/src/util.js
```diff
@@ -233,12 +233,12 @@ function getExclusiveModeOptions(ModeIcon, modes, selectedModes) {
...
- * @param {*} modes The available modes to choose from.
- * @param {*} selectedModes An array of string that lists the modes selected for a trip query.
- * @param {*} selectedCompanies The companies to show as selected (when the user selects an exclusive mode operated by multiple companies).
- * @param {*} supportedCompanies The supported companies for certain access modes.
+ * @param modes The modes defined in config.yaml.
+ * @param icons The icon set to use.
+ * @param modes The available modes to choose from.
+ * @param selectedModes An array of string that lists the modes selected for a trip query.
+ * @param selectedCompanies The companies to show as selected (when the user selects an exclusive mode operated by multiple companies).
+ * @param supportedCompanies The supported companies for certain access modes.
```
landonreed, line 241: These params in the jsdoc don't match the function.

binh-dam-ibigroup: Oh my, thanks, I'm a victim of copy-paste!

binh-dam-ibigroup:
Hi @fpurcell, thanks for pointing out these error messages. Fortunately it was a simple storybook fix.

PR #120, opened 2020-04-30
refactor: make t-v-o a pure presentational component

### LINKED ISSUE #71: PEFORMANCE: transit-vehicles-overlay and leaflet zoom & panTo
See #67 

binh-dam-ibigroup: Is there a way for the overlays to not control the map?
  setMapZoom = throttle(500, zoom => {
    this.setState({ mapZoom: zoom });
    this.recenterMap();
  });

TODO: understand why zoom and pan events don't just update the vehicles layer, as opposed to placing the zoom level in a state variable, etc.... End result will hopefully find a way to simplify the zoom (state) and panTo calls in RT Vehicles.

Why does map zooming and panTo (i.e., recenterMap()) not just work, and why do I seemingly need a state variable for zoom level for my layer to recognize viewport / zoom /  pan events?  Why do calls to leaflet.map.panTo not just move the map -- why do I also need to update data for panTo to move the map?

Note: the workarounds now are to track the map's zoom level in a state variable, which then triggers a React redraw of the vehicle marker layer.  There's got to be a better / simpler way to have the layer redraw different icons based on zoom level.

### LINKED ISSUE #101: Custom Icon prop for Vehicles Layer
We would like to have a custom icon prop for the vehicles layer. This would be an icon that we can use to mark a vehicle on our overlay and would allow us to replace the icon that's currently being used to represent a vehicle. This is a lower priority than some of the other Vehicle layer requests.

We've managed to implement the custom icon for StopsOverlay. I think something similar could work here as well but we are open to ideas on this.

### LINKED ISSUE #102: Slot for Custom Popup on Vehicle Layer
We would like to have a prop/slot for a custom popup on the vehicle layer.

As it stands, we will only need to display one of these at a time. See the example of how we would use it to display the Vehicle time. It would follow the tracked vehicle.

image

For this, I think that the slot pattern might serve us best where we can pass in a particular implementation. Inside of the implementation, we could have the component 

```jsx
<Vehicles
    ...
    customVehiclePopup={<MyCustomVehiclePopup/>}
    ...
/>
```
At a glance, we may be able to use a popup element for this but I am open on the implementation. We may try a couple of different approaches to see what works here, we haven't experimented with passing a stateful component into a popup yet but I think we can find a way to make something work. 

Alternately we can always pass in some static markup. I will run some experiments on Monday to see.

fpurcell:
- Transform the vehicles layer to a purely presentational component (set of components).
- Purely functional / hooks now ... no more class based components.
- There are (optional) supporting utils for fetching and managing data, etc...
- There is also the flexibility to use different realtime vehicle data services.
- Finally, a big change is in customizing the vehicle markers, where both circle and rectangular markers are given (and others could be created); customization extends also to popups and tooltips.

Storybook: https://trimet-sandbox.github.io/otp-ui/?path=/story/transitvehicleoverlay--real-time-rectangles

closes #101 
closes #102 
closes #71


packages/transit-vehicle-overlay/src/__tests__/app.js
```diff
@@ -0,0 +1,64 @@
...
+
+// render the component
+let root;
+act(() => {
+  root = create(makeApp())
+});
+
+// make assertions on root
+expect(root.toJSON()).toMatchSnapshot();
+*/
```
landonreed, line 55: Any reason all of this is commented out? I think we should either remove or add a better comment about why it's commented out.

packages/transit-vehicle-overlay/src/components/VehicleGeometry/index.js
```diff
@@ -0,0 +1,90 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { useLeaflet } from "react-leaflet";
+
+import { transitVehicleType } from "@opentripplanner/core-utils/lib/types";
+import ModeCircles from "../markers/ModeCircles";
+import { linterIgnoreTheseProps } from "../../utils";
+
+export default function VehicleGeometry(props) {
```
landonreed, line 9: Component is well documented, but could use some top-level jsdoc.

packages/transit-vehicle-overlay/src/utils/data.js
```diff
@@ -0,0 +1,132 @@
+/** utility functions for converting and manipulating vehicle data */
+
+/** this function is used to have props that are not used w/out lint errors */
+export function linterIgnoreTheseProps() {}
```
landonreed, line 4: This feels like cheating :)

packages/transit-vehicle-overlay/src/utils/data.js
```diff
@@ -0,0 +1,132 @@
...
+
+/** simple (python-esque) converstion of an int to a string */
```
landonreed, line 6: typo: conversion

packages/transit-vehicle-overlay/src/utils/data.js
```diff
@@ -0,0 +1,132 @@
...
+  try {
+    retVal = `${val}`;
+  } catch (e) {
+    console.error(e);
+  }
+  return retVal;
+}
+
+/** gets best appropriate vehicle tracker id from input parameter based on type and data */
+export function getVehicleId(tracked) {
```
landonreed, line 18: This arg name could be improved (e.g., `vehicle`).

packages/transit-vehicle-overlay/README.md
```diff
@@ -1,48 +1,33 @@
...
+(https://github.com/OpenTransitTools/gtfsdb_realtime/blob/master/ott/gtfsdb_realtime/model/response/vehicle_list.py).
+
+Other realtime vehicle data may be used with this layer. For example, there is demo code within the component that
+shows converting another service response format (https://developer.trimet.org/ws_docs/vehicle_locations_ws.shtml)
+to work with this layer.
+
+## NOTES:
+
+- The use of hooks was a good thing, but there's a bit of resulting redirection (HL functions, etc...)
+  in the compoent lib (and interface) which obfuscates things a bit ... appologies in advnace.
```
binh-dam-ibigroup, line 25: Typo: advance

packages/transit-vehicle-overlay/README.md
```diff
@@ -1,48 +1,33 @@
...
+- below are 2 utils I use in conjuction for converting .svg to React svg .js files
```
binh-dam-ibigroup, line 26: Typo: conjunction

packages/transit-vehicle-overlay/__mocks__/proprietaryFetchUtils.js
```diff
@@ -0,0 +1,105 @@
...
+      );
+    const pattern = utils.makePattern(geojson, vehicle.tripId);
+    if (setter) setter(pattern);
+  } catch (e) {
+    console.error(e);
+  }
+}
+
+/**
+ * the way the story is written (an maybe the way others will use the component), there are
```
binh-dam-ibigroup, line 91: Typo: and maybe

packages/transit-vehicle-overlay/src/components/markers/ModeRectangles/images/bus.js
```diff
@@ -0,0 +1,41 @@
+import * as React from "react";
+
+export default function Bus(props) {
+  return (
+    <svg
```
binh-dam-ibigroup, line 5: At some point in the future, we may need to watch for duplication with the icons package.

binh-dam-ibigroup:
@fpurcell, Is there no longer a need to show the vehicle layers in the map's legend/layer selector?
Previously, you could see this in the layer selector:
image

fpurcell:
> @fpurcell, Is there no longer a need to show the vehicle layers in the map's legend/layer selector?
> Previously, you could see this in the layer selector:
> image

Hey Binh.  Good catch. I added the name and visible props to the Component to allow layer switcher control.  (In my earlier defense, for the trimet.org project there is a custom layer switcher; so for that project, no need for things to show up in the leaflet layer selector, so just wasn't thinking about this...).

> @fpurcell, Is there no longer a need to show the vehicle layers in the map's legend/layer selector?
> Previously, you could see this in the layer selector:
> image

binh-dam-ibigroup:
Thank you @fpurcell for the changes.

fpurcell:
> Thank you @fpurcell for the changes.

Thanks for the review and good suggestions, Bihn.

packages/transit-vehicle-overlay/src/components/popups/VehicleTooltip/index.js
```diff
@@ -0,0 +1,54 @@
...
+
+  let rsn = vehicle.routeShortName;
+  if (rsn !== null && rsn.length <= 5) {
+    rsn = `Line ${rsn}:`;
+  }
+
+  return (
+    <Tooltip permanent={permanent} direction={direction} offset={offset}>
+      <TooltipStyle>
+        <TooltipStyle.Title>{rsn}&nbsp;</TooltipStyle.Title>
```
landonreed, line 24: I would prefer to see `{' '}` for whitespace rather than `&nbsp;`.

packages/transit-vehicle-overlay/src/components/popups/VehicleTooltip/index.js
```diff
@@ -0,0 +1,54 @@
...
+import { formatDurationWithSeconds } from "@opentripplanner/core-utils/src/time";
+import { TooltipStyle } from "../styled";
+import { linterIgnoreTheseProps } from "../../../utils";
+
+/** will show a (leaflet) map tooltip on a vehicle, showing route and update recency */
+export default function VehicleTooltip(props) {
+  const { vehicle, isTracked, direction, permanent, offset } = props;
+  linterIgnoreTheseProps(isTracked);
+
+  let rsn = vehicle.routeShortName;
```
landonreed, line 16: I don't love the use of this abbreviation `rsn`. I would prefer to see it called `displayName` or `name`.

packages/transit-vehicle-overlay/src/components/popups/VehicleTooltip/index.js
```diff
@@ -0,0 +1,54 @@
...
+  if (rsn !== null && rsn.length <= 5) {
+    rsn = `Line ${rsn}:`;
```
landonreed, line 18: For routes that _don't_ have a short name, it appears like this 
image

I.e., there is no colon at the end, but there is a route short name it adds a colon. Also, the use of the word "Line" feels inconsistent across different modes (works for rail, but to me it isn't quite a match for most bus services).

packages/transit-vehicle-overlay/src/index.story.js
```diff
@@ -4,168 +4,287 @@ import { withA11y } from "@storybook/addon-a11y";
...
+      />
+    </BaseMap>
+  );
 }
 
-function rtExample(name, visible) {
-  const geometryUrl = "https://newplanner.trimet.org/ws/ti/v0/index";
-  const vehicleUrl = "https://maps.trimet.org/gtfs/rt/vehicles/";
+/** with static data, show a simple version of the real-time transit vehicles layer */
+function realtimeExample(fetchVehicles, fetchPattern, markers) {
```
landonreed, line 175: I don't really understand the difference between realtimeExample and simpleExample based on the comment. Perhaps these function names could be more descriptive or the comments could be enhanced.

packages/transit-vehicle-overlay/src/index.story.js
```diff
@@ -4,168 +4,287 @@ import { withA11y } from "@storybook/addon-a11y";
...
 storiesOf("TransitVehicleOverlay", module)
   .addDecorator(withA11y)
   .addDecorator(withInfo)
   .addDecorator(withKnobs)
-  .add("by Route", routeExample)
-  .add("all Routes", allExample)
-  .add("animated VehicleLayer", animatedExample)
-  .add("real-time vehicles in layer switcher", rtExampleLayerSwitcher)
-  .add("real-time vehicles programmatic control", rtExampleProgrammatic);
+  .add("simple overlay", simple)
```
landonreed, line 285: Sometimes the tracked vehicle seems to get stuck. Screenshot shows the simple example that defaults to tracking MAX Blue. I changed to route 57, but the route geometry rendered did not switch over.

image

packages/transit-vehicle-overlay/src/utils/state.js
```diff
@@ -0,0 +1,160 @@
...
+  return [mapZoom, mapCenter, onViewportChanged];
+}
+
+/**
+ * trackedVehicleState
+ *
+ * use this hook when you want your layer to track a vehicle
+ * (and potentially show the route geometry of this vehicle)
+ *
+ * @param fetchPatternCallback() that I'll used to fetch a vehicles' pattern geom
```
landonreed, line 45: typo? `that is used`

packages/transit-vehicle-overlay/src/utils/state.js
```diff
@@ -0,0 +1,160 @@
...
+ * @param initial vehicle record
+ * @param initial route pattern
```
landonreed, line 47: It feels wrong to omit the name of the param and just include a description: https://jsdoc.app/tags-param.html

packages/transit-vehicle-overlay/src/index.story.js
```diff
@@ -4,168 +4,287 @@ import { withA11y } from "@storybook/addon-a11y";
...
+  .add("simple overlay (alternate vehicles service)", alternate)
+  .add("simple rectangles (click to select)", simpleRectangles)
+  .add("static rectangles (popups)", rectangles)
+  .add("real-time circles", rtCircles)
+  .add("real-time rectangles", rtRectangles);
```
landonreed, line 290: In general, with these stories it's helpful for me to see the actual component code as close as possible to the story declaration. I.e., 

```
  .add("simple overlay", () => {
    // convert geojson data into expected pattern format
    const patternGeometry = utils.makePattern(mockGeom, "111");
    // initial setup
    const fetchPattern = (vehicle, setter) => {
      utils.linterIgnoreTheseProps(vehicle);
      setter(patternGeometry);
    };
    // state setup for zoom (refreshes layer) and selected vehicles
    const [zoom, onViewportChanged] = utils.zoomState(INITIAL_ZOOM_LEVEL);
    const [
      getRoutePattern,
      getTrackedVehicle,
      updateTrackedVehicle
    ] = utils.trackedVehicleState(
      fetchPattern,
      utils.findVehicleById(allVehicles, "9050"), // initial vehicle_id
      patternGeometry
    );

    const [trackedVehicle, trackedRef] = getTrackedVehicle();
    utils.linterIgnoreTheseProps(trackedRef);
    VehiclePopup.defaultProps.setTracked = updateTrackedVehicle;
    return (
      <BaseMap
        center={PORTLAND}
        zoom={INITIAL_ZOOM_LEVEL}
        onViewportChanged={onViewportChanged}
      >
        <TransitVehicleOverlay
          zoom={zoom}
          vehicleList={allVehicles}
          onVehicleClicked={setClicked}
          selectedVehicle={trackedVehicle}
          pattern={getRoutePattern()}
          onRecenterMap={utils.recenterPanTo}
          // Imported components
          TooltipSlot={VehicleTooltip}
          PopupSlot={VehiclePopup}
          // Tip: Use storybook knobs to control the colors!
          color={color("color:", "#AAA")}
          highlightColor={color("isTracked color:", "#ece90d")}
          lowlightColor={color("trailing color:", "#AAA")}
        />
      </BaseMap>
    );
  })
```

There's probably a way to factor out some of the default props that are re-used by different stories, but I just generally find the usefulness of storybook in being able to click a story and see the code that represents that implementation. I know this might not be the case for some of the other otp-ui packages and that using predefined functions like you have might be somewhat unavoidable, but I do think we should strive for this.

packages/transit-vehicle-overlay/src/utils/state.js
```diff
@@ -0,0 +1,160 @@
+import { useState, useRef, useEffect, useCallback } from "react";
+import { checkRefreshInteval } from "./fetch";
+import * as data from "./data";
+
+/**
+ * use this hook when you want your layer to re-paint after otp-ui map zoom events fire
+ * this component will capture and save zoom state from otp-ui's base-map component to state
+ * :return: zoom level (number from leaflet) and the callback for base-map's onViewportChanged
+ */
+export function zoomState(initialZoom = 14) {
```
landonreed, line 10: For this (and any other hooks you define), I think it would be good for legibility to use the standard `useXyz` naming convention: https://reactjs.org/docs/hooks-custom.html#extracting-a-custom-hook

packages/transit-vehicle-overlay/src/utils/state.js
```diff
@@ -0,0 +1,160 @@
...
+    trackedVehicleRef.current = trackedVehicle;
+  }, [trackedVehicle]);
+
+  useEffect(() => {
+    routePatternRef.current = routePattern;
+  }, [routePattern]);
+
+  const getRoutePattern = vehicle => {
+    if (fetchPatternCallback && vehicle) {
+      const pid = routePatternRef.current ? routePatternRef.current.id : null;
```
landonreed, line 71: When I see `pid` I think process id. Could you change to pattern Id (or something else if that's not what this is)?

packages/transit-vehicle-overlay/src/utils/state.js
```diff
@@ -0,0 +1,160 @@
...
+  };
+
+  return [mapZoom, onViewportChanged];
+}
+
+/**
+ * use this hook when you want your layer to re-paint after otp-ui map zoom & recenter events fire
+ * this component will capture and save zoom and center from otp-ui's base-map viewport changes
+ * :return: zoom level, center [x,y] and the callback for base-map's onViewportChanged
+ */
```
landonreed, line 25: Might be useful to distinguish this jsdoc a little more explicitly from the above hook. Also, injecting some punctuation and caps might help legibility. Finally, the `:return:` syntax seems non-standard. Should this be `@returns`?

packages/transit-vehicle-overlay/src/utils/state.js
```diff
@@ -0,0 +1,160 @@
...
+      setRoutePattern(null);
+    } else if (vehicle) {
+      setTrackedVehicle(vehicle);
+      if (updatePattern) getRoutePattern(vehicle);
+    }
+  };
+
+  /**
+   * return both the tracked vehicle state variable (using will cause a redraw)
+   * and the ref to that vehcicle.
```
landonreed, line 102: typo: vehicle

packages/transit-vehicle-overlay/src/utils/state.js
```diff
@@ -0,0 +1,160 @@
...
+) {
+  const [trackedVehicle, setTrackedVehicle] = useState(initVehicle);
+  const [routePattern, setRoutePattern] = useState(initPattern);
+  const trackedVehicleRef = useRef(trackedVehicle);
+  const routePatternRef = useRef(routePattern);
+
+  // a ref + useEffect give a handle on the current trackedVehicle state in util functions, etc...
+  useEffect(() => {
+    trackedVehicleRef.current = trackedVehicle;
+  }, [trackedVehicle]);
```
landonreed, line 63: @fpurcell, is this `useRef` + `useEffect` pattern explained anywhere? From what I understand, refs are generally used to access the DOM, but I can't see where that's happening in this package. I'm pretty new to hooks, but the use of them in this package feels a bit off somehow. It feels like callbacks and hooks and tracked things in the "simple overlay" story (and others) are a bit too interwoven/interconnected.

packages/transit-vehicle-overlay/src/components/popups/VehicleTooltip/index.js
fpurcell: i.e., "Line" ... that was Myleen's suggestion a year or so back.  Going to use the 'mode' (ala "BUS", "TRAM", etc...) with "Line" as a default if no route type (mode) exists.

packages/transit-vehicle-overlay/src/index.story.js
fpurcell: There's just one route geometry (for the rail line) being pulled in from the __mocks__ directory.   The line split still happens nearest the x-axis of the selected vehicle.

fpurcell:
> @fpurcell, I don't know if others feel this way, but I'm having some trouble following the usage of all of the hooks/refs for tracking vehicles. It seems like it could be a somewhat non-standard way of doing things and I wonder if there's room for a refactor to better handle this logic.

Could always be better, and I agree it is confusing (e.g., don't look at how the sausage is made). I'm considering changes now, but not sure I have anything better to offer anytime soon. The good news is that the state utils are optional to use of the vehicle component...and there's nothing mandating a user of the component to chose these.  If a user has alternate ways to query data and track a vehicle, they are more than welcome to use that. What gets sent down to the component is simply a list of vehicles and optionally the vehicle + route geom to highlight.

packages/transit-vehicle-overlay/src/utils/state.js
fpurcell: https://overreacted.io/making-setinterval-declarative-with-react-hooks/#refs-to-the-rescue

packages/transit-vehicle-overlay/src/index.story.js
fpurcell: Your comment inspired me to create a simple -> complex stories, so a user can (hopefully) quickly get their head around the basic component.  So a good add, I think.

Placing a function between storybook .add and the implementation allowed me to demonstrate 6 different aspects of the component.  So keeping it DRY, there are 6 stories that show off various aspects (from using different service endpoints, to popups / no popups, etc...), but only 3 implementations of the component.

fpurcell:
BTW, I updated the online storybook:
https://trimet-sandbox.github.io/otp-ui/?path=/story/transitvehicleoverlay--simple-overlay

PR #140, opened 2020-05-21
Ui tweaks

### LINKED ISSUE #129: RouteViewerOverlay#leafletPath is incorrectly named (should be path according to PropTypes)
In render function the `path` prop is called `leafletPath`:
https://github.com/opentripplanner/otp-ui/blob/15b1615bf64c6b65f5af907f0ec64b8ff110b070/packages/route-viewer-overlay/src/index.js#L52

In propTypes it is called `path`:
https://github.com/opentripplanner/otp-ui/blob/15b1615bf64c6b65f5af907f0ec64b8ff110b070/packages/route-viewer-overlay/src/index.js#L91

### LINKED ISSUE #134: Adjust rendering of long bus route numbers.
Sometimes, bus route numbers don't fit in the circle (itinerary-body package).

image

### LINKED ISSUE #135: BaseMap children prop warning
This probably comes from incorrect PropTypes definition.

image

### LINKED ISSUE #136: Start location icon should be a 'dot-circle'
Current:
image

Desired:
image

### LINKED ISSUE #137: Walking directions collapser arrow should change on toggle.
image

See source comment at [this link](https://github.com/ibi-group/trimet-mod-otp/pull/260#issuecomment-627484712)

### LINKED ISSUE #138: Mobile view: grey background on map location picker.
"From" and "To" location pickers should be transparent.

image

### LINKED ISSUE #139: PropTypes validation errors under SettingsSelectorPanel
(Link to original comment)[https://github.com/ibi-group/trimet-mod-otp/pull/260#issuecomment-627470953].

image

binh-dam-ibigroup:
Addressing various visual bugs and console errors, most mentioned in https://github.com/ibi-group/trimet-mod-otp/pull/260.

This PR will:
- fix #139,
- fix #138,
- fix #137,
- fix #136,
- fix #135,
- fix #134,
- fix #129.

### DIFF
--- a/packages/core-utils/src/types.js
+++ b/packages/core-utils/src/types.js
@@ -511,7 +511,10 @@ export const queryType = PropTypes.shape({
   startTime: PropTypes.string,
   endTime: PropTypes.string,
   mode: PropTypes.string,
-  showIntermediateStops: PropTypes.bool,
+  showIntermediateStops: PropTypes.oneOfType([
+    PropTypes.bool,
+    PropTypes.string
+  ]),
   maxWalkDistance: PropTypes.number,
   maxBikeDistance: PropTypes.number,
   optimize: PropTypes.string,
@@ -522,9 +525,12 @@ export const queryType = PropTypes.shape({
   bikeSpeed: PropTypes.number,
   maxEScooterDistance: PropTypes.number,
   watts: PropTypes.number,
-  ignoreRealtimeUpdates: PropTypes.bool,
+  ignoreRealtimeUpdates: PropTypes.oneOfType([
+    PropTypes.bool,
+    PropTypes.string
+  ]),
   companies: PropTypes.string,
-  wheelchair: PropTypes.bool
+  wheelchair: PropTypes.oneOfType([PropTypes.bool, PropTypes.string])
 });
 
 export const configuredModeType = PropTypes.oneOfType([
--- a/packages/from-to-location-picker/src/styled.js
+++ b/packages/from-to-location-picker/src/styled.js
@@ -14,6 +14,7 @@ export const FromToPickerSpan = styled.span`
 `;
 
 export const Button = styled.button`
+  background: none;
   border: none;
   color: navy;
   font-family: inherit;
--- a/packages/route-viewer-overlay/src/index.js
+++ b/packages/route-viewer-overlay/src/index.js
@@ -49,13 +49,11 @@ class RouteViewerOverlay extends MapLayer {
   updateLeafletElement() {}
 
   render() {
-    const { leafletPath, routeData } = this.props;
+    const { routeData, path } = this.props;
 
     if (!routeData || !routeData.patterns) return <FeatureGroup />;
 
-    const routeColor = routeData.color
-      ? `#${routeData.color}`
-      : leafletPath.color;
+    const routeColor = routeData.color ? `#${routeData.color}` : path.color;
     const segments = [];
     Object.values(routeData.patterns).forEach(pattern => {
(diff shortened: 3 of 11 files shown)


packages/core-utils/src/types.js
```diff
@@ -511,7 +511,10 @@ export const queryType = PropTypes.shape({
   startTime: PropTypes.string,
   endTime: PropTypes.string,
   mode: PropTypes.string,
-  showIntermediateStops: PropTypes.bool,
+  showIntermediateStops: PropTypes.oneOfType([
+    PropTypes.bool,
+    PropTypes.string
+  ]),
```
landonreed, line 517: This doesn't seem like the right approach. Can we not change otp-rr to correctly set the value as boolean?

packages/from-to-location-picker/src/styled.js
```diff
@@ -14,6 +14,7 @@ export const FromToPickerSpan = styled.span`
 `;
 
 export const Button = styled.button`
+  background: none;
```
landonreed, line 17: Would inherit be better here? I'm not entirely sure the context of this change.

binh-dam-ibigroup: Inherit what?

binh-dam-ibigroup: Without the background override, the default behavior on Chromium browsers is to display these buttons gray. There is no other stylesheet that sets that anywhere else, so we have to do it here.

landonreed: Gotcha. Thanks for explaining.

packages/core-utils/src/types.js
binh-dam-ibigroup: @evansiroky or @landonreed do you know the rationale for passing `"true"` and `"false"` as a string in this test file? https://github.com/opentripplanner/otp-ui/blob/master/packages/core-utils/src/__tests__/query.js#L14

Two things I can do regardless:
- Convert all the "boolean" fields to just `String`,
- Or, just remove all fields that are not relevant to `SettingsSelectorPanel`.

landonreed: @binh-dam-ibigroup, the input params in this test file are designed to mimic what comes from directly from the URL query params (i.e., the output of [`#getUrlParams`](https://github.com/ibi-group/otp-ui/blob/update-leaflet-for-mapbox/packages/core-utils/src/query.js#L62-L64)). I think the correct approach is to just change `planParamsToQuery` to parse booleans correctly or somehow handle this with `query-params.js`.

binh-dam-ibigroup: @evansiroky what do you think? Dealing with `planParamsToQuery` or `query-params` feels more than a UI tweak.

binh-dam-ibigroup: @landonreed and I just agreed to remove `queryType` altogether.

For context, `queryType` was only used to avoid placing a lint PropType exception for `SettingsSelectorPanel` and `GeneralSettingsPanel` in package `trip-form`. Neither control uses any of the type members explicitly (the controls get which parameters to display from `planParamsToQuery` and `query-params`).

packages/route-viewer-overlay/src/index.js
```diff
@@ -49,21 +49,19 @@ class RouteViewerOverlay extends MapLayer {
   updateLeafletElement() {}
 
   render() {
-    const { leafletPath, routeData } = this.props;
+    const { routeData, path } = this.props;
```
evansiroky, line 52: it's no longer alphabetized!

PR #144, opened 2020-05-28
Update leaflet for Mapbox changes

landonreed:
This PR fixes otp-ui for the [upcoming Mapbox deprecation](https://blog.mapbox.com/deprecating-studio-classic-styles-d8892ac38cb4) and also simplifies the map-related stories a bit (removes unneeded css).


binh-dam-ibigroup:
@landonreed There is a build error, could you address it please?

binh-dam-ibigroup:
On displays with 1080p resolution or lower, I feel the street names would be too big. Thoughts @fpurcell? (The map does show nicely on "retina" displays)
image

binh-dam-ibigroup:
It looks like different tile providers have different takes on retina. Mapbox looks fine as proposed, however cartodb and stamen seem to handle resolution differently.

PR #156, opened 2020-06-18
Stop overlay: Remove listener on unmount

### LINKED ISSUE #150: stops-overlay: remove the moveend event handler on unmount
The OTP-UI stops overlay does not remove the moveend event handler when it unmounts. As a result when we turn the layer on/off for various purposes, we get a "zombie" process that continues to call the stops api. By moving between a few views, these can stack up pretty quickly. 

#124

binh-dam-ibigroup:
This PR fixes #150.

### DIFF
--- a/packages/stops-overlay/src/index.js
+++ b/packages/stops-overlay/src/index.js
@@ -14,7 +14,7 @@ class StopsOverlay extends MapLayer {
 
   // TODO: determine why the default MapLayer componentWillUnmount() method throws an error
   componentWillUnmount() {
-    // Remove the pan/zoom listener set up above only.
+    // Remove the pan/zoom listener set up above.
     this.props.leaflet.map.off("moveend", this.refreshStops);
   }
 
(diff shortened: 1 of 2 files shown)


packages/stops-overlay/src/index.js
```diff
@@ -9,13 +9,14 @@ import { FeatureGroup, MapLayer, withLeaflet } from "react-leaflet";
...
-    this.props.leaflet.map.on("moveend", () => {
-      this.refreshStops();
-    });
+    this.props.leaflet.map.on("moveend", this.refreshStops);
   }
 
   // TODO: determine why the default MapLayer componentWillUnmount() method throws an error
-  componentWillUnmount() {}
+  componentWillUnmount() {
+    // Remove the pan/zoom listener set up above only.
```
landonreed, line 17: Not sure I understand the last part of this comment "... set up above only."

binh-dam-ibigroup: Updated.

PR #157, opened 2020-06-18
LocationField: Add autoFocus prop.

### LINKED ISSUE #133: `LocationField`: focus on mount or forwarded ref functionality
I have a larger component wrapping `LocationField` and would like to focus the contained input element when my component opens. 

I was able to use a ref with `LocationField` and get the existing `inputRef` property but this breaks encapsulation and can break the focus feature if that member is ever changed within the component.

We could either have a forwarded ref to the input element or we could have a boolean prop that would focus the input ref on mount. This would be a more sustainable way for us to provide this functionality using the component. I think that a boolean prop would hold up strongly if `LocationField` was ever reworked to a function component.

If there is some functionality in `LocationField` that I may be overlooking, let me know. Thanks!

Here's some more info on forwarded refs in React https://reactjs.org/docs/forwarding-refs.html#forwarding-refs-to-dom-components

binh-dam-ibigroup:
This PR fixes #133. To use, set the `autoFocus` prop of the control that has the initial focus to `true` as you would do with other HTML input controls.

### DIFF
--- a/packages/location-field/src/LocationField.story.js
+++ b/packages/location-field/src/LocationField.story.js
@@ -239,4 +239,33 @@ storiesOf("LocationField", module)
       userLocationsAndRecentPlaces={userLocationsAndRecentPlaces}
       UserLocationIconComponent={UserLocationIconComponent}
     />
+  ))
+  .add("LocationField no auto-focus with multiple controls", () => (
+    <div>
+      <input id="example" value="Enter text" />
+      <button type="button">Click me!</button>
+      <LocationField
+        currentPosition={currentPosition}
+        geocoderConfig={geocoderConfig}
+        getCurrentPosition={getCurrentPosition}
+        inputPlaceholder="Select from location"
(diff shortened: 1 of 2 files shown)


binh-dam-ibigroup:
> Hi @binh-dam-ibigroup
> 
> I'm seeing this prop error when first selecting the focus story (http://localhost:5555/?path=/story/locationfield--locationfield-auto-focus-with-multiple-controls). I don't see the same error if I revisit the story again w/in the same session (but will see again if I refresh the browser):
> 
> Warning: Failed prop type: You provided a `value` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue`. Otherwise, set either `onChange` or `readOnly`.
> in input
> in div

@fpurcell I updated the the plain `<input>` controls used in these two stories in 41e6a8fa70ddc9722f9757ae7a154f54b1d2e77c to the correct syntax.

PR #158, opened 2020-06-18
VehicleRentalOverlay: Make refreshVehicles prop optional.

### LINKED ISSUE #127: Vehicle Rental Layer - Make refreshVehicles prop optional
The vehicle rental layer was pretty easy to implement. Thanks for building that, it was very helpful.  Having the MapSymbols prop that accounted for zoom level was particularly helpful and would be an excellent feature to have on some of the other overlays (like the stops overlay).

In one instance I found myself using the component inside of a collection: one layer for each e-scooter provider.  I wanted for them to share the same `refreshVehicles` callback though. It wasn't a hassle for me to wrap them with a component that made a call on behalf of all layers. I was forced to pass an empty function into `refreshVehicles` though since it was required.

Very minor thing and not a showstopper. Low priority.

binh-dam-ibigroup:
This PR fixes #127.

### DIFF
--- a/packages/vehicle-rental-overlay/src/index.js
+++ b/packages/vehicle-rental-overlay/src/index.js
@@ -46,6 +46,8 @@ class VehicleRentalOverlay extends MapLayer {
 
   startRefreshing() {
     const { refreshVehicles } = this.props;
+
+    // Create the timer only if refreshVehicles is a valid function.
     if (typeof refreshVehicles === "function") {
       // initial station retrieval
       refreshVehicles();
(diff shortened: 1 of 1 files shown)


packages/vehicle-rental-overlay/src/index.js
```diff
@@ -293,6 +298,7 @@ VehicleRentalOverlay.defaultProps = {
     return stationName;
   },
   mapSymbols: null,
+  refreshVehicles: null,
```
fpurcell, line 301: Should this be an empty function rather than null?

PR #163, opened 2020-06-25
Reverse geocode location name

### LINKED ISSUE #147: Zip codes in geocoder results
From @fpurcell: in both MOD and this new CALL, the reverse geocoder is appending the zipcode to results (compare the popup vs. the pelias 'label').  Want to remove that, and just use the 'label' attribute returned from Pelias (wrapper).  (Also, if this is happening on MOD, it's old code (pre OTP-UI).

image

Notes: also need to apply in OTP-RR.

binh-dam-ibigroup:
This PR fixes #147. 
The corresponding mock response from Pelias and the corresponding unit test snapshot have been updated so that the zip/country no longer appear in `location.name` from the geocoder result.


binh-dam-ibigroup:
> In each of the tests, the search text is specifically for "Mill Ends Park", so it feels wrong to change the fixture and snapshot to an address other than Mill Ends Park. Sure, the test passes and demonstrates correctness given the mock response, but it doesn't feel right.

@evansiroky is your suggestion to replace all "Mill Ends Park" with the other address that uses the zip code?

Issue #152, opened 2020-06-12
Error calling getUserTimeZone()


binh-dam-ibigroup:
This call fails:
```js
import coreUtils from '@opentripplanner/core-utils'
...
console.log(coreUtils.time.getUserTimeZone())
```

Resulting error message:
```log
time.js:110 Uncaught TypeError: Cannot read property 'guess' of undefined
    at getUserTimezone (time.js:110)
```

binh-dam-ibigroup:
This issue is seen by transpiling https://github.com/opentripplanner/otp-react-redux into https://github.com/ibi-group/trimet-mod-otp.

PR #170, opened 2020-06-30
Resolve bad moment tz import

landonreed:
re #152

### DIFF
--- a/packages/core-utils/src/__tests__/__snapshots__/time.js.snap
+++ b/packages/core-utils/src/__tests__/__snapshots__/time.js.snap
@@ -1,5 +1,3 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`time getCurrentTime should return time at specified timezone 1`] = `"22:34"`;
-
-exports[`time getUserTimezone should return user's timezone 1`] = `undefined`;
--- a/packages/core-utils/src/__tests__/__snapshots__/time.js.snap
+++ b/packages/core-utils/src/__tests__/__snapshots__/time.js.snap
@@ -1,0 +1,5 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`time getCurrentTime should return time at specified timezone 1`] = `"22:34"`;
+
+exports[`time getUserTimezone should return user's timezone 1`] = `undefined`;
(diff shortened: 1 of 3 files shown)


packages/core-utils/src/__tests__/time.js
```diff
@@ -0,0 +1,57 @@
+import { getCurrentTime } from "../time";
+
+// Sun Aug 04 2019 19:34:56 GMT-0700
+const DEFAULT_TEST_TIME = Date.UTC(2019, 7, 5, 2, 34, 56, 78);
+
+export function timeoutPromise(ms) {
```
evansiroky, line 6: Not needed.

packages/core-utils/src/__tests__/time.js
```diff
@@ -0,0 +1,57 @@
...
+ * Temporarily change the internal behavior of the Date.now method such that it
+ * returns a time that is based off of the given value in milliseconds after
+ * the epoch. Typically the method Date.UTC(YYYY, MM, DD) can be used to
+ * generate this number.
+ *
+ * Note: this stack overflow page gives more info on why we're using this:
+ * https://stackoverflow.com/a/42787232/915811 (basically, moment.js uses
+ * Date#now internally).
+ */
+export function setTestTime(time) {
```
evansiroky, line 23: We probably shouldn't export these from here.

packages/core-utils/src/__tests__/__snapshots__/time.js.snap
```diff
@@ -0,0 +1,5 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`time getCurrentTime should return time at specified timezone 1`] = `"22:34"`;
+
+exports[`time getUserTimezone should return user's timezone 1`] = `undefined`;
```
evansiroky, line 5: Not needed.

packages/core-utils/src/time.js
```diff
@@ -95,9 +94,9 @@ export function getUserTimezone() {
  * Formats current time for use in OTP query
  * The conversion to the user's timezone is needed for testing purposes.
  */
-export function getCurrentTime() {
+export function getCurrentTime(timezone = getUserTimezone()) {
```
evansiroky, line 97: Since we're adding this functionality to this method, it seems like it'd make sense to add it to `getCurrentDate` too.

PR #173, opened 2020-06-30
Remove config param from getDefaultQuery

landonreed:
Remove config param from getDefaultQuery.


landonreed:
Assigned to @evansiroky to help resolve broken tests.

PR #178, opened 2020-07-02
Geocoding fixes

binh-dam-ibigroup:
This PR mirrors the changes from https://github.com/opentripplanner/otp-react-redux/pull/182 for the geocoder and core-utils objects.

### DIFF
--- a/packages/core-utils/src/query.js
+++ b/packages/core-utils/src/query.js
@@ -1,4 +1,5 @@
 import moment from "moment";
+import getGeocoder from "@opentripplanner/geocoder";
 import qs from "qs";
 
 import {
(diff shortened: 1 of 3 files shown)


packages/core-utils/src/query.js
```diff
@@ -174,21 +210,48 @@ function parseLocationString(value) {
     : null;
 }
 
+/**
+ * Convert a string query param for a from or to place into a location. If
+ * coordinates not provided and geocoder config is present, use the first
+ * geocoded result.
+ * @param  {string} value                 [description]
+ * @param  {Object} [geocoderConfig=null] [description]
+ * @return {Location}                       [description]
```
landonreed, line 219: Please remove placeholder `[description]`

packages/core-utils/src/query.js
```diff
@@ -174,21 +210,48 @@ function parseLocationString(value) {
...
  *    params.
+ * @param config the config in the otp-rr store.
  */
-export function planParamsToQuery(params) {
+export async function planParamsToQuery(params, config) {
   const query = {};
-  Object.keys(params).forEach(key => {
+  // FIXME? Must use for to execute the awaits statements one after the other
+  // and obtain a populated query object.
+  /* eslint-disable no-await-in-loop */
```
evansiroky, line 242: Of all these disablers, I think this `no-await-in-loop` matters the most here and will affect app performance since making a geocode request for both the from and to locations will happen sequentially instead of asynchronously. If we're looking to publish this change fast, then we can ignore this for now, but if we're ever looking at performance this will definitely be a hotspot that needs eventual correction.

PR #188, opened 2020-07-21
Route util refactor

evansiroky:
This PR moves and creates some code related to various items with routes such as getting a route from config and sorting routes.


packages/core-utils/src/__tests__/route.js
```diff
@@ -0,0 +1,120 @@
...
+  route14,
+  route15
+} = require("./__mocks__/itinerary.json");
+
+function sortRoutes(...routes) {
+  routes.sort(routeComparator);
+  return routes;
+}
+
+describe("util > itinerary", () => {
```
landonreed, line 30: This is not the itinerary util file

packages/core-utils/src/route.js
```diff
@@ -0,0 +1,264 @@
+/** Looks up an operator from the provided configuration */
+export function getTransitOperatorFromId(id, transitOperators) {
```
landonreed, line 2: So, I think this function needs the jsdoc to be clarified a bit. I just found a bug in SEPTA where the agencyId is being passed to this function instead of the split route ID.

PR #189, opened 2020-07-21
Refactor handling of transitOperators, route sorting

evansiroky:
This PR does 4 things:

- refactors all functions in the coreUtils.route that deal with transitOperators. As was alluded to in https://github.com/ibi-group/OpenTripPlanner/pull/49, the feedId and agencyId must be used together in order to correctly identify a specific transit agency. Therefore, this will include a breaking change that will require all configs with `transitOperators` entries to be rewritten to include an `agencyId` and `feedId` instead of simply an `id` field.
- Adds a criteria to route sorting to use data about agencies. This resulted in a breaking change refactor of the route comparator.
- Instead of displaying an agency name for a stop as suggested in https://github.com/opentripplanner/otp-ui/issues/190, this removes the agency name entirely from the display at a stop. Multiple agencies can exist in a GTFS, so without analyzing all stopTimes, it's impossible to know which agencies in a multi-agency GTFS stops at a stop.
- adds a helper script that will update all internal dependencies.


scripts/update-internal-dependencies.js
```diff
@@ -0,0 +1,75 @@
...
+ * package.json to the latest version.
+ */
+async function main() {
+  // iterate through all packages to build up dependency tree
+  const packages = await fs.readdir("./packages");
+  const pkgLookup = {};
+  await Promise.all(
+    packages.map(async pkg => {
+      // skip paths that aren't directories
+      if (!(await fs.stat(`./packages/${pkg}`)).isDirectory()) return;
```
landonreed, line 36: Minor semantic thing (non-blocking), but why not nest the pkgLookup code below in a check for isDirectory, rather than skipping non-directories? For some reason, doing it this way seems clearer.

scripts/update-internal-dependencies.js
```diff
@@ -0,0 +1,75 @@
...
+}
+
+/**
+ * Updates all internal @opentripplanner/* dependencies in each package's
+ * package.json to the latest version.
+ */
+async function main() {
+  // iterate through all packages to build up dependency tree
+  const packages = await fs.readdir("./packages");
+  const pkgLookup = {};
```
landonreed, line 32: Should have a comment mentioning that pkgLookup stores `package.json` as parsed json for each pkg name.

scripts/update-internal-dependencies.js
```diff
@@ -0,0 +1,75 @@
...
+  );
+  // iterate again and replace all internal package dependencies that have a
+  // newer @opentripplanner dependency
+  await Promise.all(
+    Object.keys(pkgLookup).map(async pkg => {
+      // iterate through lines of package.json and update dependencies as needed
+      const newPackageJsonLines = [];
+      const pkgPackageJsonFile = `./packages/${pkg}/package.json`;
+      const pkgContents = await fs.readFile(pkgPackageJsonFile, "UTF-8");
+      pkgContents.split("\n").forEach(line => {
```
landonreed, line 50: This whole block needs more comments.

packages/itinerary-body/src/place-row.js
```diff
@@ -107,7 +104,10 @@ const PlaceRow = ({
                 TransitLegSummary={TransitLegSummary}
                 transitOperator={
                   leg.agencyId &&
-                  getTransitOperatorFromConfig(leg.agencyId, config)
+                  coreUtils.route.getTransitOperatorFromId(
+                    leg.agencyId,
```
landonreed, line 108: I think this should be using `coreUtils.route.getTransitOperatorFromOtpRoute(leg.routeId, config.transitOperators)`. The input arg for `getTransitOperatorFromId` is `feedId` not `agencyId`. This is an undocumented bug in this component. Related bug that could be addressed in this PR: https://github.com/opentripplanner/otp-ui/issues/190

evansiroky:
I'm reassigning to myself to do some more work on route sorting.

evansiroky:
Ready for review again.

scripts/update-internal-dependencies.js
```diff
@@ -0,0 +1,85 @@
...
+//   `yarn update-internal-dependencies core-utils base-map`
+//      updates packages that depend on either @opentripplanner/core-utils or
+//      @opentripplanner/base-map
+const filteredPackages = process.argv.slice(2);
+if (filteredPackages.length === 0) {
+  console.log("Updating all internal @opentripplanner dependencies!");
+}
+
+/**
+ * Updates all internal @opentripplanner/* dependencies in each package's
```
binh-dam-ibigroup, line 21: Who or what should normally trigger this new script? The user or the CI build?

scripts/update-internal-dependencies.js
binh-dam-ibigroup: Also, for future reference, is the expectation that we check-in `package.json` files and `yarn.lock` with the dependencies that already account for the pending changes?

binh-dam-ibigroup (approved):
LGTM. We'll need some guidance on how/when to use `yarn update-internal-dependencies`.

scripts/update-internal-dependencies.js
evansiroky: This should be manually run by a user.

landonreed: I think this should be captured in comments and also documented in the README in a new section about Releases.

scripts/update-internal-dependencies.js
```diff
@@ -0,0 +1,85 @@
...
+//
+// For example:
+//   `yarn update-internal-dependencies`
+//      updates everything
+//   `yarn update-internal-dependencies core-utils base-map`
+//      updates packages that depend on either @opentripplanner/core-utils or
+//      @opentripplanner/base-map
+const filteredPackages = process.argv.slice(2);
+if (filteredPackages.length === 0) {
+  console.log("Updating all internal @opentripplanner dependencies!");
```
landonreed, line 17: Is this a bad thing? The exclamation mark makes me think it's not something we should be doing. I think @binh-dam-ibigroup and my uncertainty about how this works speaks to the need to document how this should be used in the README.

scripts/update-internal-dependencies.js
```diff
@@ -0,0 +1,85 @@
...
+      // iterate through lines of package.json and update dependencies as needed
+      const newPackageJsonLines = [];
+      const pkgPackageJsonFile = `./packages/${pkg}/package.json`;
+
+      // read package's package.json file line-by-line
+      const pkgContents = await fs.readFile(pkgPackageJsonFile, "UTF-8");
+      pkgContents.split("\n").forEach(line => {
+        // try to find a match of an internal dependency
+        const match = line.match(
+          /\s*"@opentripplanner\/([\w-]*)": "\^?([\d.]*)"/
```
landonreed, line 52: Will this regex miss some instances? E.g., `@opentripplanner/foo: latest`?

packages/core-utils/src/__tests__/__snapshots__/route.js.snap
```diff
@@ -135,6 +156,27 @@ Array [
...
+    "mode": "RAIL",
+    "routeId": "1:it2",
+    "shortName": "IC",
+  },
+  Object {
+    "agencyId": "abc",
+    "agencyName": "Acclaimed agency 1",
+    "longName": "Express route X",
+    "mode": "BUS",
+    "routeId": "2:erx",
```
landonreed, line 174: Is this test working as expected? It seems like route `2:erx` should come before route `1:it2`?

packages/core-utils/src/route.js
```diff
@@ -1,31 +1,151 @@
...
+// https://github.com/opentripplanner/otp-react-redux/pull/124.
+
 /**
- * The functions below are for enhanced route sorting functions for
- * the route viewer on OTP-react-redux.
- * They address route ordering issues discussed in
- * https://github.com/opentripplanner/otp-react-redux/pull/123 and
- * https://github.com/opentripplanner/otp-react-redux/pull/124.
+ * A large comparator value that can safely be used in mathematical sort
+ * comparisons to place things at the ned of lists
```
landonreed, line 83: typo "ned"

packages/core-utils/src/route.js
```diff
@@ -1,31 +1,151 @@
...
+    leg.agencyId,
+    transitOperators
+  );
+}
+
+/**
+ * Looks up an operator from the provided configuration given an OTP route.
+ * NOTE: this assumes the use of the OTP Route model or a modified OTP
+ * RouteShort model (such as the one found in the IBI fork of OTP) that also
+ * returns the agencyId.
```
landonreed, line 49: Is there a PR to the main OTP repo to add the agencyId field? If not, we should prioritize that.

packages/core-utils/src/route.js
```diff
@@ -1,31 +1,151 @@
...
+    transitOperators
+  );
+
+  // if transit operator not found, return infinity
+  if (!transitOperator) return END_OF_LIST_COMPARATOR_VALUE;
+
+  // return the transit operator's sort value or END_OF_LIST_COMPARATOR_VALUE if
+  // the sort value is not defined
+  return typeof transitOperator.order !== "undefined"
+    ? transitOperator.order
```
landonreed, line 124: Do we care if the order field type is a number (vs. string)? Probably not, but just flagging.

packages/core-utils/src/route.js
```diff
@@ -261,11 +381,14 @@ function makeMultiCriteriaSort(...criteria) {
...
  */
-export const routeComparator = makeMultiCriteriaSort(
-  makeNumericValueComparator(obj => getRouteSortOrderValue(obj.sortOrder)),
-  routeTypeComparator,
-  alphabeticShortNameComparator,
-  makeNumericValueComparator(obj => parseInt(obj.shortName, 10)),
-  makeStringValueComparator(obj => obj.shortName),
-  makeStringValueComparator(obj => obj.longName)
-);
+export function makeRouteComparator(transitOperators) {
```
landonreed, line 384: I'm not sure the previous jsdoc exactly matches this new function.

landonreed (changes requested):
Looks generally good, just a few comments.

packages/core-utils/src/__tests__/__snapshots__/route.js.snap
evansiroky: Yes, it is working as expected. In this test method, the route comparator is sent in a config that says that feed_id 2 should be prioritized over feed_id 1. https://github.com/ibi-group/otp-ui/blob/a87a4e26bc4d961d787207ff3f850b44929b6b77/packages/core-utils/src/__tests__/route.js#L71-L78

packages/core-utils/src/route.js
evansiroky: It is included in https://github.com/ibi-group/OpenTripPlanner/pull/49. See https://github.com/ibi-group/OpenTripPlanner/pull/49/files#diff-139cb730f647135718a77f533420f7c5d731ac996a9e32c8bae292184c68ec87.

scripts/update-internal-dependencies.js
evansiroky: Yes it might, but this is unlikely to matter. If something is set that way, users will have to manually change that instead of using the script which is fine.

packages/core-utils/src/route.js
```diff
@@ -1,31 +1,151 @@
...
+  agencyId,
+  transitOperators
+) {
   return (
-    transitOperators.find(transitOperator => transitOperator.id === feedId) ||
-    null
+    transitOperators.find(
+      transitOperator =>
+        transitOperator.feedId === feedId &&
+        transitOperator.agencyId === agencyId
```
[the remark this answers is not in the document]
evansiroky: There is a way to do this, but since we don't auto-generate docs, I'm not sure it adds much value. The JSDoc comment already says that this is a value from the transitOperators list in the config.

packages/core-utils/src/route.js
evansiroky: Fixed in 5163a219c2ab5072d9338041435293fda8877c42

packages/core-utils/src/route.js
evansiroky: Changed the check from being undefined to being a number in 5163a219c2ab5072d9338041435293fda8877c42

packages/core-utils/src/route.js
evansiroky: Fixed in 5163a219c2ab5072d9338041435293fda8877c42

scripts/update-internal-dependencies.js
evansiroky: Removed scary exclamation point in 5163a219c2ab5072d9338041435293fda8877c42

scripts/update-internal-dependencies.js
evansiroky: I modified the project's README in 5163a219c2ab5072d9338041435293fda8877c42

evansiroky:
Hopefully the latest changes satisfy all concerns. Regarding the `update-internal-dependencies` script, I made this thing because it helps update all the other packages when the `core-utils` package gets updated. Prior to this, we'd have to manually go through each package and manually update each line. This script merely automates that process. It isn't intended to be anything fancy and you don't even need to use it if you don't want to.

landonreed (approved):

PR #192, opened 2020-07-30
Itinerary Body: TimeColumnContent slot

### LINKED ISSUE #191: Need to customize leg time display.
There is a need to customize how the contents in the time column is rendered in `ItineraryBody` > `PlaceRow`
(e.g. change the color of the time, add other markup/tooltips around the displayed time, etc.).

image

binh-dam-ibigroup:
This PR fixes #191 and introduces a slot for custom rendering of the contents in the time column of the itinerary component.


binh-dam-ibigroup:
Example time column slot updated with suggestions from @landonreed (no more clock...)

PR #194, opened 2020-07-31
Relax proptypes for LocationIcon

landonreed:
This PR relaxes proptypes for LocationIcon. If `from` or `to` is not used for `locationType`, a generic Circle icon will be rendered. This will allow for an icon to be used with IntermediatePlaces in the Call Taker module.

### DIFF
--- a/packages/location-icon/src/LocationIcon.story.js
+++ b/packages/location-icon/src/LocationIcon.story.js
@@ -5,10 +5,6 @@ import { withInfo } from "@storybook/addon-info";
 import styled from "styled-components";
 import LocationIcon from ".";
 
-const StyledLocationIcon = styled(LocationIcon)`
-  color: blue;
-`;
-
 storiesOf("LocationIcon", module)
   .addDecorator(withA11y)
   .addDecorator(withInfo)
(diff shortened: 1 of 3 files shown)


packages/location-icon/src/LocationIcon.story.js
```diff
@@ -17,8 +13,14 @@ storiesOf("LocationIcon", module)
...
-  .add("To LocationIcon test", () => (
-    <StyledLocationIcon type="to" size={25} />
-  ));
+  .add("'From' LocationIcon", () => <LocationIcon type="from" size={25} />)
+  .add("'To' LocationIcon", () => <LocationIcon type="to" size={25} />)
+  .add("Generic Place LocationIcon", () => (
+    <LocationIcon type="intermediate-place-1" size={25} />
+  ))
+  .add("Custom style for 'To' LocationIcon", () => {
+    const StyledLocationIcon = styled(LocationIcon)`
```
binh-dam-ibigroup, line 22: Normally you'd follow [this advice](https://styled-components.com/docs/basics#define-styled-components-outside-of-the-render-method) but it's probably fine for rendering one story component.

PR #198, opened 2020-08-05
Use the StandardRail icon for the rail mode

evansiroky:
The StandardRail icon wasn't used in the StandardModeIcon for the `rail` mode. This PR fixes that.


binh-dam-ibigroup:
For info: The only reason `StandardRail` was not used here before is because it wasn't used in here: https://github.com/opentripplanner/otp-react-redux/blob/381c3f526368bbfdc1666355dfb1cca8eeae35c9/lib/components/icons/mode-icon.js#L14.

Issue #203, opened 2020-08-13
Intermediate Place parsing from query params


landonreed:
There is an issue with the query param parser (and stringifier) for intermediate places. It splits (and joins) with a comma delimiter. The splitting causes issues because the place name often has commas (not to mention lat,lon). Joining with a comma may also not be appropriate because it is unclear how the OTP API accepts multiple entries for the `intermediatePlaces` query param.

### split
https://github.com/opentripplanner/otp-ui/blob/69d800d7d3538bf86e26f0882660774623819175/packages/core-utils/src/query.js#L270-L275

### join
https://github.com/opentripplanner/otp-ui/blob/69d800d7d3538bf86e26f0882660774623819175/packages/core-utils/src/query-params.js#L640-L644

PR #210, opened 2020-08-25
Properly parse and stringify intermediate places

evansiroky:
Fixes https://github.com/opentripplanner/otp-ui/issues/203.

I marked this PR as a fix, because it's unlikely that anybody ever used this before since it couldn't have worked very well. This PR changes the assumption about how query parameters for the `intermediatePlaces` are generated in that they are not expected to be parsed from (or stringified into) a comma-separated string, but instead should be parsed from (or stringified into) an array of strings. The `query.getRoutingParams` and `query.planParamsToQuery` are both assumed to use the [qs](https://www.npmjs.com/package/qs) package to do the parsing (and stringifying) from (and to) raw URL query params. This package has a way of parsing (and stringifying into) query params that have multiple of the same parameter which is the format that OTP expects. Therefore, this change will result in properly parsing a query string such as `...&intermediatePlaces=name::###,###&intermediatePlaces=name2::###,###&...`. This util library does not have a util function for stringifying the params, but the way to do that with the `qs` package would be to write `qs.stringify(params, { arrayFormat: 'repeat' })`

### DIFF
--- a/packages/core-utils/src/query.js
+++ b/packages/core-utils/src/query.js
@@ -271,7 +271,7 @@ export function planParamsToQuery(params) {
         // If query has intermediate places, ensure that they are parsed
         // as locations.
         query.intermediatePlaces = params.intermediatePlaces
-          ? params.intermediatePlaces.split(",").map(parseLocationString)
+          ? params.intermediatePlaces.map(parseLocationString)
           : [];
         break;
       default: {
--- a/packages/core-utils/src/__tests__/__snapshots__/query.js.snap
+++ b/packages/core-utils/src/__tests__/__snapshots__/query.js.snap
@@ -33,6 +33,98 @@ Object {
 }
 `;
 
+exports[`query getRoutingParams should create routing params for car rental query 1`] = `
+Object {
+  "arriveBy": false,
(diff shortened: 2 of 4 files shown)


packages/core-utils/src/query.js
```diff
@@ -271,7 +271,7 @@ export function planParamsToQuery(params) {
         // If query has intermediate places, ensure that they are parsed
         // as locations.
         query.intermediatePlaces = params.intermediatePlaces
-          ? params.intermediatePlaces.split(",").map(parseLocationString)
```
binh-dam-ibigroup, line 274: This is definitely easier to deal with.

evansiroky:
> Great stuff. Non-blocking question: Should we take this opportunity to add a `stringifyPlanParams` function to this library that gets imported by otp-rr? That would give us a nice opportunity to add a test/snapshot to verify.

I don't think that makes sense as the process of stringifying query parameters is becoming a little more involved. See https://github.com/opentripplanner/otp-react-redux/blob/3f3a9a78973f51d28e555fecd74e2c9c72018b33/lib/actions/api.js#L231-L236

PR #211, opened 2020-09-03
Automatically add Mapbox wordmark element

evansiroky:
This changes the base-map package so that it will automatically display the required Mapbox wordmark attribution when the base layer has a mapbox tile URL. Mapbox requires both a text attribution and a wordmark (see https://docs.mapbox.com/help/how-mapbox-works/attribution/#other-mapping-frameworks). Since the CSS for the Mapbox wordmark is somewhat large (12kb), it is included in the example, but implementing applications will have to include it on their own if they plan on using Mapbox tiles.


packages/base-map/__mocks__/mapbox-wordmark.css
```diff
@@ -0,0 +1,21 @@
+/**
```
binh-dam-ibigroup, line 1: This file is not a mock and should be in a `packages/base-map/css` folder or a resource folder that is not the mocks folder.

packages/base-map/__mocks__/mapbox-wordmark.css
```diff
@@ -0,0 +1,21 @@
...
...Fzcz0ic3QwIiBkPSJNMTAsMS4yMWMtNC44NywwLTguODEsMy45NS04LjgxLDguODFzMy45NSw4LjgxLDguODEsOC44MXM4LjgxLTMuOTUsOC44MS04LjgxQzE4LjgxLDUuMTUsMTQuODcsMS4yMSwxMCwxLjIxeiBNMTQuMTgsMTIuMTljLTEuODQsMS44NC00LjU1LDIuMi02LjM4LDIuMmMtMC42NywwLTEuMzQtMC4wNS0yLTAuMTVjMCwwLTAuOTctNS4zNywyLjA0LTguMzljMC43OS0wLjc5LDEuODYtMS4yMiwyLjk4LTEuMjJjMS4yMSwwLDIuMzcsMC40OSwzLjIzLDEuMzVDMTUuOCw3LjczLDE1Ljg1LDEwLjUsMTQuMTgsMTIuMTl6Ii8+PHBhdGggY2xhc3M9InN0MSIgZD0iTTEwLDAuMDJjLTUuNTIsMC0xMCw0LjQ4LTEwLDEwczQuNDgsMTAsMTAsMTBzMTAtNC40OCwxMC0xMEMxOS45OSw0LjUsMTUuNTIsMC4wMiwxMCwwLjAyeiBNMTAsMTguODNjLTQuODcsMC04LjgxLTMuOTUtOC44MS04LjgxUzUuMTMsMS4yLDEwLDEuMnM4LjgxLDMuOTUsOC44MSw4LjgxQzE4LjgxLDE0Ljg5LDE0Ljg3LDE4LjgzLDEwLDE4LjgzeiIvPjxwYXRoIGNsYXNzPSJzdDEiIGQ9Ik0xNC4wNCw1Ljk4Yy0xLjc1LTEuNzUtNC41My0xLjgxLTYuMi0wLjE0QzQuODMsOC44Niw1LjgsMTQuMjMsNS44LDE0LjIzczUuMzcsMC45Nyw4LjM5LTIuMDRDMTUuODUsMTAuNSwxNS44LDcuNzMsMTQuMDQsNS45OHogTTExLjg4LDkuODdsLTAuODcsMS43OGwtMC44Ni0xLjc4TDguMzgsOS4wMWwxLjc3LTAuODZsMC44Ni0xLjc4bDAuODcsMS43OGwxLjc3LDAuODZMMTEuODgsOS44N3oiLz48cG9seWdvbiBjbGFzcz0ic3QwIiBwb2ludHM9IjEzLjY1LDkuMDEgMTEuODgsOS44NyAxMS4wMSwxMS42NSAxMC4xNSw5Ljg3IDguMzgsOS4wMSAxMC4xNSw4LjE1IDExLjAxLDYuMzcgMTEuODgsOC4xNSAiLz48L2c+PC9zdmc+);
```
binh-dam-ibigroup, line 17: This mapbox logo is fine, although it is not as dark in recent mapbox implementations (e.g. https://docs.mapbox.com/mapbox-gl-js/api/)

(Deep inside, I'd like us to use the lighter mapbox logo!)

packages/base-map/src/index.story.js
```diff
@@ -10,6 +10,7 @@ import AllVehiclesOverlay from "../__mocks__/AllVehicles";
 import ContextMenuDemo from "../__mocks__/ContextMenuDemo";
 
 import "../../../node_modules/leaflet/dist/leaflet.css";
+import "../__mocks__/mapbox-wordmark.css";
```
binh-dam-ibigroup, line 13: Can this be imported inside `index.js` instead? That seems a better place because the wordmark assumes a `mapbox-wordmark` defined in that CSS, and it would avoid implementers to import that CSS file.

binh-dam-ibigroup (changes requested):
I would prefer to have the mapbox CSS file not in the `__mocks__` folder and have it imported inside `index.js` (see my comments). Looks good otherwise!

packages/base-map/__mocks__/mapbox-wordmark.css
evansiroky: See PR description:

> Since the CSS for the Mapbox wordmark is somewhat large (12kb), it is included in the example

If we added it in the example index.js file, it would always be imported and add 12kb (unzipped) of unnecessary data downloads even if the implementing application didn't use a Mapbox base layer (for example, TriMet doesn't).

packages/base-map/src/index.story.js
evansiroky: See PR description:

> Since the CSS for the Mapbox wordmark is somewhat large (12kb), it is included in the example

If we added it in the example index.js file, it would always be imported and add 12kb (unzipped) of unnecessary data downloads even if the implementing application didn't use a Mapbox base layer (for example, TriMet doesn't).

packages/base-map/__mocks__/mapbox-wordmark.css
evansiroky: This background image is taken directly from [Mapbox's site here](https://docs.mapbox.com/help/how-mapbox-works/attribution/#other-mapping-frameworks).

packages/base-map/src/index.story.js
binh-dam-ibigroup: Fair point.

binh-dam-ibigroup (approved):

packages/base-map/src/index.js
```diff
@@ -74,10 +74,25 @@ L.Evented.include({
 class BaseMap extends Component {
   overlays = [];
 
+  constructor(props) {
+    super(props);
+    const { baseLayers } = props;
+    if (baseLayers && baseLayers.length > 0) {
+      this.state = this.getMapboxWordmarkUpdate(baseLayers[0].url);
```
landonreed, line 81: If we're adding state to this component, it seems like the more appropriate/extensible way to handle would be to
- initialize the `layerIndex` to zero
- set `LayersControl.BaseLayer#checked` below to `i === this.state.layerIndex`
- remove the `baselayerchange` listener in `componentDidMount` in favor of adding the `onBaseLayerChange` prop to the `<Map>` component
-  add a `handleBaseLayerChange` class method that updates internal state layerIndex.

This is a lot to describe... so I just made a PR: https://github.com/ibi-group/otp-ui/pull/4.

landonreed (changes requested):
I think the approach should follow what I've added in: https://github.com/ibi-group/otp-ui/pull/4

The current approach feels a bit too hyper-specific for this particular use case. With a few more lines of code, we can open up the leaflet API a little bit and make things more future proof (in my view).

evansiroky:
Yeah, that seems better to use react-leaflet's props for detecting a base layer change.

PR #212, opened 2020-09-11
Handle error response from geocoder

### LINKED ISSUE #183: LocationField errors out on bad geocoding response
If a bad geocoder response is returned from (e.g. pelias), the LocationField component errors out instead of handling the bad response.

The best fix might be to catch the error in the geocoder package and just return an empty list of features if an error is encountered.

The error itself is thrown here (undefined error on `geocodedFeatures.length`): https://github.com/opentripplanner/otp-ui/blob/a61a66ad4a479aad5a02b80800f4f510b1492f24/packages/location-field/src/index.js#L318-L319

landonreed:
This PR fixes #183 by correctly handling when there are no features returned from a geocode response. It also adds a message if there is an error communicating with the geocoder or if no results are found, which I think is useful behavior (otherwise there is no indication that a search completed). Curious to have @fpurcell's thoughts on that behavior though.

A new story with a bad API key configured for geocode.earth is at http://localhost:5555/?path=/story/locationfield--locationfield-with-bad-api-key-handles-bad-autocomplete

### DIFF
--- a/packages/location-field/src/index.js
+++ b/packages/location-field/src/index.js
@@ -11,7 +11,14 @@ import LocationIcon from "@opentripplanner/location-icon";
 import PropTypes from "prop-types";
 import React, { Component } from "react";
 import ReactDOM from "react-dom";
-import { Ban, Bus, LocationArrow, Search, Times } from "styled-icons/fa-solid";
+import {
+  Ban,
+  Bus,
+  ExclamationCircle,
+  LocationArrow,
+  Search,
+  Times
+} from "styled-icons/fa-solid";
 import { throttle } from "throttle-debounce";
 
 import {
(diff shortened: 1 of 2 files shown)


packages/location-field/src/index.js
```diff
@@ -525,6 +553,17 @@ class LocationField extends Component {
...
+      const messageItem = (
+        <Option
+          icon={<ExclamationCircle size={20} />}
+          key={optionKey++}
+          title={message}
+          disabled
+        />
+      );
+      menuItems.unshift(messageItem);
+    }
```
evansiroky, line 566: Since this always shows up with a exclamation circle icon, perhaps it would be better to prefix this variable and name it `errorMessage` and `errorMessageItem` respectively.

evansiroky (approved):
Just one non-blocking comment

packages/location-field/src/index.js
```diff
@@ -43,7 +50,25 @@ class LocationField extends Component {
...
+          // Get the Pelias error message if exists.
+          // TODO: determine how other geocoders return error messages.
+          const errorMessage =
+            result &&
+            result.results &&
+            result.results.error &&
+            result.results.error.message;
+          // If the result did not contain a list of features, add special note.
+          message = "Could not reach geocoder";
+          if (errorMessage) message += ` (${errorMessage})`;
```
binh-dam-ibigroup, line 66: Using a templated string for the entire message would have been better.

binh-dam-ibigroup (approved):
Please see the code suggestion, but it is good to go nonetheless.

fpurcell (approved):
Such geocoder errors were a problem in trimet.org ... the geocoder was diaper wrapped to avoid certain exceptions that would throw the Pelias component into a tizzy.  Not sure how your changes will affect that, Landon (my guess is it will be fine).  AltSource mentioned wanting something similar to what you've implemented, ala:

packages/location-field/src/index.js
landonreed: The "no results found" message isn't really an error, so I don't think we necessarily should be calling it an error message.

packages/location-field/src/index.js
landonreed: This seems about as legible as (maybe more than) that.

Issue #215, opened 2020-10-24
ZoomControl and LayerControl location fed into BaseMap props


AvidDabbler:
I have found a way to configure the ZoomControl and LayerControl location as additional props for the BaseMap component in @opentripplanner/base-map/lib/index.js file, but I am looking at the file structure of the the repo and this file is not there. 

Is there a build step that is just don't know about before it heads off to NPM.

Excuse me for my ignorance, this is my first attempt at making a contribution to an Open Source project.

PR #216, opened 2020-10-24
added in layersPosition and zoomPosition props

AvidDabbler:
Attempt to add positioning support for ZoomControl and LayersControl to BaseMap via component props.
#215 

Was not able to test due to not knowing how to compile and test.

Not sure if the disabling zoomControl in the Map component and then enabling via ZoomControl component is the best practice, but when I did initial testing by just adding in ZoomControl component it just added another ZoomControl. As far as I know the default Map zoomControl props is just a boolean value and you have to call ZoomControl in order to access to the position setting. 

EDIT:
For reference while I was not able to compile I am using @opentripplanner via otp-react

LINKS:
ZoomControl > https://react-leaflet.js.org/docs/en/components#zoomcontrol
LayersControl > https://react-leaflet.js.org/docs/en/components#layerscontrol
Map zoomControl > https://leafletjs.com/reference-1.6.0.html#map-zoomcontrol

### DIFF
--- a/packages/base-map/src/index.js
+++ b/packages/base-map/src/index.js
@@ -1,6 +1,6 @@
 import React, { Component } from "react";
 import PropTypes from "prop-types";
-import { LayersControl, Map, Popup, TileLayer } from "react-leaflet";
+import { LayersControl, Map, Popup, TileLayer, ZoomControl } from "react-leaflet";
 import utils from "@opentripplanner/core-utils";
 import L from "leaflet";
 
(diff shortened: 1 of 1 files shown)


AvidDabbler:
Hey. I know that this isn't passing. Is there a reason why this is not being addressed? Did I push it to the wrong repo? Do you all not except outside pull requests?

landonreed:
Hi @AvidDabbler, thanks for the contribution. Typically this project's pull request priorities are driven by the implementations/customers that we work with first hand, and we aren't always the most responsive when things fall out of that process. Apologies!

It seems like the underlying problem you're looking to resolve is that the layers and zoom controls are fixed in their default positions. We'll see if we can squeeze this into our review process. If this seems like the right approach, we'll advise on resolving the failed tests/linting, or if we think another approach might be better/more flexible we'll post in review comments.

PR #217, opened 2020-10-27
Revert to old transitOperator lookup

evansiroky:
Changes were added in #189 that resulted in a new v3 release of core-utils and some releases on other packages. One of those packages was the itinerary-body package, but it hadn't yet been updated to have an official dependency on core-utils v3. Therefore, this introduced code in the 1.x.x pipeline of the itinerary-body package that is now a bug. This PR reverts changes that assumed core-utils v3 within the itinerary-body package to maintain consistency with coreUtils 2.x and the previous way of defining transitOperators in the config. This PR will create a patch release of the itinerary-body package.

Another PR will change this code back again to the v3 style transit operators config and will contain a commit with a breaking change that will make sure the itinerary-body package gets a proper bump. That will result in a itinerary-body v2.x.x bump to comply with semver.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2746,6 +2746,20 @@
   dependencies:
     "@types/node" ">= 8"
 
+"@opentripplanner/core-utils@^2.1.0":
+  version "2.1.2"
+  resolved "https://registry.yarnpkg.com/@opentripplanner/core-utils/-/core-utils-2.1.2.tgz#a19d5d788704f0a6c2aece5206a2c8997c251d16"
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -2746,6 +2746,20 @@
   dependencies:
     "@types/node" ">= 8"
 
+"@opentripplanner/core-utils@^2.1.0":
```
binh-dam-ibigroup, line 2749: Should this be deduplicated?

binh-dam-ibigroup (approved):
A deduplicate of the addition to yarn.lock would be ideal.

landonreed (approved):

PR #218, opened 2020-10-27
Update opentripplanner dependencies to latest versions (core-utils v3 update part 1)

evansiroky:
This is part 1 of 3 to update internal packages to use core-utils v3. This PR only updates the packages that have no dependencies on otp-ui packages that have a dependency on core-utils. If we were to include packages such as vehicle-rental-overlay in this PR, it would not be sufficient to ensure that all other packages that the vehicle-rental-overlay package depends on also have a dependency on core-utils v3. So, future PRs will bump core-utils and other relevant packages with a core-utils dependency until every package that has some kind of dependency on core-utils has been released. In doing so, there will be only one release per package which will look much cleaner than multiple releases that update a single package at a time. Hope I explained this well and this is ok.

### DIFF
--- a/packages/base-map/package.json
+++ b/packages/base-map/package.json
@@ -8,7 +8,7 @@
   "main": "lib/index.js",
   "private": false,
   "dependencies": {
-    "@opentripplanner/core-utils": "^2.1.0",
+    "@opentripplanner/core-utils": "^3.0.0",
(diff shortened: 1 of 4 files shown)


binh-dam-ibigroup (approved):
Seems to work.

evansiroky:
@landonreed gave approval via company chat.

PR #223, opened 2020-11-11
fix(ItineraryBody): Allow PlaceRow to take up entire width available.

binh-dam-ibigroup:
This PR partially addresses https://github.com/ibi-group/trimet-mod-otp/issues/282 by allowing `PlaceRow` to fill all the available space to the right, as shown in the screenshot below.

image

### DIFF
--- a/packages/itinerary-body/src/place-row.js
+++ b/packages/itinerary-body/src/place-row.js
@@ -82,6 +82,18 @@ const PlaceRow = ({
             <Styled.PlaceName>
               <PlaceName config={config} interline={interline} place={place} />
             </Styled.PlaceName>
+
+            {showMapButtonColumn && (
+              <Styled.MapButtonColumn hideBorder={hideBorder.toString()}>
+                <Styled.MapButton
+                  onClick={() =>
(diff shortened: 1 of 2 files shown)


evansiroky (changes requested):
Some of the storybook examples now look a little strange and don't play nicely with some storybook elements. Can you update some stories so no elements are hidden?

Screen Shot 2020-11-11 at 12 54 15 PM

packages/itinerary-body/src/place-row.js
```diff
@@ -126,15 +138,6 @@ const PlaceRow = ({
             ))}
         </Styled.PlaceDetails>
       </Styled.DetailsColumn>
-      {showMapButtonColumn && (
```
evansiroky, line 129: This changes the layout so that the MapButton will be confined within the DetailsColumn. I think this should be reverted. See proposed way of doing things here: https://github.com/ibi-group/otp-ui/pull/5.

landonreed:
Closed in favor of https://github.com/opentripplanner/otp-react-redux/pull/279

PR #231, opened 2021-01-09
Replace TravisCI with GitHub actions

evansiroky:
The opentripplanner Travis organization has ran out of credits and is only giving out credits for free to organizations that don't use any paid software developers... so we're switching to GitHub Actions.


.github/workflows/node-ci.yml
```diff
@@ -0,0 +1,35 @@
+name: Node.js CI
+
+on:
+  push:
+    branches:
+      - master
+      - alpha
+      - beta
```
landonreed, line 8: Are we sure we want to limit these branches (given that semantic-release allows for [a wider set of branches](https://semantic-release.gitbook.io/semantic-release/v/beta/usage/configuration#branches) from which to release)?

landonreed (approved):
Looks good. Just a couple of minor questions:
- Should we increase the set of branches to match semantic-release's default?
- Any reason why the spellcheck action was excluded?

binh-dam-ibigroup (approved):
LGTM, and what @landonreed said about spell check.

evansiroky:
I added some extra branches. I also added the codespell action and fixed all the misspellings it found. I'm going to go ahead and merge now.

PR #235, opened 2021-02-24
Update MapIcon hover title tag and hover cursor.

noletorious:


landonreed (changes requested):
Thanks @noletorious! That's definitely much friendlier label for that icon.

A couple of comments:
- for future patches, be sure to use [commit message format preferred by semantic-release](https://github.com/semantic-release/semantic-release#commit-message-format), this will ensure that patches get released as a new version (I'll work to update the contributor docs with this)
- Also, a link to storybook can definitely be helpful in tracking down where changes were made (in this case it wasn't too hard!)
- Finally, I've made an alternative PR (#236) that tweaks the approach a bit. Let me know if you think that approach will work

PR #236, opened 2021-02-25
Tweak map icon message

landonreed:
This PR is intended to supercede #235. It provides the following tweaks:
- moves the title prop from MapIcon to MapButton (so that a hover anywhere on the button will show the alt text)
- adds a messages prop with default values (to support future/planned internationalization).

### DIFF
--- a/packages/itinerary-body/src/place-row.js
+++ b/packages/itinerary-body/src/place-row.js
@@ -23,6 +23,7 @@ const PlaceRow = ({
   LegIcon,
   legIndex,
   LineColumnContent,
+  messages,
   PlaceName,
   RouteDescription,
   setActiveLeg,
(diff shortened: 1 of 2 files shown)


binh-dam-ibigroup (approved):
I agree with @landonreed about leaving text data out of CSS.

noletorious (approved):
Looks good!

fpurcell (approved):
Thanks for going above & beyond, Landon. 
All looks good.  See hover and new message in storybook.

landonreed:
BTW, here's the release @noletorious: https://github.com/opentripplanner/otp-ui/releases/tag/%40opentripplanner%2Fitinerary-body-v2.0.2

PR #244, opened 2021-04-07
feat(query.js): override query params w/ config#defaultQueryParams

### LINKED ISSUE #243: Cannot override default query params
There is no way currently to override the default query params (impacts `getDefaultQuery` and `isNotDefaultQuery` in `coreUtils.query`).

landonreed:
This PR closes #243 by modifying `getDefaultQuery` and `isNotDefaultQuery` to handle overriding default query params with values from `config#defaultQueryParams`.

### DIFF
--- a/packages/core-utils/src/__tests__/__mocks__/config.json
+++ b/packages/core-utils/src/__tests__/__mocks__/config.json
@@ -1,0 +1,55 @@
+{
+  "homeTimezone": "America/New_York",
+  "routingTypes": [
+    {
+      "key": "ITINERARY",
+      "text": "Exact Time"
+    }
+  ],
+  "modes": {
+    "transitModes": [
+      {
+        "mode": "BUS",
+        "label": "Bus"
+      },
+      {
+        "mode": "RAIL",
+        "label": "Rail"
+      },
+      {
+        "mode": "SUBWAY",
+        "label": "Subway"
+      },
+      {
+        "mode": "TRAM",
+        "label": "Trolleys"
+      }
+    ],
+    "accessModes": [
+      {
+        "mode": "BICYCLE",
+        "label": "Transit + Personal bike"
+      },
+      {
+        "mode": "BICYCLE_RENT",
+        "label": "Transit + Bikeshare"
+      }
+    ],
+    "exclusiveModes": ["WALK", "BICYCLE"],
+    "bicycleModes": [
+      {
+        "mode": "BICYCLE",
+        "label": "Own Bike"
+      },
+      {
+        "mode": "BICYCLE_RENT",
+        "label": "Bikeshare"
+      }
+    ]
+  },
+  "dateTime": {
+    "timeFormat": "h:mm a",
+    "dateFormat": "MM/DD/YYYY",
+    "longDateFormat": "MMMM D, YYYY"
+  }
+}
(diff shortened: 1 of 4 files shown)


packages/core-utils/src/query.js
```diff
@@ -123,55 +123,79 @@ function getDefaultQueryParamValue(param) {
...
+    let defaultValue = getDefaultQueryParamValue(paramInfo);
+    // Apply config query param overrides if it exists.
+    if (config && config.defaultQueryParams) {
+      if (param in config.defaultQueryParams) {
+        defaultValue = config.defaultQueryParams[param];
+      }
+    }
+    if (query[param] !== defaultValue) {
+      queryIsDifferent = true;
+    }
```
binh-dam-ibigroup, line 194: Is there any reason for not calling `getDefaultQuery(config)` above (less the transit params) and do an `isEqual` compare with that instead?

packages/core-utils/src/__tests__/__mocks__/config.json
```diff
@@ -0,0 +1,55 @@
...
+        "label": "Bikeshare"
+      }
+    ]
+  },
+  "dateTime": {
+    "timeFormat": "h:mm a",
+    "dateFormat": "MM/DD/YYYY",
+    "longDateFormat": "MMMM D, YYYY"
+  }
+}
```
binh-dam-ibigroup, line 55: I think there should be some `defaultQueryParams` that should be mentioned in the test.

binh-dam-ibigroup (changes requested):
I think the tests should cover `defaultQueryParams` from the mock configuration since that's what this PR is about. Also, I am not sure why `isNotDefaultQuery` does not reference `getDefaultQuery`. There should be a comment about why not, otherwise it looks more complicated than it seems.

packages/core-utils/src/query.js
landonreed: Good question! I had done that originally, but then got sidetracked by the new mode check I added. I'll swap back for `getDefaultQuery` to dry it out.

packages/core-utils/src/__tests__/__mocks__/config.json
landonreed: The tests clone/modify this default config to have the `defaultQueryParams`. I'd rather have a standard, multi-purpose config in place for multiple tests and modify that as needed for each test.

packages/core-utils/src/query.js
landonreed: Fixed in https://github.com/opentripplanner/otp-ui/pull/244/commits/b5a43ebc25f57ac38fc6419d016dd3aed35ddd23

packages/core-utils/src/query.js
```diff
@@ -171,29 +171,23 @@ export function isNotDefaultQuery(query, config) {
...
+  for (let i = 0; i < defaultParams.length; i++) {
+    const param = defaultParams[i];
+    const { applicable, routingTypes } = queryParams.find(
+      qp => qp.name === param
+    );
+    // Check that the parameter applies to the specified routingType and that
+    // the applicability test (if not missing) is satisfied.
+    const appliesToQuery =
+      typeof applicable !== "function" || applicable(query, config);
+    if (appliesToQuery && routingTypes.includes(query.routingType)) {
```
binh-dam-ibigroup, line 184: Is there a way to isolate this logic?
```js
if (isParamApplicable(param, query, config)) {...
```

binh-dam-ibigroup (changes requested):
I think one logic block should be abstracted out, then it should be good to go.

landonreed:
Good call, that's a lot cleaner.

binh-dam-ibigroup (approved):

PR #246, opened 2021-05-06
fix(TransitiveOverlay): Expose styles and labeledModes transitive props.

binh-dam-ibigroup:
For some reason I was convinced the transitive options were directly exposed but I was wrong... so we have to expose the new styles and labeledModes on the components.

EDIT: In response to https://github.com/opentripplanner/otp-ui/pull/246#discussion_r630499498, this PR now requires #248. To test, merge the changes from #248 into this branch and run `yarn dev`.

PR #248, opened 2021-05-12
Custom route labels in transitive itineraries

binh-dam-ibigroup:
This PR enables passing a custom route labeling function as suggested in https://github.com/opentripplanner/otp-ui/pull/246#discussion_r630499498 when converting OTP itineraries to transitive.js itineraries.


fpurcell (approved):
Good find, Binh!  Think this will work well.

evansiroky (approved):

PR #255, opened 2021-06-08
fix(rt vehicle): add a 'switch' parameter (boolean) to the panTo util…

### LINKED ISSUE #254: rt vehicles - switch to control map pan / zoom on vehicle selection and tracking
need to update a couple utils with a map pan boolean
add to storybook to show pan functionality

fpurcell:
Simple add of boolean parameter (pan map) to RT Vehicle tracking utilities -- fixes #254

http://localhost:5555/?path=/story/transitvehicleoverlay--simple-rectangles-click-to-select has a knob to control recentering.

### DIFF
--- a/packages/transit-vehicle-overlay/src/index.story.js
+++ b/packages/transit-vehicle-overlay/src/index.story.js
@@ -138,7 +138,7 @@ function rectangles(popup = true) {
   // give action to the popup vehicle tracking button
   VehiclePopup.defaultProps.setTracked = (vehicle, isTracked) => {
     updateTrackedVehicle(vehicle, isTracked);
-    if (!isTracked) recenter(null, 0); // no longer tracking so clear out cached coords
+    if (!isTracked) recenter(null, 0, 0); // no longer tracking so clear out cached coords
   };
 
   // record marker clicks (and more if no popup used)
@@ -149,7 +149,7 @@ function rectangles(popup = true) {
       updateTrackedVehicle(vehicle, isTracked);
       if (!isTracked)
         // clear recenter coords, so map recenters on repeated clicks of same marker
-        recenter(null, 0);
+        recenter(null, 0, 0);
     }
   };
 
@@ -234,7 +234,7 @@ function realtimeExample(fetchVehicles, fetchPattern, markers) {
   // give action to the popup vehicle tracking button
   VehiclePopup.defaultProps.setTracked = (vehicle, isTracked) => {
     updateTrackedVehicle(vehicle, isTracked);
-    if (!isTracked) recenter(null, 0); // clear out cached coords ... recenter on recently untracked vehicle
+    if (!isTracked) recenter(null, 0, 0); // clear out cached coords ... recenter on recently untracked vehicle
   };
 
   const [trackedVehicle, trackedRef] = getTrackedVehicle();
--- a/packages/transit-vehicle-overlay/src/index.story.js
+++ b/packages/transit-vehicle-overlay/src/index.story.js
@@ -107,7 +107,7 @@ function rectangles(popup = true) {
   const vehicleData = line;
   const patternGeometry = utils.makePattern(geom, "111");
 
-  const doTracking = boolean("Recenter map on tracked vehcile", true);
+  const doTracking = boolean("Recenter map when tracking a different vehicle", true);
   const recenter = utils.recenterFlyTo(null, 0, doTracking);
   const initVehicle = utils.findVehicleById(vehicleData, "9562512");
 
--- a/packages/transit-vehicle-overlay/src/index.story.js
+++ b/packages/transit-vehicle-overlay/src/index.story.js
@@ -106,7 +106,9 @@ function rectangles(popup = true) {
   // initial setup
   const vehicleData = line;
   const patternGeometry = utils.makePattern(geom, "111");
-  const recenter = utils.recenterFlyTo(null);
+
+  const doTracking = boolean("Recenter map on tracked vehcile", true);
+  const recenter = utils.recenterFlyTo(null, 0, doTracking);
   const initVehicle = utils.findVehicleById(vehicleData, "9562512");
 
   // storybook knobs to show off color & highlighting options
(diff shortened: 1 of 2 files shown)


evansiroky (approved):
LGTM

packages/transit-vehicle-overlay/src/index.story.js
```diff
@@ -106,7 +106,9 @@ function rectangles(popup = true) {
   // initial setup
   const vehicleData = line;
   const patternGeometry = utils.makePattern(geom, "111");
-  const recenter = utils.recenterFlyTo(null);
+
+  const doTracking = boolean("Recenter map on tracked vehcile", true);
```
binh-dam-ibigroup, line 110: Maybe a better way to label this flag:
```suggestion
  const doTracking = boolean("Recenter map when tracking a different vehicle", true);
```

packages/transit-vehicle-overlay/src/index.story.js
```diff
@@ -133,7 +135,7 @@ function rectangles(popup = true) {
   // give action to the popup vehicle tracking button
   VehiclePopup.defaultProps.setTracked = (vehicle, isTracked) => {
     updateTrackedVehicle(vehicle, isTracked);
-    if (!isTracked) recenter(null, 0, 0); // no longer tracking so clear out cached coords
+    if (!isTracked) recenter(null, 0, doTracking, 0); // no longer tracking so clear out cached coords
```
binh-dam-ibigroup, line 138: Is the `doTracking` argument be just `false` for vehicles that are no longer tracked?

binh-dam-ibigroup, line 138: What is the intent here? I am confused with why arguments are changing. According to https://github.com/opentripplanner/otp-ui/blob/master/packages/transit-vehicle-overlay/src/utils/recenter.js#L29, only 3 arguments are needed, the first being the map and the last two being coordinates. As written, untracking a vehicle seems to do nothing.

packages/transit-vehicle-overlay/src/index.story.js
```diff
@@ -144,7 +146,7 @@ function rectangles(popup = true) {
       updateTrackedVehicle(vehicle, isTracked);
       if (!isTracked)
         // clear recenter coords, so map recenters on repeated clicks of same marker
-        recenter(null, 0, 0);
+        recenter(null, 0, true, 0);
```
binh-dam-ibigroup, line 149: Same as https://github.com/opentripplanner/otp-ui/pull/255/files#r648383383, I am not sure to understand the change in this function call.

packages/transit-vehicle-overlay/src/index.story.js
```diff
@@ -229,7 +231,7 @@ function realtimeExample(fetchVehicles, fetchPattern, markers) {
   // give action to the popup vehicle tracking button
   VehiclePopup.defaultProps.setTracked = (vehicle, isTracked) => {
     updateTrackedVehicle(vehicle, isTracked);
-    if (!isTracked) recenter(null, 0, 0); // clear out cached coords ... recenter on recently untracked vehicle
+    if (!isTracked) recenter(null, 0, true, 0); // clear out cached coords ... recenter on recently untracked vehicle
```
binh-dam-ibigroup, line 234: Same as https://github.com/opentripplanner/otp-ui/pull/255/files#r648383383, I am not sure to understand the change in this function call.

binh-dam-ibigroup (changes requested):
@fpurcell, could you clarify what is happening around the code at this comment https://github.com/opentripplanner/otp-ui/pull/255/files#r648383383, please? The rest seems fine with the new cool flying effect.

packages/transit-vehicle-overlay/src/index.story.js
fpurcell: Hey Binh

A customer can select a vehicle, which then highlights that vehicle, shows the pattern geometry of the vehicle, and optionally pan/zooms the map to that tracked vehicle.  Was asked to add a switch for the customer to control that last bit, allowing an on or off state of the map pan / zoom (of the selected vehicle) ... and when switched off, tracking performs the rest of the tracking functionality minus the map pan/zoom to the vehicle's position.

I originally hoped to switch map.panTo on/off by setting the TransitVehicleOverlay's onRecenterMap (PropType.func) prop to either null or a function that didn't have a map.panTo().  Leaflet throws a bunch of errors if I try to replace the onRecenterMap function on a mounted component ... so no switching back and forth to the panTo() no panTo().   

So my second option was to keep the one function prop, and pass a boolean into the prop, that when it is false, it will not execute the panTo.  

Does this explaination help?  If so, I'll add it to the comments in utils/recenter.js.

fpurcell: BTW, I'm going to rename that doTracking variable to doPanMap to avoid confusion in the story.

binh-dam-ibigroup: > So my second option was to not muck with the onRecenterMap() prop, but rather pass a boolean into onRecenterMap that, when false, will not execute the map.panTo.

Indeed that is how it looks like, and you probably took care of it already when setting up `recenterFlyTo` and `recenterPanTo` around line 111.

For this particular function call, if you step into that function call, however, [this code](https://github.com/opentripplanner/otp-ui/blob/master/packages/transit-vehicle-overlay/src/utils/recenter.js#L41) gets called, and the arguments passed don't match the arguments of that code, that's where it is confusing.

binh-dam-ibigroup: So to clarify, the signature of the `recenter` function here and elsewhere is `recenter(map, lat, lon)`, so as written, this is happening:
- `map = null`
- `lat = 0`
- `lon = true/false?`

packages/transit-vehicle-overlay/src/utils/recenter.js
```diff
@@ -52,13 +68,13 @@ export function recenterPanToOffset(
 }
 
 /** callback used to move the map to some coordinates (e.g., selected vehicle) */
-export function recenterPanTo(pause = 700) {
+export function recenterPanTo(doRecenter = true, pause = 700) {
```
binh-dam-ibigroup, line 71: This might be a breaking change. (It does not seem to be a breaking change for otp-react-redux.)

packages/transit-vehicle-overlay/src/index.story.js
binh-dam-ibigroup: Resolved with https://github.com/opentripplanner/otp-ui/pull/255/commits/ad9b2124727a8718a90d3cb3bec5c451f4348a4a (undoes the changes to invocations of `recenter`).

binh-dam-ibigroup (approved):
Looks better with the last commit. Thanks @fpurcell for the changes!

PR #258, opened 2021-06-24
Add less commonly used params

landonreed:
In otp-rr, we filter out any query param that is not contained within query-params.js. This PR adds all remaining params found in https://github.com/ibi-group/OpenTripPlanner/blob/ibi-dev/src/main/java/org/opentripplanner/api/common/RoutingResource.java


packages/core-utils/src/query-params.js
```diff
@@ -651,10 +651,243 @@ const queryParams = [
...
+  },
+  {
+    name: "waitReluctance",
+    routingTypes: ["ITINERARY"],
+    default: ""
+  },
+  {
+    name: "driveDistanceReluctance",
+    routingTypes: ["ITINERARY"],
+    default: ""
```
evansiroky, line 680: All these default but empty values are making the snapshot massive and might result in sending over an unnecessarily large set of query params that aren't used. Are these default values needed or can they be filtered out when creating a default query?

evansiroky (changes requested):
One blocking comment

packages/core-utils/src/query-params.js
landonreed: Yea, my thinking was that in otp-react-redux, we would just fiilter out any query param that has an empty string or undefined value, i.e., in this filter method:

https://github.com/opentripplanner/otp-react-redux/blob/6058031865d802014f4e7ae49209e6b640c69df7/lib/actions/api.js#L245-L254

evansiroky: There is some [nearly identical code in otp-ui](https://github.com/opentripplanner/otp-ui/blob/d29cd4353b4e99a6a1233124da09129f4d8d694b/packages/core-utils/src/query.js#L366) that also does that. I'm not sure if the otp-react-redux one is much different. As originally mentioned, it seems like the `default` key in each of these could be omitted and then all these params would still be able to be accounted for in the `getRoutingParams` method, but just wouldn't be included if the default value is undefined.

landonreed: Sigh... that's disconcerting. I suppose we should remove the function from otp-react-redux and just rely on this otp-ui version then.

binh-dam-ibigroup: I would agree with @evansiroky to not include these empty strings within the snapshots.

binh-dam-ibigroup (approved):
LGTM

evansiroky (approved):

PR #264, opened 2021-07-14
Make alert icons on itinerary configurable

grant-humphries:
New story: http://localhost:5555/?path=/story/itinerarybody-otp-ui--custom-alert-icons-itinerary
* Add props to the itinerary body component for the two alert icons and
  pass them down through the itinerary tree to the appropriate child
  components
* Reconfigure the existing icon components to be overrideable prop
  defaults
* Add a story to the itinerary storybook to illustrate this new
  functionality

### DIFF
--- a/packages/itinerary-body/src/index.js
+++ b/packages/itinerary-body/src/index.js
@@ -29,7 +29,9 @@ const ItineraryBody = ({
   timeOptions,
   toRouteAbbreviation,
   TransitLegSubheader,
-  TransitLegSummary
+  TransitLegSummary,
+  AlertToggleIcon,
+  AlertBodyIcon
 }) => {
   /*
     TODO: replace component should update logic? companies is simply used to
@@ -77,6 +79,8 @@ const ItineraryBody = ({
           toRouteAbbreviation={toRouteAbbreviation}
           TransitLegSubheader={TransitLegSubheader}
           TransitLegSummary={TransitLegSummary}
+          AlertToggleIcon={AlertToggleIcon}
+          AlertBodyIcon={AlertBodyIcon}
         />
       );
     }
@@ -200,7 +204,17 @@ ItineraryBody.propTypes = {
    * - leg: the transit leg
    * - stopsExpanded: whether the intermediate stop display is currently expanded
    */
-  TransitLegSummary: PropTypes.elementType.isRequired
+  TransitLegSummary: PropTypes.elementType.isRequired,
+  /**
+   * A custom icon component inserted into the transit alert toggle button
+   * within a transit leg, if this prop is not supplied a default icon is used
+   */
+  AlertToggleIcon: PropTypes.elementType,
+  /**
+   * A custom icon component inserted into the transit alert body component
+   * within a transit leg, if this prop is not supplied a default icon is used
+   */
+  AlertBodyIcon: PropTypes.elementType
 };
 
 function noop() {}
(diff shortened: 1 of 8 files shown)


packages/itinerary-body/src/index.js
```diff
@@ -200,7 +204,17 @@ ItineraryBody.propTypes = {
    * - leg: the transit leg
    * - stopsExpanded: whether the intermediate stop display is currently expanded
    */
-  TransitLegSummary: PropTypes.elementType.isRequired
+  TransitLegSummary: PropTypes.elementType.isRequired,
+  /**
+   * A custom SVG icon component inserted into the transit alert toggle button
```
binh-dam-ibigroup, line 209: The component can be any graphics content, I suppose.

binh-dam-ibigroup (approved):
Thank you @grant-humphries for the proposed changes, I am inclined to accept.

packages/itinerary-body/src/index.js
grant-humphries: you're right, it doesn't have to specifically be an SVG, really any component could be passed here.  I will update these docstrings to reflect that

grant-humphries: removed the bit about SVG from the docstrings for the new props, I amended my existing commit and force pushed fyi in case you'll have this branch checked out locally

fpurcell (approved):
This looks good to me.  I built the code, and see the different alert icons in the story.  Don't see any errors or prop warning related to this component / change, etc...

packages/itinerary-body/src/index.js
```diff
@@ -200,7 +204,17 @@ ItineraryBody.propTypes = {
...
+  TransitLegSummary: PropTypes.elementType.isRequired,
+  /**
+   * A custom icon component inserted into the transit alert toggle button
+   * within a transit leg
+   */
+  AlertToggleIcon: PropTypes.elementType,
+  /**
+   * A custom icon component inserted into the transit alert body component
+   * within a transit leg
+   */
```
evansiroky, line 216: It might be nice to note that if nothing is provided, a default icon still shows up.

evansiroky (approved):
I have just one non-blocking comment.

packages/itinerary-body/src/index.js
grant-humphries: amended the docstrings to reflect this ^

landonreed:
@grant-humphries, thanks for the changes. A friendly reminder that we should typically never be force-pushing to remote repos. In the future, when addressing comments during code review, please just add new commits. Doing a force push could have very bad effects for others that are also working on this project when they try to pull the code down. While there seem to be mixed feelings about this on the web, GitHub says:
> If you're working in the shared repository model, we recommend that you use a topic branch for your pull request. While you can send pull requests from any branch or commit, with a topic branch you can push follow-up commits if you need to update your proposed changes. When pushing commits to a pull request, **don't force push**. Force pushing can corrupt your pull request.

I'm going to push a dummy "feat" commit to this branch with the correct commit message syntax ([see the guide here for future work](https://github.com/semantic-release/semantic-release#commit-message-format)) and then will merge.

grant-humphries:
Thanks for the review Landon, Evan, Binh and Frank.  With respect to the force push, I'm glad to follow the practices you use here and will do so going forward.

Do you all have preference on how you get tagged to review a pull request as far as it being `reviewer` or `assignee` and should I plan to tag all 4 of you folks for future PRs?

landonreed:
Feel free to tag everyone as `assignee` (and maybe drop a note in slack for good measure). We use https://github.com/pulls/assigned to see what's next on our plate, so assigning is the best bet.

PR #265, opened 2021-07-15
Hide Vehicle Overlay When Appropriate 

miles-grant-ibigroup:
Closes opentripplanner/otp-react-redux#414

The issue was the `VehicleRentalOverlay` component taking a `visible` prop, but doing nothing with it...

### DIFF
--- a/packages/vehicle-rental-overlay/package.json
+++ b/packages/vehicle-rental-overlay/package.json
@@ -18,7 +18,6 @@
     "url": "git+https://github.com/opentripplanner/otp-ui.git"
   },
   "dependencies": {
-    "@opentripplanner/base-map": "^1.1.0",
     "@opentripplanner/core-utils": "^3.0.4",
     "@opentripplanner/from-to-location-picker": "^1.0.3",
     "@opentripplanner/zoom-based-markers": "^1.0.1",
@@ -27,6 +26,7 @@
     "styled-icons": "^9.1.0"
   },
   "peerDependencies": {
+    "@opentripplanner/base-map": "^1.1.0",
     "react": "^16.8.6",
     "react-dom": "^16.8.6",
     "react-leaflet": "^2.6.1",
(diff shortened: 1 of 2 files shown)


binh-dam-ibigroup (approved):
Nice!

packages/vehicle-rental-overlay/package.json
```diff
@@ -18,6 +18,7 @@
     "url": "git+https://github.com/opentripplanner/otp-ui.git"
   },
   "dependencies": {
+    "@opentripplanner/base-map": "^1.1.0",
```
landonreed, line 21: I don't think this needs to be a dep since it's contained within peer deps. Perhaps removing this line and updating the version manually in the list of peer deps would be the easiest fix.

packages/vehicle-rental-overlay/src/index.js
```diff
@@ -158,7 +158,7 @@ class VehicleRentalOverlay extends MapLayer {
       );
     }
 
-    if (!filteredStations || filteredStations.length === 0) {
+    if (!visible || !filteredStations || filteredStations.length === 0) {
```
landonreed, line 161: This works, but we should probably just return null if not visible on line 153. Otherwise, the empty featureGroup might cause confusion to a reader/debugger and we can skip the station filtering above.

landonreed (changes requested):
Just a couple of minor tweaks, but nice work with the needle in the haystack!

packages/vehicle-rental-overlay/package.json
miles-grant-ibigroup: I somehow entirely missed that peer deps were even there. Have updated and things work nicely now thank you!

miles-grant-ibigroup: I tried this, but got serious react errors (it was loading the wrong version). When I move the dependency from peer deps to deps things work as normal... I'm not sure why

packages/vehicle-rental-overlay/src/index.js
miles-grant-ibigroup: Makes sense thanks!

miles-grant-ibigroup:
Still facing this strange dependency issue but otherwise have made all changes!

landonreed (approved):

PR #267, opened 2021-07-20
Properly render vehicle rental layer with 0 stations

evansiroky:
Also fix the vehicle rental storybook stories and improve the storybook example to show detailed controlling of the vehicle rental layer via handlers of layer control events.

### DIFF
--- a/packages/vehicle-rental-overlay/src/VehicleRentalOverlay.story.js
+++ b/packages/vehicle-rental-overlay/src/VehicleRentalOverlay.story.js
@@ -10,6 +10,7 @@ import bikeRentalStations from "../__mocks__/bike-rental-stations.json";
 import carRentalStations from "../__mocks__/car-rental-stations.json";
 import eScooterStations from "../__mocks__/e-scooter-rental-stations.json";
 import { HubAndFloatingBike } from "./DefaultMarkers";
+import LeafletLayerControlInterface from "./leaflet-layer-control-interface";
 
 import "../../../node_modules/leaflet/dist/leaflet.css";
 
@@ -34,7 +35,7 @@ const MyCircle = ({ fillColor = "gray", pixels, strokeColor }) => {
       {children}
     </CircleMarker>
   );
-  GeneratedCircle.coreUtils.types.propTypes = {
+  GeneratedCircle.propTypes = {
     children: PropTypes.node,
     entity: coreUtils.types.stationType.isRequired
   };
@@ -182,7 +183,8 @@ class ZoomControlledMapWithVehicleRentalOverlay extends Component {
       getStationName,
       mapSymbols,
       refreshVehicles,
-      stations
+      stations,
+      visible
     } = this.props;
     const { zoom } = this.state;
     return (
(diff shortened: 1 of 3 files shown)


miles-grant-ibigroup (approved):

packages/vehicle-rental-overlay/src/leaflet-layer-control-interface.tsx
```diff
@@ -0,0 +1,75 @@
...
+   */
+  visible: boolean;
+};
+
+/**
+ * This class is a necessary intermediary to handle events from the Leaflet map
+ * existing in the parent element. See more notes on the other properties and
+ * methods for how everything interacts.
+ */
+export default class LeafletLayerControlInterface extends Component<
```
binh-dam-ibigroup, line 34: (non-blocking) There seem to be two opposite approaches, one taken in OTP-UI by e.g. `ParkAndRideOverlay`, `StopsOverlay`, and `TransitVehicleOverlay` that doesn't keep the `visible` state of the layer, and this one, also used in OTP-RR by `ConnectedVehicleRentalOverlay` and `GtfsRtVehicleOverlay`... I am still missing something about the two approaches, but long term I would favor an approach similar to `ParkAndRideOverlay`.

binh-dam-ibigroup (approved):

evansiroky:
> (non-blocking) There seem to be two opposite approaches, one taken in OTP-UI by e.g. `ParkAndRideOverlay`, `StopsOverlay`, and `TransitVehicleOverlay` that doesn't keep the `visible` state of the layer, and this one, also used in OTP-RR by `ConnectedVehicleRentalOverlay` and `GtfsRtVehicleOverlay`... I am still missing something about the two approaches, but long term I would favor an approach similar to `ParkAndRideOverlay`.

After looking at the code, I completely agree with what @binh-dam-ibigroup is saying here. The OTP-RR approach of handling visibility via state should not be used at all. Therefore, I'm closing this PR and superseding it with https://github.com/opentripplanner/otp-ui/pull/269.

evansiroky:
> > (non-blocking) There seem to be two opposite approaches, one taken in OTP-UI by e.g. `ParkAndRideOverlay`, `StopsOverlay`, and `TransitVehicleOverlay` that doesn't keep the `visible` state of the layer, and this one, also used in OTP-RR by `ConnectedVehicleRentalOverlay` and `GtfsRtVehicleOverlay`... I am still missing something about the two approaches, but long term I would favor an approach similar to `ParkAndRideOverlay`.
> 
> After looking at the code, I completely agree with what @binh-dam-ibigroup is saying here. The OTP-RR approach of handling visibility via state should not be used at all. Therefore, I'm closing this PR and superseding it with #269.

All right, hopefully for a final time, I'm changing my mind and reopening this again as I now see the purpose of handling visibility via state in the `LeafletLayerControlInterface` component. I have refactored the comments of the `LeafletLayerControlInterface` class in 585c0f690db70b5dfa3c68be04ce2522749f1b82.

packages/vehicle-rental-overlay/src/VehicleRentalOverlay.story.js
```diff
@@ -34,7 +36,7 @@ const MyCircle = ({ fillColor = "gray", pixels, strokeColor }) => {
       {children}
     </CircleMarker>
   );
-  GeneratedCircle.coreUtils.types.propTypes = {
+  GeneratedCircle.propTypes = {
```
evansiroky, line 39: @landonreed, this is thing that was causing the problems that you were asking about in https://github.com/opentripplanner/otp-ui/pull/268#issuecomment-883335088

landonreed (approved):
LGTM. Thanks for your diligence on this one, @evansiroky. I'm pretty happy with this approach. My only question is do we need to make this interface change to the other overlays as well (and if so, how do we want to share this common code)?

evansiroky:
> LGTM. Thanks for your diligence on this one, @evansiroky. I'm pretty happy with this approach. My only question is do we need to make this interface change to the other overlays as well (and if so, how do we want to share this common code)?

We could... but that's probably not the next most largest value add thing we could do.

PR #268, opened 2021-07-20
Add storyshot, update React, update styled-components

evansiroky:
I realized that with https://github.com/opentripplanner/otp-ui/pull/256 I introduced a regression such that the Storybook examples for VehicleRentalOverlay were no longer rendered. Therefore, I made this PR will make sure that never happens again. This PR: 

- adds the `@storybook/addon-storyshots` package which allows automatically making snapshot tests of all of our storybook stories
- Properly uses `jest-styled-components` with storyshot
- Updates to the latest `react` version without bumping to a new major version
- Updates to the latest `styled-components` version (this was required to get `jest-styled-components` working
- Adds a bunch of mocks and changes some other tests so that stories can be snapshotted


landonreed:
@evansiroky, thanks for this -- storyshots seems like a great addition! Do you know what in #256 caused the issue?

miles-grant-ibigroup (approved):
This looks really helpful thank you!

README.md
```diff
@@ -30,16 +30,12 @@ Some packages in otp-ui depend on sibling packages (e.g., `@opentripplanner/core
 
    > "@opentripplanner/package-to-test": **"file:../package-to-test"**
 
-2. In your source files, find and replace:
-
-   > opentripplanner/package-to-test/**lib**/types
-
-   with
+2. Run:
+   `yarn && yarn dev`
```
landonreed, line 34: Can we just add a predev script that is `"predev": "yarn"` so that this becomes simply `yarn dev`?

binh-dam-ibigroup: Slightly inconsistent with the "Getting Started" section (there it says `yarn install` then `yarn dev`. If you update per @landonreed's comment above, then remember to update the "Getting Started" section.

README.md
```diff
@@ -30,16 +30,12 @@ Some packages in otp-ui depend on sibling packages (e.g., `@opentripplanner/core
...
 
-   > opentripplanner/package-to-test/**src**/types
+### Storyshot testing
 
-3. Run:
-   `yarn && yarn dev`
+This repo utilizes the [Storyshot](https://storybook.js.org/docs/react/workflows/snapshot-testing) Storybook addon to perform snapshot tests of every story in this monorepo. Whenever the script `yarn unit` is ran, the Storyshot addon will be included along with all the other tests. It will compare the output of every story to the saved snapshot of that story. This provides a quick way to make sure nothing drastic has changed and that every single story is able to render without an error. However, often times these snapshots will need to be updated and that can be accomplished by running `yarn unit -u`.
```
binh-dam-ibigroup, line 38: One thing that should be mentioned is that the snapshots are for the contents initially shown when the user initially opens a story, and that the snapshots do not cover changes from the user interacting with each story. See for instance the `base-map` stories that show empty `<div>` tags, or in the `trip-details` package where snapshot don't cover the expandable calories description.

binh-dam-ibigroup (approved):
LGTM, there should also be a mention in `README.md` about snapshots not covering elements rendered interactively.

grant-humphries:
Looks good to me! Can confirm the vehicle rental overlay story is fixed by this.  Storyshot could be a useful addition to trimet.org as well so glad to know about it

packages/trip-form/src/components.story.js
```diff
@@ -5,10 +5,10 @@ import * as Icons from "@opentripplanner/icons";
...
 import * as Core from ".";
 
-import commonModes from "./__mocks__/modes";
-import modeOptions from "./__mocks__/mode-options";
-import submodeOptions from "./__mocks__/submode-options";
-import trimet from "./__mocks__/trimet.styled";
+import commonModes from "./test-utils/modes";
+import modeOptions from "./test-utils/mode-options";
+import submodeOptions from "./test-utils/submode-options";
+import trimet from "./test-utils/trimet-styled";
```
landonreed, line 11: Why all the file name changes? Just curious.

landonreed (changes requested):
Looks good, but I do suspect that we should be updating react everywhere along with styled-components.

Also, there are a bunch of warnings in the test logs that would be nice to get rid of (we can handle elsewhere if need be): https://github.com/opentripplanner/otp-ui/pull/268/checks#step:8:55

packages/trip-form/src/components.story.js
evansiroky: I can't remember, maybe it seemed like they would be better organized in this folder since they were solely for helping render stories.

README.md
evansiroky: Fixed in 9befb32064eefdab6e9b8a0a41945a130872da10 and 1d369aea05797ecbe500da148cec00a439ec60e1.

README.md
evansiroky: Improved this documentation in 1d369aea05797ecbe500da148cec00a439ec60e1.

packages/from-to-location-picker/package.json
```diff
@@ -15,7 +15,7 @@
   },
   "peerDependencies": {
     "react": "^16.8.6",
```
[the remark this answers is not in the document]
evansiroky: Whoops. Added those in 6afe849ac17632fedc175094d2f337c18d572a63.

evansiroky:
Ready for final review.

landonreed (approved):

PR #269, opened 2021-07-20
Change back to original logic in v1.1.0 of vehicle-rental-overlay

evansiroky:
This undoes the change in https://github.com/opentripplanner/otp-ui/pull/265 that had to do with the `visibility` prop being used in the render method. The `visibility` prop is sent to the Leaflet Layer control which should handle the display of rental vehicles on it's own. The bug with https://github.com/opentripplanner/otp-react-redux/issues/414 should be fixed in otp-rr itself.


evansiroky:
Sorry for the back-and-forth, but I do think that https://github.com/opentripplanner/otp-ui/pull/267 is actually the way to go after all.

PR #271, opened 2021-07-26
fix(core-utils): move from moment-timezone to date-fns

miles-grant-ibigroup:
Is a bit of a proof of concept. Moment could probably be altogether removed. Moment-timezone is only used in core-utils.

Date-fns-tz is included, but can probably be removed. Some methods are also now superfluous, and should be refactored.

I'm not sure where the complete build is built to, so can't report on bundle size changes. However, lines of code and number of methods called both decrease.


packages/core-utils/package.json
```diff
@@ -13,11 +13,13 @@
     "@opentripplanner/geocoder": "^1.0.2",
     "@turf/along": "^6.0.1",
     "bowser": "^2.7.0",
+    "date-fns": "^2.23.0",
+    "date-fns-tz": "^1.1.4",
```
evansiroky, line 17: According to the docs, `date-fns-tz` uses the [Intl](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl) API. This isn't fully supported by IE/Edge and might need a polyfill depending on what the lowest version of IE/Edge is that we need to support. Also, the GitHub page for this recommends using Node 13+ with this for full compatibility. I'm somewhat inclined to call this a breaking change if implementing this. And if so, we should bump to at least node 13 in the `engines` section of this file.

packages/core-utils/src/time.js
```diff
@@ -34,11 +35,9 @@ export function getLongDateFormat(config) {
...
  */
 export function formatDuration(seconds) {
-  const dur = moment.duration(seconds, "seconds");
-  let text = "";
-  if (dur.hours() > 0) text += `${dur.hours()} hr, `;
-  text += `${dur.minutes()} min`;
-  return text;
+  return formatDistance(0, 1000 * seconds, {
+    includeSeconds: false
+  });
```
evansiroky, line 40: This is a breaking change since `date-fns#formatDistance` outputs "fuzzy" time such as "about 2 hours". I think we should keep the original more exact time output.

packages/core-utils/src/time.js
```diff
@@ -48,12 +47,9 @@ export function formatDuration(seconds) {
...
 export function formatDurationWithSeconds(seconds) {
-  const dur = moment.duration(seconds, "seconds");
-  let text = "";
-  if (dur.hours() > 0) text += `${dur.hours()} hr, `;
-  if (dur.minutes() > 0) text += `${dur.minutes()} min, `;
-  text += `${dur.seconds()} sec`;
-  return text;
+  return formatDistance(0, 1000 * seconds, {
+    includeSeconds: true
+  });
```
evansiroky, line 52: Same comment as https://github.com/opentripplanner/otp-ui/pull/271/files#r681324500.

packages/core-utils/src/time.js
```diff
@@ -75,10 +72,7 @@ export function formatTime(ms, options) {
  * @return {string}                   formatted text representation
  */
 export function formatSecondsAfterMidnight(seconds, timeFormat) {
-  return moment()
-    .startOf("day")
-    .seconds(seconds)
-    .format(timeFormat);
+  return format(getSeconds(startOfDay(seconds)), timeFormat);
```
evansiroky, line 75: The call to `getSeconds` returns the number of seconds which is then passed as an argument as milliseconds past the epoch to `format`. I think this is what we're looking for:

```suggestion
  return format(add(startOfDay(new Date()), { seconds }), timeFormat);
```

packages/core-utils/src/time.js
```diff
@@ -88,34 +82,27 @@ export function formatSecondsAfterMidnight(seconds, timeFormat) {
...
-  // resolved. https://github.com/opentripplanner/otp-ui/issues/152
-  if (!moment.tz || typeof moment.tz.guess !== "function") {
-    // eslint-disable-next-line no-console
-    console.warn(
-      "Error guessing user's timezone (moment.tz or moment.tz.guess not defined). Defaulting to America/New_York."
-    );
-    return "America/New_York";
-  }
-  return moment.tz.guess();
+  return Intl.DateTimeFormat().resolvedOptions().timeZone;
```
evansiroky, line 85: There's a very small chance that browser compatibility might be an issue here.

packages/core-utils/src/time.js
```diff
@@ -88,34 +82,27 @@ export function formatSecondsAfterMidnight(seconds, timeFormat) {
...
 /**
  * Formats current time for use in OTP query
  * The conversion to the user's timezone is needed for testing purposes.
  */
 export function getCurrentTime(timezone = getUserTimezone()) {
-  return moment()
-    .tz(timezone)
-    .format(OTP_API_TIME_FORMAT);
+  return format(
+    utcToZonedTime(new Date(Date.now()), timezone),
```
evansiroky, line 94: Isn't doing `new Date(Date.now())` redundant? Wouldn't `new Date()` suffice?

packages/core-utils/src/time.js
```diff
@@ -88,34 +82,27 @@ export function formatSecondsAfterMidnight(seconds, timeFormat) {
...
 /**
  * Formats current date for use in OTP query
  * The conversion to the user's timezone is needed for testing purposes.
  */
 export function getCurrentDate(timezone = getUserTimezone()) {
-  return moment()
-    .tz(timezone)
-    .format(OTP_API_DATE_FORMAT);
+  return format(
+    utcToZonedTime(new Date(Date.now()), timezone),
```
evansiroky, line 105: Isn't doing `new Date(Date.now())` redundant? Wouldn't `new Date()` suffice?

packages/core-utils/package.json
evansiroky: I looked at the `date-fns-tz` code and it seems like the only part of the Intl API that is used is the [Intl.DateTimeFormat()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat) constructor which according to [caniuse](https://caniuse.com/?search=Intl.DateTimeFormat) currently has 95.89% global support and supports nearly any browser released after the year 2017.

packages/core-utils/src/time.js
```diff
@@ -1,8 +1,9 @@
-import moment from "moment-timezone";
+import { formatDistance, startOfDay, getSeconds, format } from "date-fns";
+import { utcToZonedTime } from "date-fns-tz";
 
 // special constants for making sure the following date format is always sent to
 // OTP regardless of whatever the user has configured as the display format
-export const OTP_API_DATE_FORMAT = "YYYY-MM-DD";
+export const OTP_API_DATE_FORMAT = "yyyy-MM-dd";
```
evansiroky, line 6: This may cause problems elsewhere since this variable is exported.

packages/core-utils/src/time.js
evansiroky: See https://github.com/opentripplanner/otp-ui/pull/271/files#r681344828

yarn.lock
```diff
@@ -7680,11 +7680,21 @@ data-urls@^1.0.0:
...
 date-fns@^1.27.2:
   version "1.30.1"
   resolved "https://registry.yarnpkg.com/date-fns/-/date-fns-1.30.1.tgz#2e71bf0b119153dbb4cc4e88d9ea5acfb50dc05c"
   integrity sha512-hBSVCvSmWC+QypYObzwGOd9wqdDpOt+0wl0KbU+R+uuZBS1jN8VsD1ss3irQDknRj5NvxiTF6oj/nDRnN/UQNw==
 
+date-fns@^2.23.0:
+  version "2.23.0"
+  resolved "https://registry.yarnpkg.com/date-fns/-/date-fns-2.23.0.tgz#4e886c941659af0cf7b30fafdd1eaa37e88788a9"
+  integrity sha512-5ycpauovVyAk0kXNZz6ZoB9AYMZB4DObse7P3BPWmyEjXNORTI8EJ6X0uaSAq4sCHzM1uajzrkr6HnsLQpxGXA==
+
```
evansiroky, line 7697: When doing `yarn why` it seems that `date-fns@^1.27.2` is used by some package depended on from `lint-staged` so I don't think we need to worry about a mismatch in major versions.

evansiroky (changes requested):
See recommended changes. Also, if other reviewers concur, I think this PR should be a breaking change.

packages/core-utils/package.json
miles-grant-ibigroup: Note has been added to the readme and package.json engines are updated in 55e76870

packages/core-utils/src/time.js
miles-grant-ibigroup: fixed in 55e76870 by migrating to [formatDistanceStrict](https://date-fns.org/v2.23.0/docs/formatDistanceStrict)

packages/core-utils/src/time.js
miles-grant-ibigroup: fixed in 55e76870 by migrating to [formatDistanceStrict](https://date-fns.org/v2.23.0/docs/formatDistanceStrict)

packages/core-utils/src/time.js
miles-grant-ibigroup: Good catch thank you! Have fixed in 55e76870

packages/core-utils/src/time.js
miles-grant-ibigroup: Note added to readme in 55e76870

packages/core-utils/src/time.js
miles-grant-ibigroup: Using `new Date()` broke the unit tests, but shortening to `Date.now()` works...

packages/core-utils/src/time.js
miles-grant-ibigroup: Using `new Date()` broke the unit tests, but shortening to `Date.now()` works...

packages/core-utils/src/time.js
miles-grant-ibigroup: Have created a well-explained second variable

yarn.lock
miles-grant-ibigroup: I'm not sure what you expect here -- should we remove the old version?

miles-grant-ibigroup:
Thanks for the feedback! Have made most changes, but have a few questions

yarn.lock
evansiroky: I added this comment for other reviewers to see. In general, it is good to avoid installing multiple compatible packages which can happen frequently with the way yarn installs things. But in this case it seems like a necessity and shouldn't case problems.

packages/core-utils/src/time.js
```diff
@@ -48,12 +50,9 @@ export function formatDuration(seconds) {
...
 export function formatDurationWithSeconds(seconds) {
-  const dur = moment.duration(seconds, "seconds");
-  let text = "";
-  if (dur.hours() > 0) text += `${dur.hours()} hr, `;
-  if (dur.minutes() > 0) text += `${dur.minutes()} min, `;
-  text += `${dur.seconds()} sec`;
-  return text;
+  return formatDistanceStrict(0, 1000 * seconds, {
+    includeSeconds: true
+  });
```
evansiroky, line 55: `includeSeconds` isn't a valid argument for `date-fns#formatDistanceStrict` and there is rounding that occurs that still makes the time fuzzy. Also, these methods don't output things quite right. Consider what happens with the input of `5400` (90 minutes) with the `formatDistanceStrict` and also `formatDuration` methods of `date-fns`:

```js
console.log(
  formatDistanceStrict(
    0, 
    seconds * 1000, 
    {
      unit: 'second'
    }
  )
) // ===> "5400 seconds"
console.log(
  formatDuration(
    {
      seconds
    },
    { format: ['hours', 'minutes', 'seconds'] }
  )
) // ===> "5400 seconds"
```

It does seem like this works though:

```js
const hours = Math.floor(seconds / 3600)
const minutes = Math.floor((seconds - hours * 3600) / 60)
const seconds2 = seconds - hours * 3600 - minutes * 60
console.log(
  formatDuration(
    {
      hours,
      minutes,
      seconds: seconds2
    },
    { 
      format: ['hours', 'minutes', 'seconds']
    }
  )
)
```

And, it is possible to add a custom locale object/function like so to get to the abbreviated names:

```js
const hours = Math.floor(seconds / 3600)
const minutes = Math.floor((seconds - hours * 3600) / 60)
const seconds2 = seconds - hours * 3600 - minutes * 60
const specLookup = {
  xHours: 'hr',
  xMinutes: 'min',
  xSeconds: 'sec'
}
console.log(
  formatDuration(
    {
      hours,
      minutes,
      seconds: seconds2
    },
    { 
      format: ['hours', 'minutes', 'seconds'],
      locale: { 
        code: 'en-US',
        formatDistance: (spec, val) => {
          return `${val} ${specLookup[spec]}`
        }
      }
    }
  )
)
```

However, adding in the locale like this may not even be needed since it seems like the direction in otp-react-redux is towards using a custom message component. See https://github.com/opentripplanner/otp-react-redux/blob/518def8f187b888d40dd7dcd9aced7e1af01f74f/lib/components/narrative/default/default-itinerary.js#L48-L70.

So with all that said, maybe the best option here is to use the final code snippet I posted.

packages/core-utils/src/time.js
```diff
@@ -34,11 +38,9 @@ export function getLongDateFormat(config) {
...
  */
 export function formatDuration(seconds) {
-  const dur = moment.duration(seconds, "seconds");
-  let text = "";
-  if (dur.hours() > 0) text += `${dur.hours()} hr, `;
-  text += `${dur.minutes()} min`;
-  return text;
+  return formatDistanceStrict(0, 1000 * seconds, {
+    includeSeconds: false
+  });
```
evansiroky, line 43: Roughly the same comment as https://github.com/opentripplanner/otp-ui/pull/271/files#r682846101

packages/core-utils/src/time.js
```diff
@@ -63,8 +62,9 @@ export function formatDurationWithSeconds(seconds) {
  * @returns {string} formatted text representation
  */
 export function formatTime(ms, options) {
-  return moment(ms + (options && options.offset ? options.offset : 0)).format(
-    options && options.format ? options.format : OTP_API_TIME_FORMAT
+  return format(
+    ms + (options && options.offset ? options.offset : 0),
+    options && options.format ? options.format : OTP_API_DATE_FORMAT_DATE_FNS
```
evansiroky, line 67: This is the `formatTime` method, so this should still use `OTP_API_TIME_FORMAT`:

```suggestion
    options && options.format ? options.format : OTP_API_TIME_FORMAT
```

evansiroky (changes requested):
See latest comments.

packages/core-utils/src/time.js
miles-grant-ibigroup: Your knowledge of how things are supposed to be is helpful here thank you! Have followed your suggestion and implemented your fix in 3bcbd9b4ee5744d97fb83f6041f757dc0d8007ae

packages/core-utils/src/time.js
miles-grant-ibigroup: Implemented in d7288ba2266a543eed305d19395ad909714e1b92. Will add tests to verify that it works the way you expect

packages/core-utils/src/time.js
miles-grant-ibigroup: Very good catch thank you! Resolved in 13999c397c4642208ed8f91450f11f5c740d7846

miles-grant-ibigroup:
Took me a few tries to understand what was going on but have got this all implemented. Let me know if it matches what you were expecting!

packages/core-utils/src/__tests__/time.js
```diff
@@ -21,4 +26,13 @@ describe("time", () => {
...
     });
   });
+
+  describe("getFormattedTime", () => {
+    const withoutSeconds = formatDuration(9401);
+    const withSeconds = formatDurationWithSeconds(9401);
+
+    expect(withSeconds).not.toEqual(withoutSeconds);
+    expect(withSeconds).toMatchSnapshot();
+    expect(withoutSeconds).toMatchSnapshot();
```
evansiroky, line 36: Every test case must be contained within an `it` function. See recommended changes as follows:

```suggestion
    const durationInSeconds = 9401;

    it(`should correctly format ${durationInSeconds} seconds as a duration with seconds`, () => {
      expect(formatDurationWithSeconds(durationInSeconds)).toMatchSnapshot();
    });

    it(`should correctly format ${durationInSeconds} seconds as a duration without seconds`, () => {
      expect(formatDuration(durationInSeconds)).toMatchSnapshot();
    });
```

evansiroky (changes requested):
All non-test code looks good, but I have a blocking comment about the testing being done.

packages/core-utils/src/__tests__/time.js
miles-grant-ibigroup: Thanks for the suggestion on this one! Fixed in 6211bd1

miles-grant-ibigroup:
Thanks! Tests are corrected.

evansiroky (approved):

packages/core-utils/src/time.js
```diff
@@ -1,10 +1,60 @@
...
+  add,
+  format,
+  formatDuration as dateFnsFormatDuration
+} from "date-fns";
+import { utcToZonedTime } from "date-fns-tz";
 
 // special constants for making sure the following date format is always sent to
 // OTP regardless of whatever the user has configured as the display format
 export const OTP_API_DATE_FORMAT = "YYYY-MM-DD";
+// Date-Fns uses a different string format that moment.js
```
landonreed, line 12:
```suggestion
// Date-Fns uses a different string format than moment.js
```

packages/core-utils/src/time.js
```diff
@@ -1,10 +1,60 @@
...
+    xSeconds: "sec"
+  };
+  const locale = localize
+    ? {
+        code: "en-US",
+        formatDistance: (spec, val) => {
+          return `${val} ${specLookup[spec]}`;
+        }
+      }
+    : undefined;
```
landonreed, line 44: How will this change (if at all) once localization is in place across otp-ui? Might be worth coordinating with @binh-dam-ibigroup on this if you haven't already to future proof this approach.

packages/core-utils/src/time.js
```diff
@@ -63,7 +104,8 @@ export function formatDurationWithSeconds(seconds) {
  * @returns {string} formatted text representation
  */
 export function formatTime(ms, options) {
-  return moment(ms + (options && options.offset ? options.offset : 0)).format(
+  return format(
+    ms + (options && options.offset ? options.offset : 0),
```
landonreed, line 108: Opportunity for optional chaining? I'm not sure if that's configured here though (i.e., needs any babel plugins).

```suggestion
    ms + (options?.offset || 0),
```

packages/core-utils/src/time.js
```diff
@@ -75,10 +117,7 @@ export function formatTime(ms, options) {
  * @return {string}                   formatted text representation
  */
 export function formatSecondsAfterMidnight(seconds, timeFormat) {
-  return moment()
-    .startOf("day")
-    .seconds(seconds)
-    .format(timeFormat);
+  return format(add(startOfDay(new Date()), { seconds }), timeFormat);
```
landonreed, line 120: Non-blocking nit: This is a touch more difficult to read than the moment dot chaining. I sometimes prefer an interim variable here or there to help with legibility.
```suggestion
  const time = add(startOfDay(new Date()), { seconds })
  return format(time, timeFormat);
```

packages/core-utils/README.md
```diff
@@ -1,5 +1,9 @@
 Includes various utility functions to assist with common operations dealing with OpenTripPlanner data.
 
+This package depends on [`Intl.DateTimeFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat), which is not supported by older browsers or node versions. [As can be seen here](https://caniuse.com/?search=Intl.DateTimeFormat), support is fairly significant across major browsers. Should older browsers need to be targeted, a polyfill can be added. Such a polyfill is available [here](https://formatjs.io/docs/polyfills/intl-datetimeformat/).
+
```
landonreed, line 4: TIL about [the Year 2038 problem](https://en.wikipedia.org/wiki/Year_2038_problem) at that formatjs website.

packages/core-utils/src/time.js
```diff
@@ -88,34 +127,24 @@ export function formatSecondsAfterMidnight(seconds, timeFormat) {
  */
 export function getUserTimezone() {
```
landonreed, line 128: It'd be nice to see a unit test for this. And also to know what happens for browsers that don't support this API. Do we get a white screen of death? I'd prefer to avoid that if possible! Also the comment probably needs updating.

landonreed (changes requested):
This looks really good. Just a couple of minor nits and then some questions about how things behave if a browser does not support the Intl API.

packages/core-utils/src/time.js
binh-dam-ibigroup: We have not touched distance stuff in otp-react-redux yet, but we have [implemented this for formatting durations](https://github.com/opentripplanner/otp-react-redux/blob/dev/lib/components/narrative/default/default-itinerary.js#L51). And there will be a tipping point where we will have to merge that with core-utils.

packages/core-utils/src/time.js
miles-grant-ibigroup: Whoops! Fixed in 6db148be6701f51c7a26b67765a8b15409bb63de

packages/core-utils/src/time.js
miles-grant-ibigroup: Works perfectly! Refactored in 6db148be6701f51c7a26b67765a8b15409bb63de

packages/core-utils/src/time.js
miles-grant-ibigroup: Much cleaner thanks! Changed in 6db148be6701f51c7a26b67765a8b15409bb63de

packages/core-utils/README.md
miles-grant-ibigroup: All the nuclear power plants running Windows 2000...

packages/core-utils/src/time.js
miles-grant-ibigroup: Have done some testing with this on browser stack and you get a reference error (since `Intl` isn't in `window`). Have added a UTC+0 default for now.

landonreed (approved):

packages/core-utils/src/time.js
```diff
@@ -105,29 +99,30 @@ export function formatDurationWithSeconds(seconds) {
...
- * this website. Use a bit of hackery to force a specific timezone if in a
- * test environment.
+ * Uses Intl.DateTimeFormat() api to get the user's time zone. In a test
+ * environment, pulls timezone information from an env variable. Default to
+ * GMT+0 if the Intl API is unavailable.
  */
 export function getUserTimezone() {
   if (process.env.NODE_ENV === "test") return process.env.TZ;
-  return Intl.DateTimeFormat().resolvedOptions().timeZone;
+  return Intl?.DateTimeFormat().resolvedOptions().timeZone || "Etc/Greenwich";
```
landonreed, line 129: One final item: instead of defaulting to GMT+0, could we make that an optional arg to this method? E.g.:
```javascript
export function getUserTimezone(fallbackTimeZone = "Etc/Greenwich") {
```

landonreed (approved):
Looks good, just one final nit: https://github.com/opentripplanner/otp-ui/pull/271/files/6211bd13bef75bebf79b62ee64f61a9ccdeedd5f..5980107e9648ea90bb05d4eb472b2814a36aecfb#r694224679

Feel free to hit the merge button once addressed.

packages/core-utils/src/time.js
miles-grant-ibigroup: Good call! Changed in 8d963b4ec3e2a81a4a3e709c5efbafb94d8d098c

PR #274, opened 2021-07-30
Update TripDetails component with messages prop (i18n)

binh-dam-ibigroup:
This PR changes props in the `TripDetails` component so that implementers can pass localized messages to component for display. Localized messages are JSX content that can be built using any localization library.

BREAKING CHANGE: in the `messages` prop, `at` and `fare` are removed, `depart` has a different meaning,
`fare` is changed to `tncFare`. Implementers need to build the entire text and include any applicable dates, numbers, etc. Also, most messages now accept JSX content, so formatting markup such as `<strong>` etc. can be passed.

### DIFF
--- a/packages/trip-details/src/TripDetails.story.js
+++ b/packages/trip-details/src/TripDetails.story.js
@@ -1,3 +1,4 @@
+import coreUtils from "@opentripplanner/core-utils";
 import React from "react";
 import styled from "styled-components";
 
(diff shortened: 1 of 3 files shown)


landonreed (changes requested):
A couple of high level comments:
- this is such a big refactor that I think it warrants adding typescript along with the messages changes. Anytime we're adding more PropTypes, we're just digging a deeper hole on this.
- I would prefer to have the TripDetails component just define a single set of defaults rather than this `||` approach being used throughout the component:
```javascript
{messages.caloriesBurned || (
                  <DefaultCaloriesBurned caloriesBurned={caloriesBurned} />}
```
- Related to the above item, perhaps a better approach would be to allow the user to pass a `getMessagesForItinerary` prop that, if defined, would be merged with the default `messages` prop. The default value for `getMessagesForItinerary` could return an object filled with the various Default components you have defined in `defaults.js`. I'm not 100% on this approach because it still doesn't feel super helpful for our i18n work, but it does seem like a better direction.

It might be good for us to have a quick team chat about the changes we should move forward with.

binh-dam-ibigroup:
Closing in favor of #281.

Issue #237, opened 2021-03-29
localize OTP-UI


fpurcell:
Need to instrument the code, and support other languages.
Slack Conversation follows below...

fpurcell:
Localization was briefly discussed on the MOD / OTP-UI slack channel in Feb 2021:

https://trimet-mod-sandbox.slack.com/archives/CRC48NTJ4/p1614368852004600

fpurcell Feb 26th at 11:47 AM
bug-in-ear: please evaluate react-intl (https://github.com/formatjs/formatjs#formatjs) when looking to localize OTP-UI ... we're using the Gatsby plugin https://github.com/wiziple/gatsby-plugin-intl#gatsby-plugin-intl for localization, which is built on react-intl for the trimet.org project.  So would help us out having one way to localize both OTP-UI and the local code in the trimet.org app.   That said, if react-intl is just wrong, that's good to know too, and we'll adjust on our end. (edited) 
2 replies

Binh Dam  1 month ago
Thanks for the suggestion!
:+1:
1


fpurcell  26 days ago
BTW, just noticed the folks in Norway are using https://www.i18next.com (a gettext derivative) for an OTP 2 javascript UI : http://docs.opentripplanner.org/en/latest/Localization/#updating-translations
I really wanted gettext, in the new trimet.org, having used it in python for the current version -- but the gatsby plugin chosen instead (which also sorta makes sense).  Bottom line, maybe a bit more investigation necessary.

fpurcell:
old - cleanup

PR #262, opened 2021-07-14
Add new TripOptions mode selector

landonreed:
This PR fulfills a request by TriMet to redesign the trimet.org mode selector according to these mock ups: 
https://xd.adobe.com/view/3504215b-9e98-4242-9921-96c7bf196738-6057/

The new TripOptions component is visible in storybook (`yarn dev`) at: http://localhost:5556/?path=/story/settingsselectorpanel--trip-options

PR #281, opened 2021-08-12
Trip Details: Add react-intl and TypeScript

binh-dam-ibigroup:
This PR adds react-intl and TypeScript to the TripDetails package. A new `types` packages is introduced to hold OTP-UI -specific types shared between multiple packages.

The PR contains breaking changes in the TripDetails props. Specifically, implementers must wrap the component with `<IntlProvider>` from react-intl.  The `messages` and `timeOptions` props are dropped, replaced with customizable messages that implementers pass to `<IntlProvider>`. Default English and French message files are provided.  On the other hand, ~a new `currency` props is added~=> EDIT: removed in https://github.com/opentripplanner/otp-ui/pull/281/commits/506a8bd21c74ba893af7e8ce5aa1e9389f1c7113, and slots have been added to customize the expandable details section of departure, fare, and calories section.

package.json and other configurations are tweaked to accommodate these changes.

EDIT: This PR is a partial solution to #237.

image


binh-dam-ibigroup:
There is potential to remove the `currency` prop that is introduced in this PR if #282 is adopted, and use that PR's currency calculation logic instead.

.github/workflows/codespell.yml
```diff
@@ -12,4 +12,5 @@ jobs:
       - uses: codespell-project/actions-codespell@master
         with:
           check_filenames: true
-          skip: ./.git,yarn.lock,*__mocks__/*
+          # skip git, yarn, mocks, and i18n non-English resources.
+          skip: ./.git,yarn.lock,*__mocks__/*,*i18n/fr*
```
evansiroky, line 16: Is there a way to just ignore only the non-english i18n files?

packages/types/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "@opentripplanner/types",
+  "version": "1.0.0",
+  "description": "Typescript types used across multiple OTP-UI packages",
```
evansiroky, line 4:
```suggestion
  "description": "TypeScript types used across multiple OTP-UI packages",
```

packages/trip-details/i18n/en-US.yml
```diff
@@ -0,0 +1,12 @@
+# Default messages for the TripDetails component.
+# To use from a react-intl application, import the content of this file into the messages object
```
evansiroky, line 2: Just a suggestion

```suggestion
# To use from a react-intl application, merge the content of this file into the messages object
```

packages/trip-details/i18n/en-US.yml
```diff
@@ -0,0 +1,12 @@
...
+# that has your other localized strings, and that you pass to the IntlProvider component.
+otpUi.TripDetails.calories: "Calories Burned: <b>{calories, number, ::.}</b>"
```
evansiroky, line 4: I think it'd look a lot nicer to write out the hierarchy of these and then flatten in the code.

packages/trip-details/package.json
```diff
@@ -9,20 +9,21 @@
   "author": "Evan Siroky",
   "license": "MIT",
   "private": false,
+  "files": [
+    "i18n"
+  ],
```
evansiroky, line 14: This might result in only this folder and not the `lib` or `esm` folders being included in the release. What was the intention of putting this in here? Perhaps there is a different way to accomplish that.

packages/trip-details/src/TripDetails.story.tsx
```diff
@@ -0,0 +1,188 @@
+import React from "react";
+import { IntlProvider } from "react-intl";
+import { boolean, select, withKnobs } from "@storybook/addon-knobs";
```
evansiroky, line 3: The [@storybook/addon-knobs](https://www.npmjs.com/package/@storybook/addon-knobs) package is deprecated. Please use [Storybook args and argTypes](https://storybook.js.org/docs/react/writing-stories/args) instead.

packages/trip-details/src/TripDetails.story.tsx
```diff
@@ -0,0 +1,188 @@
...
+    DepartureDetails={DepartureDetails}
+    FareDetails={FareDetails}
+    itinerary={itinerary}
+  />
+);
+
+const intlDecorator = story => {
+  const useMessages = boolean("Use localized messages", true);
+  const locale = select("Locale", ["en-US", "fr"], "en-US");
+  const overrideDefaultMessages = boolean("Override default messages", false);
```
evansiroky, line 95: The [@storybook/addon-knobs](https://www.npmjs.com/package/@storybook/addon-knobs) package is deprecated. Please use [Storybook args and argTypes](https://storybook.js.org/docs/react/writing-stories/args) instead.

packages/trip-details/src/TripDetails.story.tsx
```diff
@@ -0,0 +1,188 @@
...
+      messages={useMessages ? mergedMessages : null}
+    >
+      {story()}
+    </IntlProvider>
+  );
+};
+
+export default {
+  component: TripDetails,
+  decorators: [intlDecorator, withKnobs],
```
evansiroky, line 119: The [@storybook/addon-knobs](https://www.npmjs.com/package/@storybook/addon-knobs) package is deprecated. Please use [Storybook args and argTypes](https://storybook.js.org/docs/react/writing-stories/args) instead.

packages/trip-details/src/index.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+const MESSAGE_ID_PREFIX = "otpUi.TripDetails.";
+const messageIds = {
+  calories: `${MESSAGE_ID_PREFIX}calories`,
+  caloriesDescription: `${MESSAGE_ID_PREFIX}caloriesDescription`,
+  departure: `${MESSAGE_ID_PREFIX}departure`,
+  title: `${MESSAGE_ID_PREFIX}title`,
+  tncFare: `${MESSAGE_ID_PREFIX}tncFare`,
+  transitFare: `${MESSAGE_ID_PREFIX}transitFare`
+};
```
evansiroky, line 22: Since we're assuming the use of react-intl, I think we shouldn't bother using default messages.

packages/trip-details/src/index.tsx
```diff
@@ -0,0 +1,232 @@
...
+import React, { ReactElement } from "react";
+import { FormattedMessage, FormattedNumber } from "react-intl";
+import { CalendarAlt, Heartbeat, MoneyBillAlt } from "styled-icons/fa-solid";
+
+import * as Styled from "./styled";
+import TripDetail from "./trip-detail";
+import { CaloriesDetailsProps, TripDetailsProps } from "./types";
+
+// Load the default messages.
+import defaultMessages from "../i18n/en-US.yml";
```
evansiroky, line 12: Since we're assuming the use of react-intl, I think we shouldn't bother using default messages.

packages/trip-details/src/index.tsx
evansiroky, line 22: This is creating a fair amount of duplication and long-term need to update this object every time a new message is created. Perhaps it'd be better to make a function that assists as follows:

```suggestion
function makeMessageId (suffix: string) {
  return `otpUi.TripDetails.${suffix}`;
}
```

packages/trip-details/src/index.tsx
```diff
@@ -0,0 +1,232 @@
...
+  bikeSeconds,
+  calories,
+  walkSeconds
+}: CaloriesDetailsProps): React.Element {
+  return (
+    <FormattedMessage
+      defaultMessage={defaultMessages[messageIds.caloriesDescription]}
+      id={messageIds.caloriesDescription}
+      values={{
+        // eslint-disable-next-line react/display-name
```
evansiroky, line 46: I think it'd actually be cleaner to adhere to the linting rule and put the following anonymous function into a named function. This should help with debugging.

packages/trip-details/src/index.tsx
```diff
@@ -0,0 +1,232 @@
...
+        a: contents => (
```
evansiroky, line 47: In order to show that more than just an `a` tag is used here, I think it'd be good to rename this to something like `dietaryLink`.

packages/trip-details/src/index.tsx
```diff
@@ -0,0 +1,232 @@
...
+              defaultMessage={defaultMessages[messageIds.transitFare]}
+              id={messageIds.transitFare}
+              values={{
+                b: BoldText,
+                transitFare: (
+                  <FormattedNumber
+                    currency={currency}
+                    value={transitFare / 100}
+                    // eslint-disable-next-line react/style-prop-object
+                    style="currency"
```
evansiroky, line 102: Gosh, this seems like a poor prop choice on the part of react-intl. Is there an alias we can use?

packages/trip-details/src/TripDetails.story.tsx
evansiroky, line 95: Can the `overrideDefaultMessages` be renamed to `useCustomMessages`?

evansiroky (changes requested):
See various comments, but I think we are definitely heading down the right track here finally. I really like how this is shaping up.

packages/trip-details/src/index.tsx
binh-dam-ibigroup: I think default messages are still useful if implementers don't provide their own messages (e.g. while getting started).

packages/trip-details/src/TripDetails.story.tsx
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-ui/pull/281/commits/da0eb14bbfb78531b95e966f84d82529b0c53db5.

packages/trip-details/src/TripDetails.story.tsx
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-ui/pull/281/commits/da0eb14bbfb78531b95e966f84d82529b0c53db5.

packages/trip-details/src/TripDetails.story.tsx
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-ui/pull/281/commits/da0eb14bbfb78531b95e966f84d82529b0c53db5.

packages/trip-details/src/TripDetails.story.tsx
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-ui/pull/281/commits/da0eb14bbfb78531b95e966f84d82529b0c53db5.

.github/workflows/codespell.yml
binh-dam-ibigroup: [I tried here](https://github.com/opentripplanner/otp-react-redux/blob/9770827d47c39ccc9103958790a402acc8f32788/.github/workflows/codespell.yml#L15), but the pattern `./i18n/!(en-US.yml)` isn't recognized for some reason (suggestions welcome).

packages/trip-details/package.json
binh-dam-ibigroup: Yeah that was not necessary. Removed in https://github.com/opentripplanner/otp-ui/pull/281/commits/51d8398e9227f98c910f7b20db5363a707c5f97a.

packages/trip-details/i18n/en-US.yml
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-ui/pull/281/commits/21beebcf571549e35ac87db6072146d47d737417, ended up looking better that way!

packages/trip-details/src/index.tsx
binh-dam-ibigroup: I ended up using a slightly different approach in https://github.com/opentripplanner/otp-ui/pull/281/commits/21beebcf571549e35ac87db6072146d47d737417, but that big block is gone.

packages/trip-details/src/index.tsx
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-ui/pull/281/commits/8940535c0c10fc9e77c5aaac8b44e350cf04eff9.

packages/trip-details/src/index.tsx
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-ui/pull/281/commits/8940535c0c10fc9e77c5aaac8b44e350cf04eff9.

__snapshots__/storybook.test.ts.snap
```diff
@@ -48387,2912 +48387,6 @@ exports[`Storyshots Trip Form Components Submode Selector 1`] = `
 </div>
 `;
 
-exports[`Storyshots TripDetails Bike Only Itinerary 1`] = `
```
evansiroky, line 48390: These should not be deleted.

packages/trip-details/src/index.tsx
```diff
@@ -0,0 +1,257 @@
...
+  Object.keys(sourceObject).forEach((k: string) => {
+    ids[k] = `${idPrefix}.${k}`;
+  });
+  return ids;
+}
+
+const messageIds = getMessageIds<TripDetailsMessages>(
+  defaultMessages,
+  "otpUi.TripDetails"
+);
```
evansiroky, line 41: I'm kind of wondering if all of this is really necessary. It is possible to simply provide string literals for the message IDs and default messages in each `FormattedMessage` call. I suppose this approach might help with static typing to make sure the messages are coded correctly, but there is still no guarantee that the imported data from the messages yml file will have exact matches (especially with the default messages which could prove to be a problem if that fallback were to occur). There does exist some helpful [tooling](https://formatjs.io/docs/tooling/cli) that can automatically detect message definitions and also the default messages which could then be compared against what exists in the yml files. Additionally, removing this might simplify things a bit for downstream implementers so that they aren't required to input yml files in their transpilation workflow.

packages/trip-details/src/TripDetails.story.tsx
```diff
@@ -0,0 +1,251 @@
...
+import TripDetails from ".";
+import * as TripDetailsClasses from "./styled";
+import {
+  CaloriesDetailsProps,
+  DepartureDetailsProps,
+  FareDetailsProps
+} from "./types";
+
+import defaultEnglishMessages from "../i18n/en-US.yml";
+import defaultFrenchMessages from "../i18n/fr.yml";
```
evansiroky, line 17: To reduce confusion with default messages for react-intl, I'd simply call these as follows:

```suggestion
import englishMessages from "../i18n/en-US.yml";
import frenchMessages from "../i18n/fr.yml";
```

packages/trip-details/src/index.tsx
binh-dam-ibigroup: I felt it was necessary to collect message ids because each id is used twice in each `FormattedMessage`, once to set the id of that message, and once to set the default English message.

packages/trip-details/src/TripDetails.story.tsx
```diff
@@ -0,0 +1,251 @@
...
+};
+
+/**
+ * Helper to simplify story declaration.
+ */
+function makeStory(args, Component) {
+  const BoundTripDetails = createTripDetailsTemplate(Component).bind({});
+  BoundTripDetails.args = args;
+  return BoundTripDetails;
+}
```
evansiroky, line 141: This might be what was causing all of those Storyshot snapshots to disappear. I think Storyshot might be unable to process dynamically generated stories like this.

packages/trip-details/src/index.tsx
evansiroky: Yeah, but the point still stands that if the messages aren't actually defined in the yml file, then it won't work properly. Furthermore, now that I think about it, this is probably incompatible with how the formatjs cli extracts message IDs (see [here](https://formatjs.io/docs/tooling/cli)).

packages/trip-details/src/TripDetails.story.tsx
```diff
@@ -0,0 +1,251 @@
...
+    currency: "USD",
+    locale: "en-US",
+    useCustomMessages: false,
+    useLocalizedMessages: true
+  },
+  component: TripDetails,
+  decorators: [intlDecorator],
+  parameters: { controls: { sort: "alpha" } },
+  title: "TripDetails"
+};
```
evansiroky, line 186: Much better, looks great!

evansiroky (changes requested):
We shouldn't delete the storyshot snapshots and we should come to an agreement about how to set the default messages and make message IDs.

packages/trip-details/src/TripDetails.story.tsx
binh-dam-ibigroup: I spent some time replaying the commits of this PR, and a change in https://github.com/opentripplanner/otp-ui/pull/281/commits/6ef1c8493b2073386b830a10020dc237dd158ca6 is causing storyshot to throw out all of these snapshots. Digging further, it appears that importing the YML files is what is causing that, so we have to check that loading YML files is supported in storyshot.

binh-dam-ibigroup: This one was really tricky, I have added a YML loader, tried to play well with other code, and updated the snapshots in https://github.com/opentripplanner/otp-ui/pull/281/commits/9725b4bf3d43303d383b1d42219fe199c3b7601e.

packages/trip-details/src/index.tsx
binh-dam-ibigroup: Partially addressed by https://github.com/opentripplanner/otp-ui/pull/281/commits/9725b4bf3d43303d383b1d42219fe199c3b7601e, bu I suspect we should use the `defineMessages` API?

__snapshots__/storybook.test.ts.snap
binh-dam-ibigroup: Snapshots reinstated and updated in https://github.com/opentripplanner/otp-ui/pull/281/commits/9725b4bf3d43303d383b1d42219fe199c3b7601e.

packages/trip-details/src/index.tsx
binh-dam-ibigroup: `defineMessages` API added in https://github.com/opentripplanner/otp-ui/pull/281/commits/298ffc9e50465de75a7f4c365fc3f12e6db9af16.

binh-dam-ibigroup:
Okay, probably ready for review again with TypeScript fully in force now.

packages/trip-details/src/index.tsx
```diff
@@ -0,0 +1,272 @@
...
+const messages = expandAndDefineMessages("otpUi.TripDetails", {
+  calories:
+    "Text showing the number of calories for the walking and biking legs of a trip.",
+  caloriesDescription:
+    "Text describing how the calories relate to the walking and biking duration of a trip.",
+  departure: "Text showing the departure date/time for a trip.",
+  title: "Title (heading) text of the component.",
+  tncFare: "Text showing the price paid to transportation network companies.",
+  transitFare: "Text showing the price paid for public transit tickets."
+});
```
evansiroky, line 56: I went ahead and tried using the `@formatjs/cli` cli tool and have verified that this approach is not able to properly generate a list of messages because that tool does not appear to do any runtime analysis of what is imported from the `../i18n/en-US.yml` file. It only really seemed to work if message strings were explicitly spelled out as full strings without any prefix or dynamic help. 

Therefore, I'm not convinced this is the exemplar approach that we should be using in every otp-ui file moving forward. The reason I used the `react-intl#defineMessages` method in otp-rr was because I thought if I didn't obtain that variable from the props in a functional component like in [this example](https://formatjs.io/docs/tooling/cli) that it wouldn't work. However, the cli tool does appear to pick up a signature of `intl.formatMessage({ id: "someId" })` anywhere, so I think the `defineMessages` thing is only useful when the same message ID would be used in multiple `intl.formatMessage` calls.

So, after this analysis, my recommendation is as follows:

- go ahead and still import the default messages, and I guess we have to flatten them because of jest (although it'd be ideal if we didn't have to)
- Remove the `expandAndDefineMessages` function and import of `react-intl#defineMessages`.
- In each FormattedMessage component 
  - spell out the entire flattened message ID in the `id` prop of a `FormattedMessage` 
  - spell out the entire flattened message ID in a lookup of the default message from the imported flattened default messages
  - if we are to keep the descriptions, also add a description prop

evansiroky (changes requested):
See latest comment.

packages/trip-details/src/index.tsx
binh-dam-ibigroup: Thanks for clarifying the recommendations, I made the changes in https://github.com/opentripplanner/otp-ui/pull/281/commits/5f530377e60008a64f3408968fab91ec4f02e380.

binh-dam-ibigroup: `@formatjs/cli` is interesting to add, I'll try that in a separate PR.

binh-dam-ibigroup: https://github.com/ibi-group/otp-ui/pull/9 has a basic message id comparison between code and ymls using `@formatjs/cli`.

binh-dam-ibigroup:
> There is potential to remove the `currency` prop that is introduced in this PR if #282 is adopted, and use that PR's currency calculation logic instead.

#282 is now adopted, so commit https://github.com/opentripplanner/otp-ui/pull/281/commits/506a8bd21c74ba893af7e8ce5aa1e9389f1c7113 removes the currency prop.

packages/trip-details/src/index.tsx
```diff
@@ -0,0 +1,247 @@
+import flatten from "flat";
+// @ts-expect-error FIXME: Create TypeScript types for core-utils packages.
```
miles-grant-ibigroup, line 2: I would personally push for the approach used in #262 of putting the message in package.json so it's harder to forget about, but am sway-able

packages/trip-details/src/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+// - the yaml loader for webpack returns a nested object,
+// - the yaml loader for jest returns messages with flattened ids.
+const defaultMessages: Record<string, string> = flatten(defaultEnglishMessages);
+
+/**
+ * Format text bold (used with FormattedMessage).
+ */
+// TODO: Find a better place for this utility.
+function BoldText(contents: ReactElement): ReactElement {
+  return <b>{contents}</b>;
```
miles-grant-ibigroup, line 27: Do we want `<b>` instead of `<strong>`? Looking at the components, it seems like the bold text is semantically bold.

packages/trip-details/src/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+ */
+export function TripDetails({
+  CaloriesDetails = DefaultCaloriesDetails,
+  className = "",
+  DepartureDetails = null,
+  FareDetails = null,
+  itinerary
+}: TripDetailsProps): ReactElement {
+  let companies = "";
+  itinerary.legs.forEach(leg => {
```
miles-grant-ibigroup, line 81: This may be unrelated to your work, but might be cleaner as a `filter()` and `map()`?

packages/trip-details/src/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+              id="otpUi.TripDetails.tncFare"
+              values={{
+                b: BoldText,
+                companies: (
+                  <Styled.TNCFareCompanies>
+                    {companies.toLowerCase()}
+                  </Styled.TNCFareCompanies>
+                ),
+                maxTNCFare: (
+                  <FormattedNumber
```
miles-grant-ibigroup, line 133: We seem to be using a few of the same props on a few of these `FormattedNumber`s. Do you think it's worth putting it into a helper method/wrapper component?

yarn.lock
```diff
@@ -2715,6 +2784,21 @@
...
+  resolved "https://registry.yarnpkg.com/@opentripplanner/core-utils/-/core-utils-3.2.4.tgz#68366f1527fe5a4593a0efe59207c45c0f54c2d2"
+  integrity sha512-3MYdErBl1mpMNj8qkuvyA77OokTRqwzAqcJt5AgIz6aEgPqXfx3znLtoUr/oBw3/OSjMDJB2cCzXEC/iFmGmnQ==
+  dependencies:
+    "@mapbox/polyline" "^1.1.0"
+    "@opentripplanner/geocoder" "^1.0.2"
+    "@turf/along" "^6.0.1"
+    bowser "^2.7.0"
+    lodash.isequal "^4.5.0"
+    moment "^2.24.0"
+    moment-timezone "^0.5.27"
```
miles-grant-ibigroup, line 2798: When merging #285 into this, please make sure `moment-timezone` is out of the lock file!

miles-grant-ibigroup (changes requested):
This looks really great! Especially the Typescript architecture is looking really cool and looking forward to seeing everything else typed as well. I just have a few small questions, and then some stuff about merging #285

packages/trip-details/src/index.tsx
miles-grant-ibigroup: Or even a reducer

packages/trip-details/src/TripDetails.story.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+const intlDecorator = (
+  Story: StoryType,
+  context: {
+    args: StoryArgs;
+  }
+): ReactElement => {
+  const { args } = context;
+  const { locale, useCustomMessages, useLocalizedMessages } = args;
+  const messages = locale === "en-US" ? englishMessages : frenchMessages;
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Could you clarify your comment, please?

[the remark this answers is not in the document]
miles-grant-ibigroup: We're explicitly importing French or English messages based on the locale arg, and I was asking if this is story-specific or not, and if it is _not_ story specific if we could make it more dynamic. But I believe it's story-specific so shouldn't be an issue!

packages/trip-details/src/index.tsx
binh-dam-ibigroup: Good idea, refactored in https://github.com/opentripplanner/otp-ui/pull/281/commits/d39e81032da5df5266a7dfd0e664940c9cdbdd5f.

packages/trip-details/src/index.tsx
binh-dam-ibigroup: Refactored in https://github.com/opentripplanner/otp-ui/pull/281/commits/d39e81032da5df5266a7dfd0e664940c9cdbdd5f.

packages/trip-details/src/index.tsx
binh-dam-ibigroup: Right, changed to render `<strong>` in https://github.com/opentripplanner/otp-ui/pull/281/commits/d39e81032da5df5266a7dfd0e664940c9cdbdd5f.

packages/trip-details/src/TripDetails.story.tsx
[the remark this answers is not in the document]
binh-dam-ibigroup: The locale and messages are story-specific (i.e. the implementer will have to provide the messages), and which messages gets passed to `<IntlProvider>` depends on the args set by the user in the Storybook Controls tab for that story. For simplicity, the contents of the YML files is statically loaded.

packages/trip-details/src/index.tsx
binh-dam-ibigroup: `<b>` tags replaced with `<strong>` in https://github.com/opentripplanner/otp-ui/pull/281/commits/d39e81032da5df5266a7dfd0e664940c9cdbdd5f.

packages/trip-details/src/index.tsx
binh-dam-ibigroup: I added the message to the `tsc` script of the trip-details package, in addition to the TypeScript exception comment which is required or a compilation error occurs. (see https://github.com/opentripplanner/otp-ui/pull/281/commits/d39e81032da5df5266a7dfd0e664940c9cdbdd5f)

yarn.lock
binh-dam-ibigroup: Yep, moment-timezone is out after updating the core-utils dependency (see https://github.com/opentripplanner/otp-ui/pull/281/commits/7735f8b5d76bee70fe23841ee6a9d01a34da5238)

packages/trip-details/package.json
```diff
@@ -4,29 +4,33 @@
...
+  "types": "lib/index.d.js",
   "repository": "https://github.com/opentripplanner/otp-ui.git",
   "homepage": "https://github.com/opentripplanner/otp-ui#readme",
   "author": "Evan Siroky",
   "license": "MIT",
   "private": false,
   "dependencies": {
-    "@opentripplanner/core-utils": "^3.0.4",
-    "@opentripplanner/humanize-distance": "^1.1.0",
+    "@opentripplanner/core-utils": "^3.2.4",
```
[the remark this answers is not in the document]
binh-dam-ibigroup: I brought updates from #285, it was simpler to resolve conflicts from yarn.lock.

binh-dam-ibigroup:
Ready for review again @miles-grant-ibigroup, but I would wait for #285 to be merged before merging this one.

packages/trip-details/src/index.tsx
miles-grant-ibigroup: Really like this approach thanks!

miles-grant-ibigroup (approved):
Looks really great! Thanks so much for all your work on moving this forward.

packages/trip-details/src/types.ts
```diff
@@ -0,0 +1,44 @@
+import moment from "moment";
+// HACK: Probably ok to import the file using a relative path as it is used during dev only.
+// Also prettier does not recognize the import type syntax.
+// eslint-disable-next-line prettier/prettier
+import type { Itinerary } from "../../types/src";
```
fpurcell, line 5: This should be fixed before merging, as the note implies.  Will either @opentripplanner/types/src  -or- #/types/src work?

fpurcell (changes requested):
I took a brief look.  I feel like this will work for the TORA project.  I'm good with merging now, and going from there ... would like two minor issues addressed first tho, Binh.

packages/trip-details/src/index.tsx
```diff
@@ -0,0 +1,240 @@
...
+        {transitFare && (
+          <Styled.TransitFare>
+            <FormattedMessage
+              defaultMessage={defaultMessages["otpUi.TripDetails.transitFare"]}
+              description="Text showing the price of tickets on public transportation."
+              id="otpUi.TripDetails.transitFare"
+              values={{
+                strong: boldText,
+                transitFare: renderFare(currencyCode, transitFare / 100)
+              }}
```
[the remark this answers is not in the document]
binh-dam-ibigroup: @fpurcell Fares returned from OTP are in cents. Previously, TripDetails was using a utility in the OTP-UI package to convert cents to dollars, which we replaced to accommodate the currency returned in the itinerary.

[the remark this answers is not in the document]
binh-dam-ibigroup: Placing formatting tags inside the resource strings makes the most sense, especially in the expandable details for the calories where the order/number of words to be formatted can vary.

packages/trip-details/src/types.ts
binh-dam-ibigroup: Thanks for pointing out this one. `@opentripplanner/types` seems to work also so I used that in https://github.com/opentripplanner/otp-ui/pull/281/commits/1fa98e123757bdfce2c79781f885d8261283c21a.

binh-dam-ibigroup: After changing the import to use the "cleaner" syntax, it looks like the `types` package needs to be published first for the CI to pass, so I opened #291.

binh-dam-ibigroup:
Blocking this PR pending #291 or alternate advice for https://github.com/opentripplanner/otp-ui/pull/281#discussion_r713417109.

packages/trip-details/src/types.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export type CaloriesDetailsProps = {
+  bikeSeconds: number;
+  calories: number;
+  walkSeconds: number;
+};
+
+export type DepartureDetailsProps = {
+  departureDate: moment.Moment;
+};
```
fpurcell, line 14: See notes above about getting rid of moment as a dependency.

binh-dam-ibigroup: Good suggestion, thanks, removed in https://github.com/opentripplanner/otp-ui/pull/281/commits/a9adf69db199e0306ae19215a8bd88625b28634f.

packages/trip-details/src/index.tsx
```diff
@@ -0,0 +1,242 @@
+import flatten from "flat";
+// @ts-expect-error FIXME: Create TypeScript types for core-utils packages.
+import coreUtils from "@opentripplanner/core-utils";
+import moment from "moment";
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Removed in https://github.com/opentripplanner/otp-ui/pull/281/commits/a9adf69db199e0306ae19215a8bd88625b28634f.

grant-humphries (approved):
All looks good here to me and I think the issues that Frank raised have been addressed, are we ready to merge?

binh-dam-ibigroup:
> All looks good here to me and I think the issues that Frank raised have been addressed, are we ready to merge?

@grant-humphries Almost! There is a dependency in #291 for the build, and @fpurcell has the blocking requested change for this PR.

fpurcell:
Apologies @binh-dam-ibigroup, I didn't realize I was a blocker till this morning.  (feel free to ping on slack if I don't respond via github -- that should get my attention)

PR #282, opened 2021-08-12
Make calculateFares also return currency code

binh-dam-ibigroup:
Proposing to expose the currency code of OTP itineraries through method `calculateFares`, so that implementers don't have to re-write that function's code to figure out the currency used. The intended use is to remove the `currency` prop introduced in #281 from the `TripDetails` component.

### DIFF
--- a/packages/core-utils/src/itinerary.js
+++ b/packages/core-utils/src/itinerary.js
@@ -466,10 +466,14 @@ export function getTransitFare(fareComponent) {
   let digits = 2;
   let transitFare = 0;
   let symbol = "$";
+  let currencyCode = "USD";
   if (fareComponent) {
-    digits = fareComponent.currency.defaultFractionDigits;
+    ({
+      currencyCode,
+      defaultFractionDigits: digits,
+      symbol
+    } = fareComponent.currency);
     transitFare = fareComponent.cents;
-    symbol = fareComponent.currency.symbol;
   }
   // For cents to string conversion, use digits from fare component.
   const centsToString = cents => {
@@ -480,6 +484,7 @@ export function getTransitFare(fareComponent) {
   const dollarsToString = dollars => `${symbol}${dollars.toFixed(2)}`;
   return {
     centsToString,
+    currencyCode,
     dollarsToString,
     transitFare
   };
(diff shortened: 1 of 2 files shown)


packages/core-utils/src/itinerary.js
```diff
@@ -480,36 +484,45 @@ export function getTransitFare(fareComponent) {
...
       // TODO: Support non-USD
       minTNCFare += minCost;
       maxTNCFare += maxCost;
+      tncCurrencyCode = currency;
     }
   });
+
   return {
     centsToString,
+    // Hopefully, the currency code is consistent throughout.
```
evansiroky, line 524: I think it might be better to explicitly state in the JSDoc of the `calculateFares` funciton that fares from an itinerary are all assumed to be in the same currency.

evansiroky (approved):
Good enough for now given a stated assumption that only one currency code is present in an itinerary.

packages/core-utils/src/itinerary.js
binh-dam-ibigroup: Yep, thanks, changed in https://github.com/opentripplanner/otp-ui/pull/282/commits/72d732d80a97f84b160024ab08460adcff6fcb1b.

packages/core-utils/src/itinerary.js
```diff
@@ -466,10 +466,14 @@ export function getTransitFare(fareComponent) {
...
   let transitFare = 0;
   let symbol = "$";
+  let currencyCode = "USD";
   if (fareComponent) {
-    digits = fareComponent.currency.defaultFractionDigits;
+    ({
+      currencyCode,
+      defaultFractionDigits: digits,
+      symbol
+    } = fareComponent.currency);
```
landonreed, line 475: Huh, I never knew this was possible! Might be useful to include a comment with link to the explanation for this syntax: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#assignment_without_declaration
```suggestion
    // Assign values without declaration. See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#assignment_without_declaration
    ({
      currencyCode,
      defaultFractionDigits: digits,
      symbol
    } = fareComponent.currency);
```

landonreed (approved):
LGTM. Just one suggestion about including a note to that syntax.

PR #288, opened 2021-09-16
feat(loc-field): Allow custom component for clear button

noletorious:
This PR would allow for a custom component to replace the default Times clear button next to the location-field input.

### DIFF
--- a/packages/location-field/src/index.js
+++ b/packages/location-field/src/index.js
@@ -318,6 +318,7 @@ class LocationField extends Component {
       geocoderConfig,
       inputPlaceholder,
       location,
+      clearButtonIcon,
       LocationIconComponent,
       locationType,
       sessionOptionIcon,
@@ -591,7 +592,7 @@ class LocationField extends Component {
             aria-label="Clear location"
             onClick={this.onClearButtonClick}
           >
-            <Times size={13} />
+            {clearButtonIcon}
           </S.Button>
         </S.InputGroupAddon>
       ) : null;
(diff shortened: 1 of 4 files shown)


packages/location-field/src/index.js
```diff
@@ -593,7 +594,7 @@ class LocationField extends Component {
             aria-label="Clear location"
             onClick={this.onClearButtonClick}
           >
-            <Times size={13} />
+            <ClearButtonIcon />
```
grant-humphries, line 597: this line needs to look like this
```js
{ClearButtonIcon}
```

The idea is that you would instantiate the Icon component before you pass it in a as prop, that gives the user of the `location-field` component the control to pass it whatever props they want to the icon component which can allow custom styling that otp-ui won't have to worry about.  You did this with the default value, so when the default is used you trying to instantiate an already instantiated componet

grant-humphries:
Add a storybook story for that passes a custom icon to this new prop.  This will allow to verify that the new functionality is working as expected without having to plug it into an app. You should be able to get an alternate icons from the `styled-icons` package that is already a dependency for this package

binh-dam-ibigroup (approved):
LGTM!

grant-humphries:
I think this is not related to this merge request because the clear button doesn't work in other stories in this section, but just an fyi in storybook under LocationField --> DesktopContext --> Selected Location & Selected Location Custom Clear the clear button doesn't clear the text.  It does work under the MobileContext section.

Great work @noletorious! @binh-dam-ibigroup will leave it to you and the IBI team to merge

grant-humphries (approved):

miles-grant-ibigroup (approved):
Looks great! Thanks for submitting.

PR #293, opened 2021-09-30
Fix regressions introduced in previous releases

miles-grant-ibigroup:
#283 changed the role of the menu items in the location field dropdown. This prevented the onFocus handler from ensuring their click handler fires before the dropdown disappears, resulting in menu items not being clickable.

This PR also updates the geocoder to work in more build environments, updating to use the latest lonlat.

Also includes a few related patches for location-field to make integration into otp-rr easier.

### DIFF
--- a/packages/location-field/src/index.js
+++ b/packages/location-field/src/index.js
@@ -471,7 +471,7 @@ class LocationField extends Component {
     if (nearbyStops.length > 0 && !suppressNearby) {
       // Add the menu sub-heading (not a selectable item)
       menuItems.push(
-        <S.MenuItem header key="ns-header">
+        <S.MenuItem header centeredText key="ns-header">
           Nearby Stops
         </S.MenuItem>
       );
@@ -516,7 +516,7 @@ class LocationField extends Component {
     if (sessionSearches.length > 0) {
       // Add the menu sub-heading (not a selectable item)
       menuItems.push(
-        <S.MenuItem header key="ss-header">
+        <S.MenuItem header centeredText key="ss-header">
           Recently Searched
         </S.MenuItem>
       );
@@ -552,7 +552,7 @@ class LocationField extends Component {
     if (userLocationsAndRecentPlaces.length > 0 && showUserSettings) {
       // Add the menu sub-heading (not a selectable item)
       menuItems.push(
-        <S.MenuItem header key="mp-header">
+        <S.MenuItem header centeredText key="mp-header">
           My Places
         </S.MenuItem>
       );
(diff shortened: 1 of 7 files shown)


packages/location-field/src/styled.js
```diff
@@ -22,7 +22,8 @@ export const DropdownButton = styled(BaseButton)`
 `;
 
 export const DropdownContainer = styled.span`
-  position: relative;
+  position: contents;
```
binh-dam-ibigroup, line 25: One of these CSS changes causes the dropdown suggestion list to be limited by the width of the text field, and I don't think we want that. On desktop views, the dropdown width should extend to accommodate as much of its contents as possible.
image

packages/location-field/src/index.js
```diff
@@ -430,6 +430,9 @@ class LocationField extends Component {
         .filter(feature => feature.properties.source !== "transit")
         .slice(0, suggestionCount);
 
+      const nonOtherFeaturesPresent =
```
binh-dam-ibigroup, line 433: Rename to `standardFeatures`?

binh-dam-ibigroup (changes requested):
Please advise regarding changes to the width of the dropdown menu in desktop view.

packages/location-field/src/index.js
miles-grant-ibigroup: This is much better thank you! Renamed in 026fb0c

packages/location-field/src/styled.js
miles-grant-ibigroup: I had to make some changes in https://github.com/opentripplanner/otp-react-redux/pull/462 to accommodate this but thank you for suggesting in the end it is much nicer to be able to display the entire title! Updated in be62bb29b9db033a82706971d3c4031a337aef23

packages/location-field/src/index.js
```diff
@@ -430,7 +430,9 @@ class LocationField extends Component {
         .filter(feature => feature.properties.source !== "transit")
         .slice(0, suggestionCount);
 
-      const nonOtherFeaturesPresent =
+      // If no categories of features are returned, this variable is used to
+      // avoid displaying headers
+      const nonStandardFeaturesPresent =
```
binh-dam-ibigroup, line 435: I thought stops and stations fall under "standard" features?

binh-dam-ibigroup: Rename to `standardFeatures`, I think.

binh-dam-ibigroup (approved):
Good to go after renaming variable for standard features.

packages/location-field/src/index.js
miles-grant-ibigroup: I don't think we have any other features, I read it as "other" is the standard feature and the stations and stops were special. I've renamed to `transitFeatures` in 8ea983c6ab004d320de2da88e76bc66ee2be95e3 which I think is more clear.

philip-cline (approved):

PR #273, opened 2021-07-30
Allow customizing labels and options in queryParams

binh-dam-ibigroup:
This PR adds a function intended for external to to inject custom text into the `queryParams` labels, and custom option values and text for the queryParams parameter options.

The PR does not address location names passed to the `formatPlace()` function (they are provided by the geocoder in many OTP-RR instances).

The change is intended to be used by #272 (see #270 for all changes together).

PR #301, opened 2021-10-27
Trip Options Mode Selector Bug Fixes

miles-grant-ibigroup:
This PR comprises different bug fixes in the new TripOptions component.

### DIFF
--- a/packages/core-utils/src/query-params.js
+++ b/packages/core-utils/src/query-params.js
@@ -163,7 +163,7 @@ const queryParams = [
       query.mode && hasTransit(query.mode) && query.mode.indexOf("WALK") !== -1,
     default: 1207, // 3/4 mi.
     selector: "DROPDOWN",
-    label: "Maximum walk",
+    label: "Maximum Walk",
     options: [
       {
         text: "1/10 mile",
@@ -267,14 +267,14 @@ const queryParams = [
     routingTypes: ["ITINERARY"],
     default: "QUICK",
     selector: "DROPDOWN",
-    label: "I prefer",
+    label: "Optimize for",
     options: [
       {
-        text: "The quickest trip",
+        text: "Speed",
         value: "QUICK"
       },
       {
-        text: "The fewest transfers",
+        text: "Fewest Transfers",
         value: "TRANSFERS"
       }
     ]
@@ -287,19 +287,19 @@ const queryParams = [
     routingTypes: ["ITINERARY"],
     default: "SAFE",
     selector: "DROPDOWN",
-    label: "I prefer",
+    label: "Optimize for",
     options: query => {
       const opts = [
         {
-          text: "The quickest trip",
+          text: "Speed",
           value: "QUICK"
         },
         {
-          text: "The bike-friendliest trip",
+          text: "Bike-Friendly Trip",
           value: "SAFE"
         },
         {
-          text: "The flattest trip",
+          text: "Flat Trip",
           value: "FLAT"
         }
       ];
@@ -307,7 +307,7 @@ const queryParams = [
       // Include transit-specific option, if applicable
       if (hasTransit(query.mode)) {
         opts.splice(1, 0, {
-          text: "The fewest transfers",
+          text: "Fewest Transfers",
           value: "TRANSFERS"
         });
       }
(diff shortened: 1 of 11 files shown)


miles-grant-ibigroup:
Thanks @grant-humphries for figuring out how to reproduce this! This is now fixed.

CI tests are failing pending the `core-utils` being separated into a second PR, which will happen once this PR is approved.

packages/core-utils/src/query-params.js
```diff
@@ -267,14 +267,14 @@ const queryParams = [
     routingTypes: ["ITINERARY"],
     default: "QUICK",
     selector: "DROPDOWN",
-    label: "Optimize for",
+    label: "I prefer",
```
grant-humphries, line 270: This language is kind of niche so I feel like the text for these labels should come from the config file.  Also our marketing team likes to change this kind of thing often so would be helpful in that sense too.  Don't see this as blocking to merge this code though

grant-humphries (approved):
I'd recommend changing the default mode in the TripOption story to some kind of transit only variant because that's a common setting for users and it has caused (no resolved) issues that can't be reproduced when the staring value in walk only

miles-grant-ibigroup:
Thanks @grant-humphries! Good call with the default params, have updated this. As to how the strings will be added, I will defer to the rest of my team. Thanks for all your help and guidance!

Assigning to @binh-dam-ibigroup

packages/core-utils/src/query-params.js
binh-dam-ibigroup: Customization of these strings is already covered by https://github.com/opentripplanner/otp-ui/pull/273.

packages/trip-form/src/TripOptions/OptionButton.tsx
```diff
@@ -5,14 +5,15 @@ import * as S from "./styled";
 // eslint-disable-next-line prettier/prettier
 import type { ButtonProps } from "./Checkbox"
 
-export default function OptionButton({ onClick, label, children, className, selected, checked }: {
+export default function OptionButton({ onClick, label, children, className, selected, checked, disabled }: {
```
binh-dam-ibigroup, line 8: Deserves prop sorting.

packages/trip-form/src/test-utils/modes.js
```diff
@@ -128,12 +128,10 @@ const commonModes = {
   accessModes: [
     {
       mode: "WALK",
-      showWheelchairSetting: true,
```
binh-dam-ibigroup, line 131: These deletions of the wheelchair setting will affect demos involving accessibility. Is that intended?

binh-dam-ibigroup (commented):
@grant-humphries, could you advise regarding string customization on your implementation using functionality introduced by #273, and @miles-grant-ibigroup, could you advise regarding removing some of the accessibility demo params.

packages/trip-form/src/test-utils/modes.js
miles-grant-ibigroup: You are right we can probably bring wheelchairs back for some modes!

miles-grant-ibigroup: It's in the car modes which still allows the setting to be tested.

packages/trip-form/src/TripOptions/OptionButton.tsx
miles-grant-ibigroup: Sorted in c58e1ba1b18a029d706b79ede998dd3b486e1d21

miles-grant-ibigroup:
Strings outside of `trip-form` have been reverted as to not cause compatibility issues with others using the project. @grant-humphries as Binh said you should be able to use internationalization to replace these strings.

grant-humphries:
@binh-dam-ibigroup @miles-grant-ibigroup sounds good with respect the label strings, our team will plan to get things wired up via internationalization this week and let you know how it goes

packages/trip-form/src/TripOptions/OptionButton.tsx
```diff
@@ -5,14 +5,15 @@ import * as S from "./styled";
 // eslint-disable-next-line prettier/prettier
 import type { ButtonProps } from "./Checkbox"
 
-export default function OptionButton({ onClick, label, children, className, selected, checked }: {
+export default function OptionButton({ disabled, checked, selected, className, children, label, onClick }: {
```
binh-dam-ibigroup, line 8: Props need to be sorted.

binh-dam-ibigroup (approved):
One place to sort props and it's good to go.

packages/trip-form/src/TripOptions/OptionButton.tsx
miles-grant-ibigroup: Sorted in 
62634db

philip-cline (approved):

PR #304, opened 2021-11-02
fix(packages/location-field): generate better labels for all fields where locations are shown

daniel-heppner-ibigroup:
In the last update, only the such results showed the enhanced location information (street, zip code/neighborhood). With this update, that information is displayed in the main location field as well as recent results history.

### DIFF
--- a/packages/location-field/src/index.js
+++ b/packages/location-field/src/index.js
@@ -552,6 +552,7 @@ class LocationField extends Component {
             <Option
               icon={sessionOptionIcon}
               key={optionKey++}
+              // just use the name if there is no main/secondary field
               title={sessionLocation.main || sessionLocation.name}
               subTitle={sessionLocation.secondary || ""}
               onClick={locationSelected}
--- a/packages/location-field/src/index.js
+++ b/packages/location-field/src/index.js
@@ -98,7 +98,8 @@ class LocationField extends Component {
       /* FIXME only disabled this because it'd take longer to refactor */
       /* eslint-disable-next-line */
       this.setState({
-        value: location.name,
+        //  location could be null if none is set
+        value: location?.name || "",
         geocodedFeatures: []
       });
     }
--- a/packages/location-field/src/index.js
+++ b/packages/location-field/src/index.js
@@ -331,8 +331,17 @@ class LocationField extends Component {
       getGeocoder(geocoderConfig)
         .getLocationFromGeocodedFeature(feature)
         .then(geocodedLocation => {
+          // Create an improved label for the location
+          // Same label as displayed in the search results list
+          const { main, secondary } = generateLabel(
+            geocodedLocation.rawGeocodedFeature.properties
+          );
+          const geocodedLocationImprovedLabel = {
+            ...geocodedLocation,
+            name: `${main}, ${secondary}`
+          };
           // Set the current location
-          this.setLocation(geocodedLocation, "GEOCODE");
+          this.setLocation(geocodedLocationImprovedLabel, "GEOCODE");
           // Add to the location search history. This is intended to
           // populate the sessionSearches array.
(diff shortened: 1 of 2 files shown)


daniel-heppner-ibigroup:
how do I update the test snapshot?

packages/location-field/src/index.js
```diff
@@ -326,11 +341,19 @@ class LocationField extends Component {
...
     const locationSelected = () => {
       getGeocoder(geocoderConfig)
         .getLocationFromGeocodedFeature(feature)
         .then(geocodedLocation => {
+          // add the friendly location labels
+          // FIXME: should this be a copy of the object so it's not mutated?
+          geocodedLocation.main = main;
+          geocodedLocation.secondary = secondary;
           // Set the current location
           this.setLocation(geocodedLocation, "GEOCODE");
```
miles-grant-ibigroup, line 358: we may want to consider overwriting `name`. I don't see us needing the original name ever, and not setting it causes otp-rr to put the wrong string in the URL path. 

To avoid a breaking change we could add a new prop for `nameOverride` and then create a second PR in otp-rr that uses that new prop? The other alternative is to patch OTP-RR to assemble the query parameter differently, which will be a lot more work than this fix.

miles-grant-ibigroup (changes requested):
This looks great! Very clean. The only issue is the otp-rr URL is filled with the `name` rather than our new parameters, which means that when you reload an otp-rr page, the "old" `name` is put in the location field box. We either need to overwrite `name` when we set the location on line 358 (ideally only when a certain flag is set) or refactor otp-rr significantly to make use of the new fields.

packages/location-field/src/index.js
daniel-heppner-ibigroup: I started looking into OTP-RR and I think making changes there would be a biiig project. It involves not just history, but also saved trips, and probably other stuff. I decided to replace the name field and it seems to work fine. Old saved trips will just have the "old" format, which doesn't strike me as a "breaking" change.

binh-dam-ibigroup (approved):
Seems good, maybe try to add an example in storybook for the expected outcome.

miles-grant-ibigroup (approved):
This looks really good and replacing the name seems to solve all the issues we're having! I think the only worry is that this is no longer backwards compatible if people are expecting the original label to be returned, but personally I feel we're safe to change this. Maybe we can mark it as a breaking change? (@binh-dam-ibigroup)

packages/location-field/src/index.js
```diff
@@ -326,11 +327,19 @@ class LocationField extends Component {
...
+
     // Create the selection handler
     const locationSelected = () => {
       getGeocoder(geocoderConfig)
         .getLocationFromGeocodedFeature(feature)
         .then(geocodedLocation => {
+          // add the friendly location labels for use later on
+          geocodedLocation.main = main;
+          geocodedLocation.secondary = secondary;
+          geocodedLocation.name = getCombinedLabel(feature.properties);
```
miles-grant-ibigroup, line 342: This is needed to avoid a crash!

```suggestion
          geocodedLocation.name = getCombinedLabel(feature.properties) || '';
```

miles-grant-ibigroup (changes requested):
Through testing have realized there is a problem!

packages/location-field/src/utils.js
```diff
@@ -50,3 +50,14 @@ export const generateLabel = properties => {
```
miles-grant-ibigroup, line 50: Now that there's a new method in this file, we can get rid of the eslint disable line on line 46!

miles-grant-ibigroup (approved):
Sorry about all the back and forth there, but we are now very ready to go on this! Thanks so much for finding the bugs and getting them fixed this elegantly.

Just one last quick tweak and then you can go ahead and merge!

PR #300, opened 2021-10-27
GTFS FLEX Support - `core-utils` changes only

miles-grant-ibigroup:
This PR contains all `core-utils` changes needed for #297.

PR #309, opened 2021-11-09
fix(core-utils/map): use legGeometry instead of drawing straight lines

miles-grant-ibigroup:
Avoids drawing straight lines through roads and houses when `interStopGeometry` is missing by using `legGeometry`, if present. This does bring the caveat of some transitive features no longer being supported when the `interStopGeometry` is missing.

Closes https://github.com/opentripplanner/otp-react-redux/issues/489

### DIFF
--- a/packages/core-utils/src/map.js
+++ b/packages/core-utils/src/map.js
@@ -184,6 +184,7 @@ export function itineraryToTransitive(itin, companies, getRouteLabel) {
     if (isTransit(leg.mode)) {
       // determine if we have valid inter-stop geometry
       const hasInterStopGeometry = !!leg.interStopGeometry;
+      const hasLegGeometry = !!leg.legGeometry?.points;
       const hasIntermediateStopGeometry =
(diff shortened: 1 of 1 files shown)


daniel-heppner-ibigroup (approved):
I am not totally sure what legGeometry and interStopGeometry are, but I'm guessing:
- legGeometry is geometry for the entire "leg" of a trip (from boarding to alighting)
- interStopGeometry is when we have information between each stop.

miles-grant-ibigroup:
Realizing now this PR was not needed as these changes already existed in #300.

miles-grant-ibigroup:
@daniel-heppner-ibigroup that's exactly it! `interStopGeometry` contains geometry between each stop

leonardehrenfried:
But it's only needed if you want these 45 degree angle maps: https://s3.amazonaws.com/transitive.js/index.html?path=/story/example-transitive--profile-with-45-degree-angles

OTP-RR doesn't ever display these.

PR #311, opened 2021-11-15
ACCESSIBILITY: leaflet marker prop to control tabindex 

fpurcell:
Add an optional Marker prop called 'keyboard' (defaulted to a value of 'false'), which is used by leaflet to either set tabindex="0" (when true) or nothing in the resulting div.  This is important for keyboard-only and SR users, since the current map has potentially 10s to 100s of markers, each that are in the tab order -- the result is that one gets stuck in tab hell on the map, without an exit.  The keyboard=false value will now allow tabbing thru the app, with any variety of marker layer active (stop, vehicle, P&R, bike/scooter share, etc...), and no black hole where the tab order gets stuck w/out an exit.

### DIFF
--- a/packages/park-and-ride-overlay/src/index.js
+++ b/packages/park-and-ride-overlay/src/index.js
@@ -21,7 +21,7 @@ class ParkAndRideOverlay extends MapLayer {
   updateLeafletElement() {}
 
   render() {
-    const { parkAndRideLocations, setLocation } = this.props;
+    const { parkAndRideLocations, setLocation, keyboard } = this.props;
     if (!parkAndRideLocations || parkAndRideLocations.length === 0)
       return <FeatureGroup />;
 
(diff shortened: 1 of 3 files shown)


fpurcell:
BTW, in order to test things on the TORA app, I have hand published -xp (experimental) packages to @opentripplanner on npm.  (Made doubly sure that the changes passed lint and test, and the changes were benign enough to not cause major issues in-case someone picked these packages up).

miles-grant-ibigroup (approved):
Thanks very much for this -- this is long overdue and I appreciate making this a backwards compatible change! 

Have you noticed similar issues the stop markers?

fpurcell:
> Have you noticed similar issues the stop markers?

Hey Miles.  Yes I did notice the same issue when we zoom into the map.  The CircleMarker that is used as the default stop icon is not problematic in capturing tab focus, but the image based markers (used on TORA and in one OTP-UI story) are a problem.  I will see about adding keyboard=false directly to that OTP-UI story (the Markers are created in the story).

For TORA, I we are creating the stop markers on this end, so simply added keyboard=true here.

miles-grant-ibigroup:
That's amazing thanks so much! Assigning @binh-dam-ibigroup to give this a second review and then we should probably be ready to merge

PR #313, opened 2021-11-17
Hopefully Fix Package Releases

miles-grant-ibigroup:
As per https://github.com/semantic-release/github/issues/175#issuecomment-877059469, this PR prevents persisting credentials used during checkout which will hopefully fix the semantic release bot.


binh-dam-ibigroup (approved):
Let's try that.

PR #316, opened 2021-11-23
Accessibility changes: Adding aria tags to improve screen reader performance

c-harper:
Accessibility  updates from a 3rd party review of trimet.org.

### DIFF
--- a/packages/location-field/package.json
+++ b/packages/location-field/package.json
@@ -14,7 +14,13 @@
     "@opentripplanner/geocoder": "^1.1.2",
     "@opentripplanner/humanize-distance": "^1.1.0",
     "@opentripplanner/location-icon": "^1.4.0",
+    "@opentripplanner/core-utils": "^4.1.0",
+    "@opentripplanner/geocoder": "^1.1.0",
+    "@opentripplanner/humanize-distance": "^1.1.0",
+    "@opentripplanner/location-icon": "^1.3.0",
     "@styled-icons/fa-solid": "^10.34.0",
+    "lodash.uniqueid": "^4.0.1",
+    "prop-types": "^15.7.2",
     "throttle-debounce": "^2.1.0"
   },
   "devDependencies": {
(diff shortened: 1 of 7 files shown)


packages/location-field/package.json
```diff
@@ -14,7 +14,13 @@
     "@opentripplanner/geocoder": "^1.1.2",
     "@opentripplanner/humanize-distance": "^1.1.0",
     "@opentripplanner/location-icon": "^1.4.0",
+    "@opentripplanner/core-utils": "^4.1.0",
+    "@opentripplanner/geocoder": "^1.1.0",
+    "@opentripplanner/humanize-distance": "^1.1.0",
+    "@opentripplanner/location-icon": "^1.3.0",
```
miles-grant-ibigroup, line 20: Why were some of these packages downgraded?

packages/location-field/src/index.js
```diff
@@ -74,7 +75,8 @@ class LocationField extends Component {
       menuVisible: false,
       message: null,
       geocodedFeatures: [],
-      activeIndex: null
+      activeIndex: null,
+      listBoxId: `listbox-${uniqueId()}`
```
miles-grant-ibigroup, line 79: There is already an `optionKey` variable which is being used to set a unique key for react. I would suggest either re-using that for the aria-id if possible, or replacing that variable with the new listBoxId.

miles-grant-ibigroup (changes requested):
Thanks so much for adding accessibility to this function! These are long-needed changes.

I only have a few small questions about how the keys are generated as well as what's going on in the dependencies. 

Thanks again for your changes!

binh-dam-ibigroup (approved):
Conditional approval subject to sorting props where indicated. I favor using the proposed method to generate the id for aria properties.

packages/location-field/src/index.js
c-harper: I switched it up to use the optionKey. It works fine as can be seen in storybook and the snapshot tests. Always better to not introduce another dependency.

packages/location-field/package.json
c-harper: I had run a lerna command that I really didn't understand what it was supposed to do and i didn't catch that it added these duplicate refs with lower version numbers. Fixed now.

miles-grant-ibigroup (approved):
Thanks for all your changes! This is looking great

PR #317, opened 2021-11-24
Post-types package fixes

miles-grant-ibigroup:
The introduction of the types package seems to have had some side-effects that prevents `yarn install` from working properly. This PR attempts to address these issues.

### DIFF
--- a/packages/trip-details/src/index.tsx
+++ b/packages/trip-details/src/index.tsx
@@ -1,4 +1,6 @@
 import flatten from "flat";
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore FIXME: Create TypeScript types for core-utils packages.
 import coreUtils from "@opentripplanner/core-utils";
 import { FlexBookingInfo } from "@opentripplanner/types";
 import React, { ReactElement } from "react";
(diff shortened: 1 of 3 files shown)


packages/trip-details/src/index.tsx
```diff
@@ -1,5 +1,4 @@
 import flatten from "flat";
-// @ts-expect-error FIXME: Create TypeScript types for core-utils packages.
```
binh-dam-ibigroup, line 2: You still need this exception, this the change that is breaking the build.

binh-dam-ibigroup (changes requested):
Reinstate the ts expect error exception.

packages/trip-details/src/index.tsx
miles-grant-ibigroup: Moving to `ts-ignore` in 6783ac03e75892e3319d7d755759b836fac5cbda seems to fix things!

miles-grant-ibigroup:
Looks like `expect-error` was breaking things in ways `ignore` doesn't. All works now!

binh-dam-ibigroup (approved):
Good find!

daniel-heppner-ibigroup (approved):

PR #318, opened 2021-11-30
Accessibility trip form buttons

fpurcell:
- Add ariaLabel and ariaChecked to TripOptions/OptionButton(s).
- Adds aria label to the (?) image for shared modes, to describe via screen reader what those buttons opened up.
- git ignore / cleanup

### DIFF
--- a/packages/trip-form/src/TripOptions/styled.ts
+++ b/packages/trip-form/src/TripOptions/styled.ts
@@ -8,7 +8,7 @@ import { isServerEnv } from "./util";
 import type { ButtonProps } from "./Checkbox"
 
 // todo: move this string to localization file (and possibly add more exact info on each particular mode)
-const modeButtonAriaLabel = "Opens a dialog that describes this mode, with optional to third party services.";
+const modeButtonAriaLabel = "Opens a dialog that describes this mode, with optional links to third party services.";
 
 export const TripOptionsContainer = styled.div`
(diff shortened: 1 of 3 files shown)


packages/trip-form/src/TripOptions/styled.ts
```diff
@@ -7,6 +7,9 @@ import { isServerEnv } from "./util";
 // eslint-disable-next-line prettier/prettier
 import type { ButtonProps } from "./Checkbox"
 
+// todo: move this string to localization file (and possibly add more exact info on each particular mode)
+const modeButtonAriaLabel = "Opens a dialog that describes this mode, with optional links to third party services.";
```
miles-grant-ibigroup, line 11: I worry it might be a pain to insert i18n strings from within the styled file. Is there a reason to not pass this string in via prop?

miles-grant-ibigroup (approved):
This looks great! Thanks so much for doing this. 

One quick thing I'd do before assigning to Binh is merge master into this branch, as the gitignore change you made is now also merged in master, so probably best to make sure the merge will be clean.

packages/trip-form/src/TripOptions/styled.ts
fpurcell: good suggestion, Miles .. done.

binh-dam-ibigroup (approved):
LGTM!

PR #322, opened 2021-12-15
Render Flex Legs as arcs in Itinerary View

miles-grant-ibigroup:
This is a small change to not use the OTP-generated shapes. These could often be inaccurate. 

The route viewer is not affected by these changes, as the route viewer does not render flex legs via transitive.


daniel-heppner-ibigroup (approved):
LGTM! Sorry took so long!

philip-cline (approved):
Looks great to me

PR #312, opened 2021-11-16
TripOptions Icon Bugfixes

miles-grant-ibigroup:
This PR will encompass a number of icon-related bugfixes. It also slightly adjusts the styles and adds some props to support style overrides

PR #323, opened 2021-12-15
Trip form i18n and TypeScript

binh-dam-ibigroup:


fpurcell (approved):
Looks good, Binh (lot of work / changes).  Looking forward to getting this merged.

fpurcell (commented):
BTW, I'm a bit lost on all the recent merges ... hoping to see #312 and all the other trip form work merged into an upcoming release.  (BTW, one thing I don't see in this PR's storybook are the company icons in the buttons of the new mode selector).   Let me know if I can help...

packages/trip-form/src/DateTimeSelector/index.tsx
```diff
@@ -0,0 +1,280 @@
...
+  isSelected?: boolean;
+  text: ReactNode;
+  type: DepartArriveValue;
+}
+
+/**
+ * Determines whether the browser supports a particular <input type=<type> /> control,
+ * so we can take advantage of native controls
+ * (especially date/time selection) on modern (mobile) browsers.
+ * @param {*} type One of the HTML5 input types.
```
miles-grant-ibigroup, line 80: Is it worth trying to list these out? Are they in some package somewhere

packages/trip-form/src/SettingsSelectorPanel.story.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+import commonCompanies from "./__mocks__/companies";
+import commonModesEmpty from "./__mocks__/modes-empty";
+import trimet from "./__mocks__/trimet-styled";
+
+import englishMessages from "../i18n/en-US.yml";
+import frenchMessages from "../i18n/fr.yml";
+
+// Customize a few strings from query-params to demonstrate French locale.
+// Normally, a translator would go through the entire query-params file.
```
miles-grant-ibigroup, line 16: Are we decided on this approach? Would it not be possible for the query-params file to contain internationalization keys that then get fed through `formatMessage` here? That feels cleaner than creating many query-params files (in case the query params ever update)

packages/trip-form/src/__mocks__/modes-fr.ts
```diff
@@ -0,0 +1,193 @@
...
+    },
+    {
+      image: "https://trimet.org/global/img/mode-max.png",
+      mode: "TRAM",
+      label: "MAX & Portland Streetcar"
+    },
+    {
+      image: "https://trimet.org/global/img/mode-wes.png",
+      mode: "RAIL",
+      label: "WES Trains régionaux"
```
miles-grant-ibigroup, line 20: Here again would it not maybe make more sense to store internationalization keys?

packages/trip-form/src/types.ts
```diff
@@ -0,0 +1,113 @@
...
+export interface QueryParamChangeEvent {
+  [key: string]: string | number | boolean;
+}
+
+/**
+ * Describes the text/value pair for each query parameter option.
+ */
+export interface QueryParamOptions {
+  text: string;
+  value: SVGForeignObjectElement;
```
miles-grant-ibigroup, line 16: Is a QueryParam value always an SVG?

miles-grant-ibigroup (commented):
More complex than I expected with the default messages, but it makes sense and is a good solution for those who don't want to deal with internationalization.

I just have a few small architecture questions

packages/trip-form/src/types.ts
binh-dam-ibigroup: Good catch... must have been an auto-complete thing.

binh-dam-ibigroup:
@miles-grant-ibigroup hopefully this is ready for review again after what we discussed regarding having query params readily localizable.

miles-grant-ibigroup (approved):
This looks great and implementing it in otp-rr also works great and is easy and painless.  

The default strings are cool, but make implementation difficult, as the ids are no longer rendered by default. This can make it tricky to understand if localization is working correctly. Perhaps we could add some `console.warn` message? But that's quite subjective and definitely not a blocking change. 

Thanks for figuring out how to handle these complex config files!

binh-dam-ibigroup:
> Perhaps we could add some `console.warn` message? 

React-intl already displays warnings like the one below that when a language string is missing:
```
Error: [@formatjs/intl Error MISSING_TRANSLATION] Missing message: "otpUi.queryParameters.maxWalkDistance" for locale "fr", using default message as fallback.
```

philip-cline (approved):

PR #325, opened 2021-12-21
feat(route-viewer-overlay): support clipping route outside of flex zones

miles-grant-ibigroup:
This PR adds a prop to the `route-viewer-overlay` that allows not drawing a route within that route's flex zones. 

Looking for feedback on: 
- How the points are processed. It does a bad job when the points are badly defined, or right next to the zone. A library might do this better.


binh-dam-ibigroup:
On first impression, I would try to show the difference between when the `clipToPatternStops` is applied or not, either using separate stories or using story args that provide interactive control of that parameter. As a nice-to-have, perhaps render the flex stops as part of the story/stories to provide the visual reference of where the clipping is supposed to happen.

packages/route-viewer-overlay/src/index.js
```diff
@@ -55,13 +56,33 @@ class RouteViewerOverlay extends MapLayer {
     Object.values(routeData.patterns).forEach(pattern => {
       if (!pattern.geometry) return;
       const pts = polyline.decode(pattern.geometry.points);
+      let clippedPts = pts;
+      if (clipToPatternStops) {
+        // First, go through all stops to find flex zones
```
binh-dam-ibigroup, line 61: This block could be a function.

miles-grant-ibigroup: Good call! Done in b8c1753

miles-grant-ibigroup:
I originally wanted to add `StopsOverlay` to the story to show how the clipping works, but ran into too many issues with the import (as `StopsOverlay` isn't a dependency of the `RouteViewerOverlay` package). I have, however, added a button to enable and disable the clipping. 

If we'd really like to show the zone we could add a raw `GeoJSON` leaflet layer and render the stops into there... Let me know what you think.

binh-dam-ibigroup:
> If we'd really like to show the zone we could add a raw `GeoJSON` leaflet layer and render the stops into there... Let me know what you think.

That would be ideal!

miles-grant-ibigroup:
GeoJson layer is added in 80d872c!

packages/route-viewer-overlay/src/RouteViewerOverlay.story.js
```diff
@@ -32,3 +36,28 @@ export const WithPathStyling = () => (
...
+
+// Storyshot can't render a story with hooks. Creating a function like this
+// is the workaround. See https://github.com/storybookjs/storybook/issues/8177#issuecomment-599866282
+const FlexRouteWithClipButton = () => {
+  const [clip, setClip] = useState(true);
+  return (
+    <>
+      <button type="button" onClick={() => setClip(!clip)}>
+        {clip ? "Unclip" : "Clip"} Route to Outside Flex Zone
+      </button>
```
binh-dam-ibigroup, line 48: Replace this button with "story args", see https://storybook.js.org/blog/introducing-storybook-args/ and examples in #323 or #281.

packages/route-viewer-overlay/src/RouteViewerOverlay.story.js
```diff
@@ -32,3 +36,28 @@ export const WithPathStyling = () => (
...
+      <BaseMap center={POWDER_SPRINGS} zoom={zoom}>
+        <RouteViewerOverlay
+          clipToPatternStops={clip}
+          routeData={flexRouteData}
+          visible
+        />
+        {/* Since the data is fixed, we know that stops[1] will contain the relevant flex zone. 
+        Using the stopsOverlay is not possible as it is very complex to implement */}
+        <GeoJSON data={flexRouteData.stops[1].geometries.geoJson} />
```
binh-dam-ibigroup, line 57: +1

packages/route-viewer-overlay/src/index.js
```diff
@@ -16,6 +17,28 @@ const isGeomComplete = routeData => {
   );
 };
 
+// helper function that removes all points from array of points that are
+// within flex zones defined in an array of stops
```
binh-dam-ibigroup, line 21: Use JSDoc style `/**` comment.

binh-dam-ibigroup (changes requested):
Use Storybook's Story Args to let users set the clipping prop and it will be good to go.

packages/route-viewer-overlay/src/RouteViewerOverlay.story.js
miles-grant-ibigroup: I've moved to args and added a control in 6945aafe0033c070e7556d47894b2c7de63b4441, but worry that it's a bit less discoverable than the button. Moving to template also has required a few hacks

packages/route-viewer-overlay/src/index.js
miles-grant-ibigroup: Done in 6945aafe0033c070e7556d47894b2c7de63b4441

miles-grant-ibigroup:
Clipping prop can now be set via Storybook Args, but I worry that it's more difficult to use and the code more messy. Let me know what you think.

packages/route-viewer-overlay/src/RouteViewerOverlay.story.js
```diff
@@ -1,34 +1,76 @@
...
   title: "RouteViewerOverlay",
-  component: RouteViewerOverlay
+  component: RouteViewerOverlay,
+  argTypes: {
+    center: hiddenProp,
+    zoom: hiddenProp,
+    routeData: hiddenProp,
+    path: hiddenProp,
+    extraLayer: hiddenProp
+  }
```
binh-dam-ibigroup, line 30: Sort attributes.

packages/route-viewer-overlay/src/RouteViewerOverlay.story.js
```diff
@@ -1,34 +1,76 @@
+/* eslint-disable react/jsx-props-no-spreading */
```
binh-dam-ibigroup, line 1: Neah, the props were previously set explicitly for `<BaseMap>` and `<RouteViewerOverlay>`.

packages/route-viewer-overlay/src/RouteViewerOverlay.story.js
```diff
@@ -1,34 +1,76 @@
...
-      routeData={routeData}
-      visible
-    />
-  </BaseMap>
-);
+export const Default = Template.bind({});
+Default.args = {
+  center: PORTLAND,
+  zoom,
+  routeData
```
binh-dam-ibigroup, line 49: Sort attributes (also in stories below).

binh-dam-ibigroup (approved):
Besides some finishing touches it's good to go.

packages/route-viewer-overlay/src/RouteViewerOverlay.story.js
miles-grant-ibigroup: Sorted in f13469bc7ec8b24af48c993101ec3ca3bb69afcb

packages/route-viewer-overlay/src/RouteViewerOverlay.story.js
miles-grant-ibigroup: Oh good call don't know why I didn't think of this -- cleaned up in f13469bc7ec8b24af48c993101ec3ca3bb69afcb

packages/route-viewer-overlay/src/RouteViewerOverlay.story.js
miles-grant-ibigroup: Sorted in f13469bc7ec8b24af48c993101ec3ca3bb69afcb

miles-grant-ibigroup:
Thanks for that prop spreading catch! Looking much better.

philip-cline (approved):

PR #328, opened 2022-01-12
Add Mapillary Image Link to Walking Itineraries

miles-grant-ibigroup:
To assist in way finding, this PR adds a small button to walking, cycling, and e-scooter itineraries with a link to a Mapillary image. This image is retrieved as the component is loaded, using a key passed into the component via props.

The `itinerary-body` component now also accepts a `mapillaryCallback` prop which allows the implementer to use the matched Mapillary image ID. Screen Shot 2022-01-12 at 1 49 08 PM


packages/itinerary-body/src/AccessLegBody/mapillary-button.tsx
```diff
@@ -0,0 +1,110 @@
...
+ * @param coord     The coordinate to convert to a bounding box
+ * @returns         A bounding box 1 meter around the passed coordinate
+ */
+const generateBoundingBoxFromCoordinate = (coord: [number, number]) => {
+  const WINDOW = 0.000075;
+  const south = coord[0] - WINDOW;
+  const north = coord[0] + WINDOW;
+  const west = coord[1] - WINDOW;
+  const east = coord[1] + WINDOW;
+  return [west, south, east, north];
```
daniel-heppner-ibigroup, line 18: The bounding box uses NESW instead of "NE corner", "SE corner", etc?

packages/itinerary-body/src/AccessLegBody/mapillary-button.tsx
```diff
@@ -0,0 +1,110 @@
...
+      if (json?.data?.length > 0) {
+        setImageId(json.data[0].id);
+      }
+    };
+
+    if (!imageId && !!mapillaryKey) getMapillaryId();
+  }, [coords]);
+
+  const handleClick = () => {
+    if (clickCallback) clickCallback(imageId);
```
daniel-heppner-ibigroup, line 87: Is this clickCallback prop for a future use case? I didn't see anywhere where it actually gets used (other than the props being forwarded from a parent).

daniel-heppner-ibigroup:
Looks great! The vertical alignment of the icon with the time bugs me a little bit, but it seems like that might be an issue for whatever container that is in rather than this icon. 
image

packages/itinerary-body/src/AccessLegBody/mapillary-button.tsx
```diff
@@ -0,0 +1,110 @@
...
+import { StreetView } from "@styled-icons/fa-solid";
+
+/**
+ * Helper method to generate bounding box from a location. Adding the WINDOW to the coordinate
+ * creates a bounding box of approximately 1 meter around the coordinate, which is likely to
+ * encompass any imagery available.
+ * @param coord     The coordinate to convert to a bounding box
+ * @returns         A bounding box 1 meter around the passed coordinate
+ */
+const generateBoundingBoxFromCoordinate = (coord: [number, number]) => {
```
daniel-heppner-ibigroup, line 12: Maybe spread coord into lat and lon?

daniel-heppner-ibigroup (approved):

packages/itinerary-body/src/AccessLegBody/mapillary-button.tsx
miles-grant-ibigroup: [Documentation](https://www.mapillary.com/developer/api-documentation#entities) says "minx,miny,maxx,maxy" which... could be clearer.

packages/itinerary-body/src/AccessLegBody/mapillary-button.tsx
miles-grant-ibigroup: https://github.com/opentripplanner/otp-react-redux/pull/521 for spoilers

packages/itinerary-body/src/AccessLegBody/mapillary-button.tsx
miles-grant-ibigroup: Good call! The array was more for testing anyway. Refactored in db984ce117a40465d9a64def798af631a0dd0679

miles-grant-ibigroup:
I've added some padding so the little person is a bit more aligned with the text! Good catch.

demory (approved):

PR #332, opened 2022-01-21
fix(core-utils): more resiliently expand flex mode

miles-grant-ibigroup:
Part of https://github.com/opentripplanner/otp-react-redux/pull/530

Small adjustment to how we replace `FLEX` with the 3 OTP flex modes. The new approach is more computationally expensive, but more resilient in more situations.

### DIFF
--- a/packages/core-utils/src/__tests__/query-params.js
+++ b/packages/core-utils/src/__tests__/query-params.js
@@ -42,12 +42,16 @@ describe("query-params", () => {
 
 describe("flex-reducer", () => {
   it("should not touch a query that doesn't include flex modes", () => {
-    expect(reduceOtpFlexModes(["WALK", "TRANSIT", "BIKE"])).toMatchSnapshot();
+    expect(reduceOtpFlexModes(["WALK", "TRANSIT", "BIKE"])).toStrictEqual([
+      "WALK",
+      "TRANSIT",
+      "BIKE"
+    ]);
   });
   it("should modify a query that includes some flex modes", () => {
     expect(
       reduceOtpFlexModes(["WALK", "TRANSIT", "BIKE", "FLEX_DIRECT"])
-    ).toMatchSnapshot();
+    ).toStrictEqual(["WALK", "TRANSIT", "BIKE", "FLEX"]);
   });
   it("should modify a query that includes all flex modes", () => {
     expect(
@@ -59,7 +63,7 @@ describe("flex-reducer", () => {
         "FLEX_ACCESS",
         "FLEX_EGRESS"
       ])
-    ).toMatchSnapshot();
+    ).toStrictEqual(["WALK", "TRANSIT", "BIKE", "FLEX"]);
     expect(
       reduceOtpFlexModes([
         "FLEX_DIRECT",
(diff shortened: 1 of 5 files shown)


daniel-heppner-ibigroup (approved):

packages/core-utils/src/query.js
```diff
@@ -466,8 +466,19 @@ export function getRoutingParams(config, currentQuery, ignoreRealtimeUpdates) {
...
   if (params.mode) {
     // Ensure query is in reduced format to avoid replacing twice
-    const mode = reduceOtpFlexModes(params.mode.split(",")).join(",");
-    params.mode = mode.replace("FLEX", "FLEX_EGRESS,FLEX_ACCESS,FLEX_DIRECT");
+    const mode = reduceOtpFlexModes(params.mode.split(","));
+    params.mode = mode
+      .map(m => {
+        // If both the expanded and shrunk modes are included, remove the exapnded one
+        if (m === "FLEX_EGRESS" || m === "FLEX_ACCESS" || m === "FLEX_DIRECT") {
+          if (mode.includes("FLEX")) return "";
```
binh-dam-ibigroup, line 474: Is this going to result in an OTP mode list where there is a comma followed by nothing or another comma?

packages/core-utils/src/query.js
```diff
@@ -466,8 +466,19 @@ export function getRoutingParams(config, currentQuery, ignoreRealtimeUpdates) {
...
+        }
+        if (m === "FLEX") {
+          return "FLEX_EGRESS,FLEX_ACCESS,FLEX_DIRECT";
+        }
+        return m;
+      })
+      .join(",");
```
binh-dam-ibigroup, line 481: You should extract this changed block into its own method and add a test so we can see the expected inputs/outputs.

binh-dam-ibigroup (changes requested):
I think a method should be created to hide that logic, and tests should be added to document the expected behavior.

packages/core-utils/src/query.js
miles-grant-ibigroup: This is checked by the new unit tests and seems to work correctly

packages/core-utils/src/query.js
miles-grant-ibigroup: Good call! Lots of unit tests added in 70d83757ca12f50aea4417d847732854c6c660b8

packages/core-utils/src/__tests__/query-params.js
```diff
@@ -38,3 +39,51 @@ describe("query-params", () => {
...
 });
+
+describe("flex-reducer", () => {
+  it("should not touch a query that doesn't include flex modes", () => {
+    expect(reduceOtpFlexModes(["WALK", "TRANSIT", "BIKE"])).toMatchSnapshot();
+  });
+  it("should modify a query that includes some flex modes", () => {
+    expect(
+      reduceOtpFlexModes(["WALK", "TRANSIT", "BIKE", "FLEX_DIRECT"])
+    ).toMatchSnapshot();
```
binh-dam-ibigroup, line 50: You probably don't need snapshots for simple object comparison. Use `expect(...).toBe(<expected_value>)`.

binh-dam-ibigroup (changes requested):
Thanks for adding the tests, see https://github.com/opentripplanner/otp-ui/pull/332/files/095d0131000373cfc065d33cfcefecd5260cb820..d942adf911c22009318496f5c71facfcd836ca84#r792690034 to simplify.

packages/core-utils/src/__tests__/query-params.js
miles-grant-ibigroup: Snapshots moved in 7be6cdf4055042904e0d895d22825d5df980920f

miles-grant-ibigroup:
Thanks Binh, unit tests refactored

packages/core-utils/src/__tests__/query.js
```diff
@@ -102,6 +102,58 @@ describe("query", () => {
...
       ).toMatchSnapshot();
     });
+
+    it("should create routing params for standard flex query", () => {
+      expect(
+        getRoutingParams(fakeConfig, {
+          ...makeBaseTestQuery(),
+          mode: "WALK,TRANSIT,FLEX"
+        })
+      ).toMatchSnapshot();
```
binh-dam-ibigroup, line 112: Those are probable fine as snapshots.

binh-dam-ibigroup (approved):
Looks good!

PR #338, opened 2022-02-03
fix(geocoder): HERE API bugs

daniel-heppner-ibigroup:
Fixes some issues with the HERE API.
- Certain API responses could cause a crash.
- Boundary is now calculated internally when a focusPoint is also provided.


packages/geocoder/src/apis/here/index.ts
```diff
@@ -66,6 +72,22 @@ function autocomplete({
...
+    const res = await run({
+      options,
+      query,
+      url: AUTOCOMPLETE_URL
+    });
+    if (boundary?.rect) {
+      // HERE does not support a boundary when you use a focus point
+      // This workaround filters the results internally to the boundary
+      res.items = res.items.filter(checkItemInBoundary(boundary))
+      return res
```
miles-grant-ibigroup, line 87: We can probably remove this line right?

packages/geocoder/src/geocoders/here.ts
```diff
@@ -73,8 +73,9 @@ export default class HereGeocoder extends Geocoder {
     const { items } = response;
     return {
       features: items
-      // Here can return continued query suggestions, which we do not support.
-      ?.filter(item => item.resultType !== "chainQuery")
+      // Here has various types of responses, some of which are not locations.
+      // We only want the actual places, so throw out any without a position.
+      ?.filter(item => item.position !== undefined)
```
miles-grant-ibigroup, line 78:
```suggestion
      ?.filter(item => !!item.position)
```
Does this also work?

miles-grant-ibigroup (changes requested):
Looks really good! Just a few small clean up things

packages/geocoder/src/geocoders/here.ts
daniel-heppner-ibigroup: Sure, I was being unnecessarily verbose.

miles-grant-ibigroup (approved):
Looking good!

binh-dam-ibigroup (approved):

PR #340, opened 2022-02-03
Fix rounding so that default selector values work

### LINKED ISSUE #339: Default max walk/bike distances are incorrect
In storybook or on implemented UIs, the default max. walk distance is incorrectly showing as 1/10 mile. Per query-params.js, it should be 3/4 mile. Similar remarks for the default max. bike and escooter distances that should be 3 miles.

binh-dam-ibigroup:
This PR fixes #339 by using a more precise `METERS_PER_MILE` value and some rounding logic.


miles-grant-ibigroup (approved):
This is a good, clean solution. Thanks for figuring this one out!

philip-cline (approved):

PR #342, opened 2022-02-04
Add CO2 Calculation to core-utils

daniel-heppner-ibigroup:
This PR adds a CO2 calculation to Core Utils, allowing OTP to show the CO2 emissions by a given trip option.


daniel-heppner-ibigroup:
I must have branched off the wrong commit, since there's a feat(geocoder) in here. However, nothing changes in any geocoder files since those commits are already merged in. Will this be okay or will it cause an unneeded version bump to geocoder?

PR #345, opened 2022-02-09
Add co2 calc trip details

daniel-heppner-ibigroup:
This adds CO2 value to trip details. Core-utils update appears to have already made it into master.


packages/trip-details/src/types.ts
```diff
@@ -66,4 +72,8 @@ export interface TripDetailsProps {
    * Itinerary that the user has selected to view, contains multiple legs.
    */
   itinerary: Itinerary;
+  /**
+   * Object containing the CO2 config. CO2 only shown if this value is provided.
```
miles-grant-ibigroup, line 76: This comment looks like it might be outdated

packages/trip-details/src/types.ts
```diff
@@ -9,6 +9,12 @@ export interface CaloriesDetailsProps {
   walkSeconds: number;
 }
 
+export interface CO2ConfigType {
+  carbonIntensity: { [index: string]: number };
+  units: string;
+  enabled: boolean;
```
miles-grant-ibigroup, line 15: I would make this optional since it'll still be false if missing

packages/core-utils/src/itinerary.js
```diff
@@ -488,6 +488,57 @@ export function calculatePhysicalActivity(itinerary) {
   };
 }
 
+const CARBON_INTENSITY_DEFAULTS = {
```
miles-grant-ibigroup, line 491: This PR is blocked until the other one is merged right? so this change goes away

packages/trip-details/src/index.tsx
```diff
@@ -330,6 +338,32 @@ export function TripDetails({
             }
           />
         )}
+        {co2 > 0 && co2Config.enabled && (
+          <TripDetail
```
miles-grant-ibigroup, line 342: Do we want to add an info button that links to the resource we used to generate the values? Or maybe just explains the formula or something?

miles-grant-ibigroup (approved):
This looks really cool thank you! 

Just a few small touchups and I think we should be good to go!

packages/trip-details/src/index.tsx
daniel-heppner-ibigroup: Added!

packages/trip-details/src/index.tsx
```diff
@@ -330,6 +338,41 @@ export function TripDetails({
             }
           />
         )}
+        {co2 > 0 && co2Config.enabled && (
```
miles-grant-ibigroup, line 341: You might want to throw a ? after co2config. Getting some white screen of death errors

packages/trip-details/src/index.tsx
```diff
@@ -330,6 +338,41 @@ export function TripDetails({
...
+                />
+              </S.CO2Summary>
+            }
+            description={
+              <FormattedMessage
+                defaultMessage={
+                  defaultMessages["otpUi.TripDetails.co2description"]
+                }
+                description="Text explaining how the CO2 emissions is calculated."
+                id="otpUi.TripDetails.co2description"
```
miles-grant-ibigroup, line 371: Could we change this to include some of the values we use? For example "CO2 output is calculated by multiplying the distance of the trip (2.4km) by the average CO2 emitted by each mode involved."

miles-grant-ibigroup (changes requested):
Having some issues testing this from within otp-rr (because of the separate core-utils) so I would say let's leave this blocked until that is through.

packages/trip-details/src/index.tsx
daniel-heppner-ibigroup: I honestly don't think that would be all that useful information, and it isn't that simple to implement. I don't have immediate access to the trip distance (but I could easily calculate it with a reducer on legs). More importantly, unless we always displayed km, this would also involve unit conversion. And I don't see it as being very useful, since at no point in the calculation are we using the total trip distance (since it's calculated per-leg).

miles-grant-ibigroup: That makes sense! Definitely a nice-to-have so if it's a ton of effort to implement probably not worth it.

Do you think linking to our average emission info is a good idea?

daniel-heppner-ibigroup: What do you mean by linking to our average emissions?

miles-grant-ibigroup: We should probably link to the pdf we used to get the values we use to calculate the emissions the same way we do for the calories burned!

demory:
With this code checked out (including the corresponding branches in otp-rr and configurations), I get the following error after planning a trip and dragging a start/end marker on the map:

image

demory:
A couple accessibility related notes:
- I cannot get a screenreader (using Mac VoiceOver w/ Chrome) to read the "Trip Details" box, including the CO2 info, beyond the header. It reads "Trip Details" but then jumps back to the itinerary summary. Can anyone reproduce?
- The blue "?" buttons are just read as "what does this mean" for both the fare and CO2 items. These should be more specific to each one, maybe "Show fare details" / "Show CO2 details" or similar?

daniel-heppner-ibigroup:
@binh-dam-ibigroup could you help out with some translations?

packages/trip-details/src/index.tsx
daniel-heppner-ibigroup: And added a link to the spreadsheet containing the formula.

packages/trip-details/src/index.tsx
daniel-heppner-ibigroup: Link added.

packages/trip-details/src/index.tsx
```diff
@@ -41,6 +42,19 @@ function dietaryLink(contents: ReactElement): ReactElement {
     </a>
   );
 }
+
+function CO2DescriptionLink(contents: ReactElement): ReactElement {
+  return (
+    <a
+      href="https://www.itf-oecd.org/sites/default/files/life-cycle-assessment-calculations-2020.xlsx"
+      rel="noopener noreferrer"
```
miles-grant-ibigroup, line 50: Do we not want the itf to know we're using their report?

miles-grant-ibigroup (approved):
I know this has been a long work in progress but it looks and works great! Can't wait to see the full otp-rr PR!

packages/trip-details/src/index.tsx
daniel-heppner-ibigroup: ESlint wants those on there for security reasons if your target is a new window... I don't ask questions.

daniel-heppner-ibigroup: ESlint wants those on there for security reasons if your target is a new window... I don't ask questions.

packages/trip-details/i18n/fr.yml
```diff
@@ -7,6 +7,10 @@ otpUi:
       {walkMinutes, plural, one {<strong># minute</strong> effectuée} other {<strong># minutes</strong> effectuées}} à pied et
       {bikeMinutes, plural, one {<strong># minute</strong> effectuée} other {<strong># minutes</strong> effectuées}} en vélo sur ce trajet.
       (d'après <dietaryLink>Dietary Guidelines for Americans 2005, page 16, Table 4</dietaryLink>)
+    co2: "CO<sub>2</sub> émis: <strong>{co2}</strong>"
+    co2description:
```
demory, line 11: Should this be translated?

binh-dam-ibigroup: Done in https://github.com/opentripplanner/otp-ui/pull/345/commits/1afba88c57b910c63040c7eed23e5b5ff64b1a6a.

demory (approved):

PR #346, opened 2022-02-09
Add co2 calc to core-utils

daniel-heppner-ibigroup:
This PR adds a CO2 calculator to core utils


daniel-heppner-ibigroup:
Will the feat(geocoder) be a problem? Not sure how that snuck in here, after merging it doesn't cause any file changes.

packages/core-utils/src/itinerary.js
```diff
@@ -488,6 +488,30 @@ export function calculatePhysicalActivity(itinerary) {
...
 }
 
+/**
+ * @param  {carbonIntensity} carbonIntensity carbon intensity by mode in grams/meter
+ * @param {units} units units to be used in return value
+ * @param  {itinerary} itinerary OTP trip itinierary
+ */
+export function calculateEmissions(carbonIntensity, units, itinerary) {
+  // Distance is in meters, totalCarbon is in grams
+  const totalCarbon = itinerary.legs.reduce((val, leg) => {
```
miles-grant-ibigroup, line 498: I would rename `val` to `total` so it's a bit clearer what's going on!

packages/core-utils/src/itinerary.js
```diff
@@ -488,6 +488,30 @@ export function calculatePhysicalActivity(itinerary) {
   };
 }
 
+/**
+ * @param  {carbonIntensity} carbonIntensity carbon intensity by mode in grams/meter
+ * @param {units} units units to be used in return value
+ * @param  {itinerary} itinerary OTP trip itinierary
+ */
+export function calculateEmissions(carbonIntensity, units, itinerary) {
```
miles-grant-ibigroup, line 496: I haven't seen the trip-details part of this yet, but do you think it's possible/makes sense to put carbon density as a constant inside of core-utils?

miles-grant-ibigroup (approved):
Excited to see the rest of this! Just some tiny nitpicks

miles-grant-ibigroup:
The geocoder commit should be fine... will re-release the geocoder package but won't break things

packages/core-utils/src/itinerary.js
daniel-heppner-ibigroup: Good idea, I'll provide defaults in here.

miles-grant-ibigroup (approved):
So annoying with the extra core-utils PR. I feel like I'm seeing a spoiler for what's coming next.

demory:
Code looks fine to me but I have some questions about the defaults. Are we saying walking is more carbon intensive than some types of transit (cable car / gondola) and micromobility? Maybe I'm I reading that wrong?

daniel-heppner-ibigroup:
@demory Are you maying missing the zero? I think walking is less carbon intensive than most transit modes except maybe gondola.

packages/core-utils/src/itinerary.ts
```diff
@@ -445,6 +445,69 @@ export function calculatePhysicalActivity(
...
 
+/**
+ * Sources:
+ * - https://www.itf-oecd.org/sites/default/files/docs/environmental-performance-new-mobility.pdf
+ * - https://www.thrustcarbon.com/insights/how-to-calculate-emissions-from-a-ferry-journey
+ * Other values extrapolated.
+ */
+const CARBON_INTENSITY_DEFAULTS = {
+  walk: 0.026,
+  bicycle: 0.014,
```
demory, line 456: Private bike: 0.017 (cell AG98)

packages/core-utils/src/itinerary.ts
```diff
@@ -445,6 +445,69 @@ export function calculatePhysicalActivity(
...
+  car: 0.16,
+  tram: 0.062,
+  subway: 0.062,
```
demory, line 459: Subway/metro: .066 (cell ED98) -- use for other rail-based transit modes too

packages/core-utils/src/itinerary.ts
```diff
@@ -445,6 +445,69 @@ export function calculatePhysicalActivity(
...
+  rail: 0.062,
+  bus: 0.09,
+  ferry: 0.082,
+  cable_car: 0.021,
+  gondola: 0.021,
+  funicular: 0.062,
+  transit: 0.062,
+  leg_switch: 0,
+  airplane: 0.382,
+  micromobility: 0.014
```
demory, line 469: Shared e-bike: .083 (cell AJ98)
Shared e-scooter: .106 (cell AE98)
Use average of two? (.095)

packages/core-utils/src/itinerary.ts
```diff
@@ -445,6 +445,69 @@ export function calculatePhysicalActivity(
...
+/**
+ * Sources:
+ * - https://www.itf-oecd.org/sites/default/files/docs/environmental-performance-new-mobility.pdf
+ * - https://www.thrustcarbon.com/insights/how-to-calculate-emissions-from-a-ferry-journey
+ * Other values extrapolated.
+ */
+const CARBON_INTENSITY_DEFAULTS = {
+  walk: 0.026,
+  bicycle: 0.014,
+  car: 0.16,
```
demory, line 457: Cell AO98 (private car - ICE) shows 0.162 if we want to be precise. Is there a way to distinguish between "normal" driving and ride-hail? The latter is significantly higher (0.276! cell BI98)

demory (commented):
See inline comments on defaults from ITF report spreadsheet (https://www.itf-oecd.org/sites/default/files/life-cycle-assessment-calculations-2020.xlsx)

packages/core-utils/src/itinerary.ts
daniel-heppner-ibigroup: Sadly no, OTP doesn't differentiate between them in terms of mode.

demory (approved):

Issue #335, opened 2022-01-26
LocationField: transit landmarks, fare outlets and shared-use racks / hubs


fpurcell:
A change back in 2021 dropped support for Pelias records that have 'source' === 'transit', which are an other than 'layer' === 'station' or 'stop'.  Other 'transit' (source) records beyond stops and stations (gtfs), ala landmarks (csv), bikeshare hubs (gbfs) and fare outlets (json) are also loaded via the Pelias Transit loader.   

This ticket will add support back for other types of records under 'source' === 'transit'

PR #350, opened 2022-02-15
location field - other transit data + old browsers

### LINKED ISSUE #348: LocationField: replace the use of replaceAll() -- browser support
Use of replaceAll() was introduced to OTP-UI in August 2021.  

https://github.com/opentripplanner/otp-ui/blob/ac4557162e31db5653522505752669299abae98d/packages/location-field/src/index.tsx#L354

Browser support for replaceAll() appears to be confined to newer browser versions (circa mid-2020): https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll#browser_compatibility

This ticket will replace replaceAll() in OTP-UI, and bring support back for customers who may not have the latest and greatest browsers available (e.g., we've gotten 'complaints' about the TORA app from customers).   Suggest we also periodically check how well OTP-UI works on BrowserStack using older browsers, so that we don't inadvertently break things...   

e.g., firing up Storybook in Firefox 72 (released in Jan 2020 -- at least two of our customers use this browser, and complained our site doesn't work), shows we've got a problem on (slightly) older browsers:

Screen Shot 2022-02-11 at 1 15 47 PM

Screen Shot 2022-02-11 at 10 22 59 AM

fpurcell:
fixes #348 and #335

### DIFF
--- a/packages/location-field/src/index.js
+++ b/packages/location-field/src/index.js
@@ -365,7 +365,7 @@ class LocationField extends Component {
       featureIdComponents[1].length > 0
     ) {
       const operatorName = featureIdComponents[1]
-        .replaceAll(" ", "-")
+        .replace(/ /g, "-")
         .toLowerCase();
       classNames.push(`operator-${operatorName}`);
       operatorIcon = operatorIconMap[operatorName];
--- a/packages/location-field/src/index.js
+++ b/packages/location-field/src/index.js
@@ -359,7 +359,7 @@ class LocationField extends Component {
     let operatorIcon;
     // Operator only exists on transit features
     const featureIdComponents = source === "transit" && id.split("::");
-    if (featureIdComponents.length > 0) {
+    if (featureIdComponents.length > 0 && featureIdComponents[1]) {
       const operatorName = featureIdComponents[1]
         .replaceAll(" ", "-")
         .toLowerCase();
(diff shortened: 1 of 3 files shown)


packages/location-field/src/mocks/handlers.js
```diff
@@ -7,6 +7,7 @@ export default [
   rest.get(
     "https://ws-st.trimet.org/pelias/v1/autocomplete",
     (req, res, ctx) => {
+      // return res; /* note: remove the // to make autocomple work in the story */
```
miles-grant-ibigroup, line 10: When I use this line instead of the currently uncommented one I get "could not reach geocoder". I'm open to re-adding support for a live geocoder, maybe via a new story?

packages/location-field/src/index.js
```diff
@@ -359,9 +359,13 @@ class LocationField extends Component {
     let operatorIcon;
     // Operator only exists on transit features
     const featureIdComponents = source === "transit" && id.split("::");
-    if (featureIdComponents.length > 0) {
+    if (
+      featureIdComponents.length > 0 &&
+      featureIdComponents[1] &&
+      featureIdComponents[1].length > 0
```
miles-grant-ibigroup, line 365: We have optional chaining on, so I think we can do this instead!
```suggestion
     featureIdComponents.length > 0 && featureIdComponents?.[1].length > 0
```

miles-grant-ibigroup (changes requested):
Thanks for these changes! Happy to see regressions removed. Just one quick cleanup and one question!

packages/location-field/src/mocks/handlers.js
fpurcell: Hey Miles ... a new story would be great.
But maybe not for this PR...   Happy to delete my comment if need be.

fpurcell:
code has changed enough that will resubmit this in future PR

PR #351, opened 2022-02-17
fix(route-viewer-overlay): avoid crash on empty but not missing geometry

miles-grant-ibigroup:
This PR resolves crashes found in OTP-RR. We were checking for lack of geometry, but not empty geometry.

### DIFF
--- a/packages/route-viewer-overlay/src/index.js
+++ b/packages/route-viewer-overlay/src/index.js
@@ -62,7 +62,8 @@ class RouteViewerOverlay extends MapLayer {
         },
         []
       );
-      if (allPoints.length > 0) this.props.leaflet.map.fitBounds(allPoints);
+      if (allPoints.length > 0 && this.props.leaflet.map)
+        this.props.leaflet.map.fitBounds(allPoints);
     }
   }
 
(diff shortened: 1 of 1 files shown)


philip-cline (approved):
Nice fix!

packages/route-viewer-overlay/src/index.js
```diff
@@ -53,19 +53,17 @@ class RouteViewerOverlay extends MapLayer {
   // TODO: determine why the default MapLayer componentWillUnmount() method throws an error
   componentWillUnmount() {}
 
-  componentDidUpdate(prevProps) {
-    // if pattern geometry just finished populating, update the map points
-    if (
-      !isGeomComplete(prevProps.routeData) &&
```
binh-dam-ibigroup, line 59: Why was this condition removed?

binh-dam-ibigroup: Perhaps the condition needs to include checking whether the routeData has changed too.

miles-grant-ibigroup: Other than `path` which I don't think we need to expect changing after render all props that can change should trigger a re-center

binh-dam-ibigroup (approved):

packages/route-viewer-overlay/src/index.js
```diff
@@ -109,6 +107,10 @@ class RouteViewerOverlay extends MapLayer {
 }
 
 RouteViewerOverlay.propTypes = {
+  /**
+   * This boolean value allows disabling of map centering and panning.
+   */
+  allowMapCentering: PropTypes.boolean,
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Nice catch thank you!

PR #353, opened 2022-02-22
fix(transitive-overlay): avoid crash on slow leaflet load

miles-grant-ibigroup:
Transitive overlay was relying on the Leaflet map object being available pretty quickly. In some environments it takes some time to arrive, so this PR resolves the issue by skipping draw when the map object isn't available.

### DIFF
--- a/packages/transitive-overlay/src/leaflet-canvas-layer.js
+++ b/packages/transitive-overlay/src/leaflet-canvas-layer.js
@@ -119,7 +119,7 @@ L.CanvasLayer = (L.Layer ? L.Layer : L.Class).extend({
     //------------------------------------------------------------------------------
     drawLayer: function () {
         // -- todo make the viewInfo properties  flat objects.
-        if (!this._map) return
+        if (!this._map) return;
         var size   = this._map.getSize();
         var bounds = this._map.getBounds();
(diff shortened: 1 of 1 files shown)


packages/transitive-overlay/src/leaflet-canvas-layer.js
```diff
@@ -119,6 +119,7 @@ L.CanvasLayer = (L.Layer ? L.Layer : L.Class).extend({
     //------------------------------------------------------------------------------
     drawLayer: function () {
         // -- todo make the viewInfo properties  flat objects.
+        if (!this._map) return
```
binh-dam-ibigroup, line 122: Insert `;` after `return`.

binh-dam-ibigroup (approved):

packages/transitive-overlay/src/leaflet-canvas-layer.js
miles-grant-ibigroup: So easy to forget the luxuries of an eslint environment. Fixed in 9ad268ac61f7db821681691dd134176a5b3bba75!

daniel-heppner-ibigroup (approved):
Pretty complicated review, but after combing through all the code changes, it gets my approval.

PR #355, opened 2022-03-01
Use storybook-react-intl addon

binh-dam-ibigroup:
This PR will avoid the extra boiler plate code to load messages and provide an `<IntlProvider>` wrapper for each component.

As a result,
- the locale selector is now **on the Storybook toolbar**.
- some controls under the stories are moved or replaced.


binh-dam-ibigroup:
This is ready for review now.

.storybook/preview.js
```diff
@@ -43,4 +46,16 @@ export const parameters = {
       date: /Date$/,
     },
   },
-}
\ No newline at end of file
+  locale: reactIntl.defaultLocale,
+  locales: {
+    "en-US": { title: "English (US)", left: "🇺🇸" },
+    fr: { title: "Français", left: "🇫🇷" },
+    unknown: { title: "Unknown", left: "🚫" }
```
miles-grant-ibigroup, line 53: Why do we have this? On safari at least, it renders strings from a random language (?) instead of the keys which is what I'd assume this to do

.storybook/react-intl.js
```diff
@@ -0,0 +1,50 @@
...
+];
+
+/** Messages for all packages AND locales above. */
+const messages = {};
+
+if (!isRunningJest()) {
+  // Populate messages if not running snapshots.
+  // (Message printouts would be unnecessary replicated in snapshots without that check.)
+  packages.forEach((pkg) => {
+    locales.forEach((locale) => {
```
miles-grant-ibigroup, line 27: We're setting messages here, but we're also doing it here? https://github.com/opentripplanner/otp-ui/pull/355/files#diff-09d1644f1de8e7f7e06a9d8d0603adaca2c57c37b7c8abe25d324632793e469fR260 (not sure if that link worked -- in TripDetails at the bottom of the file)

I may be missing something -- could you possibly explain what's going on?

miles-grant-ibigroup (approved):
Well done getting this added so seamlessly! Thanks also for cleaning up the snapshots now. Looks much better.

I have a few quick questions about some small details

.storybook/preview.js
binh-dam-ibigroup: - Some components are set to render English if the string resource file doesn't exist for the specified locale, but other components may just render something very basic or fallback on a message id.
- For `TripDetails`, the fare portion is no part of the language strings, it is set via config, so for the story, the behavior is not defined if the fare strings have not been set for the locale.

.storybook/react-intl.js
binh-dam-ibigroup: In the `TripDetails` story, we demonstrate overriding the built-in messages with custom ones. I'll add a comment regarding that.

binh-dam-ibigroup: Covered in https://github.com/opentripplanner/otp-ui/pull/355/commits/bc143e233eb07e6fc967154384e6c1ef61c1696c.

.storybook/preview.js
miles-grant-ibigroup: This makes a lot of sense thanks!

packages/trip-form/src/components.story.tsx
```diff
@@ -11,23 +9,22 @@ import modeOptions from "./__mocks__/mode-options";
...
 
 // Events
 const onChange = action("onChange");
 const onClick = action("onClick");
 const onQueryParamChange = action("onQueryParamChange");
 
-const intlDecorator = (Story: StoryType): ReactElement => (
-  <IntlProvider locale="en-US" messages={flatten(englishMessages)}>
+const decorator = (Story: StoryType): ReactElement => (
+  <div>
```
daniel-heppner-ibigroup, line 20: Is a surrounding div necessary? Would <> work?

daniel-heppner-ibigroup (approved):
Looks great! Feels very proper.

packages/trip-form/src/components.story.tsx
binh-dam-ibigroup: That's possible indeed (see https://github.com/opentripplanner/otp-ui/pull/355/commits/2108869a6cf786fc78255b1bd12746ccf7523128).

PR #356, opened 2022-03-02
fix(transit-vehicle-overlay): support subway

miles-grant-ibigroup:
The transit vehicle overlay did not support subway icons and instead rendered them as a big dot. This renders them as trains.


daniel-heppner-ibigroup (approved):
I love trains.

philip-cline (approved):
<3

PR #358, opened 2022-03-03
Humanize distance intl

binh-dam-ibigroup:
~Requires #355 (for an updated correct dependency version).~ Update: #355 is now merged in master.

This PR adds i18n support to the `humanize-distance` package while maintaining backwards compatibility (use of `react-intl` remains optional with this package).

### DIFF
--- a/packages/humanize-distance/src/index.ts
+++ b/packages/humanize-distance/src/index.ts
@@ -11,25 +11,21 @@ export function humanizeDistanceStringImperial(
 ): string {
   const feet = meters * 3.28084;
 
-  let unit;
-  let value;
-  let unitIfNoIntl;
+  let unit = "mile";
+  let unitIfNoIntl = abbreviate ? "mi" : "miles";
+  let value = roundToOneDecimalPlace(feet / 5280);
 
   if (feet < 528) {
     unit = "foot";
     unitIfNoIntl = abbreviate ? "ft" : "feet";
     value = Math.round(feet);
-  } else {
-    unit = "mile";
-    unitIfNoIntl = abbreviate ? "mi" : "miles";
-    value = roundToOneDecimalPlace(feet / 5280);
   }
 
   return intl
     ? intl.formatNumber(value, {
+        style: "unit",
         unit,
-        unitDisplay: abbreviate ? "short" : "long",
-        style: "unit"
+        unitDisplay: abbreviate ? "short" : "long"
       })
     : `${value} ${unitIfNoIntl}`;
 }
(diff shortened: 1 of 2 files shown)


binh-dam-ibigroup:
With #355 merged, this PR is now ready for review @philip-cline cc: @fpurcell.

philip-cline (approved):

packages/humanize-distance/src/index.ts
```diff
@@ -1,32 +1,74 @@
...
+  let unitIfNoIntl;
+
+  if (feet < 528) {
+    unit = "foot";
+    unitIfNoIntl = abbreviate ? "ft" : "feet";
+    value = Math.round(feet);
+  } else {
+    unit = "mile";
+    unitIfNoIntl = abbreviate ? "mi" : "miles";
+    value = roundToOneDecimalPlace(feet / 5280);
```
miles-grant-ibigroup, line 25: We could probably clean this up a bit if we used these values as the default case for those `let`s

packages/humanize-distance/src/index.ts
```diff
@@ -1,32 +1,74 @@
...
+    value = Math.round(km);
+  } else if (km > 1) {
     // 1.1 km => 99.9 km
-    return `${km.toFixed(1)} km`;
+    value = roundToOneDecimalPlace(km);
+  } else {
+    unit = "meter";
+    shortUnit = "m";
+    value = Math.round(meters);
   }
```
miles-grant-ibigroup, line 55: We could probably clean this up a bit if we used these values as the default case for those `let`s

packages/humanize-distance/src/index.ts
```diff
@@ -1,32 +1,74 @@
...
-  if (feet < 528)
-    return Math.round(feet) + (abbreviate === true ? " ft" : " feet");
-  return Math.round(feet / 528) / 10 + (abbreviate === true ? " mi" : " miles");
+
+  let unit;
+  let value;
+  let unitIfNoIntl;
+
+  if (feet < 528) {
+    unit = "foot";
```
miles-grant-ibigroup, line 19: Am I right in thinking that `formatMessage` will re-write `foot`?

miles-grant-ibigroup (approved):
More than I expected but looks good to me and works well in my (limited) testing!

packages/humanize-distance/src/index.ts
binh-dam-ibigroup: It will also use the correct decimal separator.

packages/humanize-distance/src/index.ts
binh-dam-ibigroup: Sure, addressed in https://github.com/opentripplanner/otp-ui/pull/358/commits/9dcf6e7f50034fd0c9f980005e1637eb4e4c3796.

packages/humanize-distance/src/index.ts
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-ui/pull/358/commits/9dcf6e7f50034fd0c9f980005e1637eb4e4c3796.

PR #361, opened 2022-03-22
fix(geocoder): fix chainquery bug with HERE boundary filter

daniel-heppner-ibigroup:
This is a one line fix for the Geocoder package, which was sometimes failing to correctly generate correct results using HERE when a boundary was provided along with focusPoint. Most of the changes in this PR was enhancements for the story to make testing easier.

### DIFF
--- a/packages/geocoder/src/apis/here/index.ts
+++ b/packages/geocoder/src/apis/here/index.ts
@@ -42,6 +42,7 @@ function run({ options, query, url }: HereFetchArgs): Promise<HereResponse> {
 }
 
 const checkItemInBoundary = ({ rect }: Boundary) => ({ position }: Item) => {
+  if (!position) return true // Chain queries might not have position, ignore them
   const { maxLat, maxLon, minLat, minLon } = rect
   const { lat, lng } = position
   return lng <= maxLon && lng >= minLon && lat <= maxLat && lat >= minLat
--- a/packages/geocoder/src/index.story.tsx
+++ b/packages/geocoder/src/index.story.tsx
@@ -1,4 +1,5 @@
 import React, { ReactElement, useState } from "react";
+import { AnyGeocoderQuery } from "./geocoders/types";
 import getGeocoder from "./index";
 
 const GeocoderTester = ({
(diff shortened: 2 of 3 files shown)


packages/geocoder/src/index.story.tsx
```diff
@@ -39,21 +48,23 @@ const GeocoderTester = ({
...
+  const searchObj: AnyGeocoderQuery = {
+    text: searchTerm
+  };
+
+  if (enableBoundary) {
+    searchObj.boundary = { rect: boundary };
+  }
+  if (enableFocusPoint) {
+    searchObj.point = focusPoint;
+    searchObj.focusPoint = focusPoint;
```
miles-grant-ibigroup, line 60: We update `point` and `focusPoint`?

packages/geocoder/src/apis/here/index.ts
```diff
@@ -42,6 +42,7 @@ function run({ options, query, url }: HereFetchArgs): Promise<HereResponse> {
 }
 
 const checkItemInBoundary = ({ rect }: Boundary) => ({ position }: Item) => {
+  if (!position) return true // Chain queries might not have position, ignore them
```
miles-grant-ibigroup, line 45: This may be me not fully understanding how HERE responses work, but why do we mark unknown position queries as being in the bounding box? Is this to avoid a situation with 0 results returned?

miles-grant-ibigroup (approved):
Looks really good! Loving the immense power of the story for this. Thanks for finding the fix!

Have a few questions but nothing huge

packages/geocoder/src/apis/here/index.ts
daniel-heppner-ibigroup: There are those weird chain query responses, and in most cases they get removed in a later stage. It doesn't really matter if this is true or false, since they'll get removed later, but I set it to true just to be consistent in case we want to use them at some point in the future.

packages/geocoder/src/index.story.tsx
daniel-heppner-ibigroup: I think it's because point is used for reverse and focusPoint is used for others. But since the story just has the one field I just set both of them in the query.

miles-grant-ibigroup: That makes sense thanks

packages/geocoder/src/apis/here/index.ts
miles-grant-ibigroup: Sounds good! Agreed better to keep them then to remove them

binh-dam-ibigroup (approved):
LGTM

PR #365, opened 2022-03-30
Remove use of deprecated `core-utils` methods

miles-grant-ibigroup:
This PR removes use of deprecated methods, or adds notices not to use deprecated methods. 

Blocked by #364

### DIFF
--- a/packages/core-utils/src/deprecated.js
+++ b/packages/core-utils/src/deprecated.js
@@ -1,7 +1,14 @@
 import moment from "moment";
 
-// Need to avoid cyclic dependency resolution
+/**
+ * To disable cyclic dependency resolution we need to require() within methods
+ * This is a good reason to disable this eslint-rule
+ */
 /* eslint-disable global-require */
+
+/**
+ * Generates a warning to tell developer that they are using deprecated methods!
+ */
 export function logDeprecationWarning(method, alternative) {
   console.warn(
     `${method ||
--- a/packages/core-utils/src/deprecated.js
+++ b/packages/core-utils/src/deprecated.js
@@ -1,0 +1,316 @@
+import moment from "moment";
+
+// Need to avoid cyclic dependency resolution
+/* eslint-disable global-require */
+export function logDeprecationWarning(method, alternative) {
+  console.warn(
+    `${method ||
+      "This method"} is deprecated and will be removed in a future otp-ui release. All language functionality should be handled using react-intl.
+        ${
+          alternative
+            ? `
+
+        Use ${alternative} instead, which provides a new interface that doesn't return English strings.`
+            : ""
+        }`
+  );
+}
+
+// itinerary.js
+
+export function getStepDirection(step) {
+  logDeprecationWarning("getStepDirection");
+
+  switch (step.relativeDirection) {
+    case "DEPART":
+      return `Head ${step.absoluteDirection.toLowerCase()}`;
+    case "LEFT":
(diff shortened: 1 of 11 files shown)


daniel-heppner-ibigroup (approved):

packages/core-utils/src/itinerary.js
```diff
@@ -159,51 +181,6 @@ export function getMapColor(mode) {
...
-  return `${getStepDirection(step)} ${conjunction} ${step.streetName}`;
-}
-
-export function getStepStreetName(step) {
-  if (step.streetName === "road") return "Unnamed Road";
-  if (step.streetName === "path") return "Unnamed Path";
-  return step.streetName;
-}
-
 export function toSentenceCase(str) {
```
binh-dam-ibigroup, line 184: This one might need to be deprecated.

packages/core-utils/src/profile.js
```diff
@@ -27,8 +27,8 @@ export function filterProfileOptions(response) {
   return response;
 }
 
-function locationString(str, defaultStr) {
-  return str ? str.split(",")[0] : defaultStr;
+function locationString(str) {
+  return str ? str.split(",")[0] : null;
```
binh-dam-ibigroup, line 31: `str?.split...`

packages/core-utils/src/query-params.js
```diff
@@ -1,3 +1,7 @@
+// TODO: Remove this entire file, as it is deprecated in favor of i18n-queryParams within
```
binh-dam-ibigroup, line 1: Why don't we move the deprecated contents to `deprecated.js`?

packages/core-utils/src/time.js
```diff
@@ -92,8 +97,8 @@ export function formatDuration(seconds) {
  * @param {number} seconds duration in seconds
  * @returns {string} formatted text representation
```
binh-dam-ibigroup, line 98: Add new param to JSdoc.

packages/core-utils/src/itinerary.js
miles-grant-ibigroup: There are a bunch of otp-rr keys that rely on this method to be generated. We could go through and replace all those keys, but we'd still need some form of string recapitalization/normalization. What do you think? The only other use from within otp-ui is in a deprecated method

packages/core-utils/src/query-params.js
miles-grant-ibigroup: There's an upcoming PR that just deletes this file so I didn't see the value in moving everything over

packages/core-utils/src/deprecated.js
```diff
@@ -0,0 +1,316 @@
+import moment from "moment";
+
+// Need to avoid cyclic dependency resolution
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Added in 607cb0f2263907b88396ccd78574129c94d0b1a2

packages/core-utils/src/profile.js
miles-grant-ibigroup: Refactored in 607cb0f2263907b88396ccd78574129c94d0b1a2

packages/core-utils/src/time.js
miles-grant-ibigroup: Added in 607cb0f2263907b88396ccd78574129c94d0b1a2

PR #369, opened 2022-04-06
fix(location-field): avoid using geocoder generated name if possible

miles-grant-ibigroup:
This PR fixes the bug of duplicate zip codes and states. It turns out the HERE -> Geocode.Earth conversion includes that data in the `name` field which is then being re-appended by the location-field. This is addressed by generating the address name ourselves.


philip-cline (approved):

packages/location-field/src/utils.ts
```diff
@@ -21,10 +21,12 @@ const zipOrNeighborhood = ({
 // more layer types with custom rendering.
 const layerDisplayMap = {
   address: (properties: Properties): Label => {
-    const { locality, name, postalcode, region_a: state } = properties;
+    const { housenumber, locality, name, street, postalcode, region_a: state, region } = properties;
```
binh-dam-ibigroup, line 24: Sort props

binh-dam-ibigroup (approved):
LGTM, it would be nice to see how the house numbers render in storybook.

packages/location-field/src/utils.ts
miles-grant-ibigroup: Sorted in 0ee6e8c85ef16d0a68531803f17a22c393c8f5e2

miles-grant-ibigroup:
The mock autocomplete results are updated in 0ee6e8c85ef16d0a68531803f17a22c393c8f5e2

PR #370, opened 2022-04-08
Add walk reluctance to `core-utils`

miles-grant-ibigroup:
A stub was present, but this PR adds a full entry for it. A new `otp2` field in the query object is used to switch between walk reluctance and the max walk distance, which is no longer supported by OTP2.


binh-dam-ibigroup (approved):
Thanks for adding the clarifications.

philip-cline (approved):

PR #374, opened 2022-04-14
better screen reader support

miles-grant-ibigroup:
Some packages had elements which caused issues with screen readers. This PR addresses these by adding relevant aria tags.

### TODO:
~- [ ] icons are still rendered as svg without aria labels~
- [ ] correctly apply aria tags via styled components
- [ ] `transitFareWrapper` needs to be an aria button only when it has children.
- [ ] `transitFareWrapper` needs to be adjusted to match html spec
- [ ] internal `core-utils` versions need to be updated

### DIFF
--- a/packages/itinerary-body/src/AccessLegBody/index.tsx
+++ b/packages/itinerary-body/src/AccessLegBody/index.tsx
@@ -104,7 +104,7 @@ class AccessLegBody extends Component<Props, State> {
             showLegIcon={showLegIcon}
           />
           <S.StepsHeader
-            aria-roledescription="dropdown"
+            role="button"
             aria-expanded={expanded}
             onClick={this.onStepsHeaderClick}
           >
--- a/packages/itinerary-body/src/AccessLegBody/index.tsx
+++ b/packages/itinerary-body/src/AccessLegBody/index.tsx
@@ -104,6 +104,7 @@ class AccessLegBody extends Component<Props, State> {
             showLegIcon={showLegIcon}
           />
           <S.StepsHeader
+            aria-roledescription="dropdown"
             aria-expanded={expanded}
             onClick={this.onStepsHeaderClick}
           >
--- a/packages/trip-details/src/index.tsx
+++ b/packages/trip-details/src/index.tsx
@@ -182,7 +182,7 @@ export function TripDetails({
       <S.Fare>
         <TransitFareWrapper>
           <summary
-            aria-roledescription="dropdown"
+            role="button"
             style={{ display: fareKeys.length > 1 ? "list-item" : "" }}
           >
             <TransitFare
--- a/packages/trip-details/src/index.tsx
+++ b/packages/trip-details/src/index.tsx
@@ -181,7 +181,10 @@ export function TripDetails({
     fare = (
       <S.Fare>
         <TransitFareWrapper>
-          <summary style={{ display: fareKeys.length > 1 ? "list-item" : "" }}>
+          <summary
+            aria-roledescription="dropdown"
+            style={{ display: fareKeys.length > 1 ? "list-item" : "" }}
+          >
             <TransitFare
               fareNameFallback={
                 <FormattedMessage
--- a/packages/trip-form/src/DateTimeSelector/index.tsx
+++ b/packages/trip-form/src/DateTimeSelector/index.tsx
@@ -220,6 +220,7 @@ export default function DateTimeSelector({
         {departureOptions.map(opt => (
           <ModeButton
             key={opt.type}
+            uniqueId={opt.type}
             onClick={setDepartArrive(opt)}
             selected={opt.isSelected}
           >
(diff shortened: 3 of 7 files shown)


packages/trip-form/src/SubmodeSelector/index.tsx
```diff
@@ -57,9 +57,10 @@ export default function SubmodeSelector({
       {label && <LabelType>{label}</LabelType>}
       <RowType>
         {modes &&
-          modes.map(option => (
+          modes.map(( option, index ) => (
             <ModeButton
               key={option.id}
+              uniqueId={option.id + index}
```
binh-dam-ibigroup, line 63: Sort props

packages/trip-form/src/ModeSelector/index.tsx
```diff
@@ -53,6 +53,7 @@ export default function ModeSelector({
   const makeButton = (option: ModeSelectorOption): ReactElement => (
     <ModeButton
       key={option.id}
+      uniqueId={option.id}
```
binh-dam-ibigroup, line 56: Sort props

packages/trip-form/src/ModeButton/index.tsx
```diff
@@ -53,25 +57,31 @@ export default function ModeButton({
...
   return (
     <S.ModeButton className={className} style={style}>
       <S.ModeButton.Button
         aria-pressed={selected}
+        aria-label={title}
         className={`${activeClassName} ${disabledClassName}`}
         disabled={!enabled}
         onClick={onClick}
         title={title}
+        id={buttonId}
```
binh-dam-ibigroup, line 77: Sort props

packages/trip-form/src/ModeButton/index.tsx
```diff
@@ -53,25 +57,31 @@ export default function ModeButton({
...
   const activeClassName = selected ? "active" : "";
   const disabledClassName = enabled ? "" : "disabled";
 
+  const buttonId = `${title?.replace(/ /g, "-")}-button-${uniqueId}`;
+
   return (
     <S.ModeButton className={className} style={style}>
       <S.ModeButton.Button
         aria-pressed={selected}
+        aria-label={title}
```
binh-dam-ibigroup, line 72: Sort props

packages/trip-form/src/DateTimeSelector/index.tsx
```diff
@@ -220,6 +220,7 @@ export default function DateTimeSelector({
         {departureOptions.map(opt => (
           <ModeButton
             key={opt.type}
+            uniqueId={opt.type}
```
binh-dam-ibigroup, line 223: Sort props

packages/trip-details/src/index.tsx
```diff
@@ -181,7 +181,10 @@ export function TripDetails({
     fare = (
       <S.Fare>
         <TransitFareWrapper>
-          <summary style={{ display: fareKeys.length > 1 ? "list-item" : "" }}>
+          <summary
+            role="button"
```
binh-dam-ibigroup, line 185: Add an `aria-expanded` attribute. 

I would disagree with `role="button"` for the `<summary>` element, so maybe we shouldn't use `<summary>`? Also, in another PR, we'll need to implement `aria-controls` or `aria-owns` for this expandable component.

packages/itinerary-body/src/AccessLegBody/index.tsx
```diff
@@ -104,6 +104,7 @@ class AccessLegBody extends Component<Props, State> {
             showLegIcon={showLegIcon}
           />
           <S.StepsHeader
+            role="button"
```
binh-dam-ibigroup, line 107: Is this needed? It is already a `<button>`.

binh-dam-ibigroup: Do we need an `aria-controls` or `aria-owns` attribute?

binh-dam-ibigroup (approved):
Approving subject to sort props and adding an `aria-expanded` attribute where noted.

packages/itinerary-body/src/AccessLegBody/index.tsx
miles-grant-ibigroup: Yes the `role="button"` seems to not be needed, but `aria-expanded` is what seems to be needed.

packages/trip-details/src/index.tsx
miles-grant-ibigroup: I followed https://codepen.io/scottohara/pen/aaJXYG. I think this is very much a summary but also think that there may be a better way than a random codepen to make it properly accessible

binh-dam-ibigroup: The component that reacts to clicking should be the button element... but I'm not sure which one it is!

binh-dam-ibigroup: And we'll have to fix this. Basically, the `<summary>` should be outside of `<details>`.
image

PR #375, opened 2022-04-15
fix(VehicleRentalOverlay): Listen to zoom change events

binh-dam-ibigroup:


miles-grant-ibigroup (approved):
Works great thanks for the fix! Good call on merging in that layer control interface too.

philip-cline (approved):

PR #378, opened 2022-04-18
feat(FromToLocationPicker): Add label prop.

binh-dam-ibigroup:
This PR adds a `label` prop to the `FromToLocationPicker` component. If `label` is set to `true`, a default i18n-able label text is shown, that will simplify reusing message strings code in other packages in this repo and others that implement this component.


miles-grant-ibigroup (approved):
Looks and works great

philip-cline (approved):

Issue #373, opened 2022-04-14
default max walk distance


fpurcell:
the maxWalkDistance is currently hard-coded in core-uilts as 3/4 mile (1207 meters), which is something we'd like to change (given some customer complaints)

desired: make the _default_ maxWalkDistance something configurable, rather than hard-coding the value

alt: change the hard-coded _default_ value to 1 mile (1609 meters)

related to: #335 and #348

fpurcell:
old ... cleanup

PR #383, opened 2022-04-20
some geocoder search form and trip planner form fixes.

### LINKED ISSUE #335: LocationField: transit landmarks, fare outlets and shared-use racks / hubs

### LINKED ISSUE #348: LocationField: replace the use of replaceAll() -- browser support

fpurcell:
- closes #348 - replaceAll() is too new a javascript function
- closes #335 - don't filter other 'transit' records from Pelias (ala transit landmarks, routes, fare, etc...)
- addresses 373 - default max walk distance (default is 1 mile)...future improvement could make this configurable rather than hard coded (but we want 1 mile as the default to TORA, and hard coding the value in core-utils is the only current way to change that)


miles-grant-ibigroup (approved):
Looks and works great thanks for all the fixes!

binh-dam-ibigroup:
Is the solution for #373 only partial?

packages/core-utils/src/query-params.js
```diff
@@ -176,7 +176,7 @@ const queryParams = [
       query.mode &&
       hasTransit(query.mode) &&
       query.mode.indexOf("WALK") !== -1,
-    default: 1207, // 3/4 mi.
+    default: 1609, // 1 mi.
```
binh-dam-ibigroup, line 179: This change will affect existing OTP-RR deployments where the default walk distance is 3/4 mile. You can change the default walking distance to be specific to you implementation if you use `getCustomQueryParams` from `core-utils/query-params` with your overrides. Please see https://github.com/opentripplanner/otp-ui/blob/b3260af2279671825167cc35f36f56da5de0eebd/packages/core-utils/src/query-params.js#L911

binh-dam-ibigroup (changes requested):
Requesting to customize the default `queryParams` as indicated in https://github.com/opentripplanner/otp-ui/pull/383/files#r859196276.

packages/core-utils/src/query-params.js
fpurcell: Binh, using that creaky API call won't work for us (already looked).  I'm sure your implementations can change to 1 mile as a default, and not miss a beat.  This is a library that TriMet funded, and we want 1 mile as the default.  If IBI wants to go the extra distance to make this configurable (https://github.com/opentripplanner/otp-ui/issues/373), that's up to you.

fpurcell:
> Is the solution for #373 only partial?

No...373 is not addressed here.  1 mile is the default we want.

packages/core-utils/src/query-params.js
binh-dam-ibigroup: Hi @fpurcell, IBI has the ability to customize the default walk distance that way (but we'd have to make changes to OTP-RR after this PR is merged), so I'm fine to approve.

binh-dam-ibigroup (approved):
Conditional approval upon resolving merge conflicts.

fpurcell:
> Conditional approval upon resolving merge conflicts.

Thanks Binh...think this is ready to merge, as it is rebased, with conflicts resolved.

binh-dam-ibigroup:
Hi @fpurcell, feel free to merge this if you are done with changes.

PR #384, opened 2022-04-21
Vehicle rental i18n only

binh-dam-ibigroup:
Requires/blocked by #375, #378.

This PR adds i18n support to the vehicle rental overlay.
Typescript is not included in this PR because of intricated issues TypeScript-ing map components.

For what TypeScript-ing would involve, see https://github.com/ibi-group/otp-ui/compare/vehicle-rental-i18n-only...ibi-group:vehicle-rental-i18n?expand=1.


philip-cline (approved):

miles-grant-ibigroup (approved):
Looks great thanks so much!

PR #387, opened 2022-04-26
update internal dependency versions

miles-grant-ibigroup:
Some package updates aren't working in otp-rr due to missing internal dependency updates. This PR resolves that.

### DIFF
--- a/packages/core-utils/package.json
+++ b/packages/core-utils/package.json
@@ -13,7 +13,7 @@
   "private": false,
   "dependencies": {
     "@mapbox/polyline": "^1.1.0",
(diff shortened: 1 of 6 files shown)


daniel-heppner-ibigroup (changes requested):
This is causing yarn.lock to update for me when I run yarn. I deleted node_modules and started over but still got a change to yarn.lock.

miles-grant-ibigroup:
Have committed a new lock file, let me know if it works for you

daniel-heppner-ibigroup (approved):
Thank you!

philip-cline (approved):
Looking good to me too

PR #388, opened 2022-04-27
fix(VehicleRentalOverlay): Ensure rendering when displaying for first time

### LINKED ISSUE #376: VehicleRentalOverlay should listen to zoom change
The implementation of `VehicleRentalOverlay` should update rendering when the zoom level on the parent map changes (it currently doesn't - In storybook, an additional component is currently used to communicate zoom changes to `VehicleRentalOverlay`).

binh-dam-ibigroup:
Fix #376 again, this time on first render when the user initially clicks the layer (copy `lib` or `esm` to OTP-RR to see that in action).

### DIFF
--- a/packages/vehicle-rental-overlay/src/index.js
+++ b/packages/vehicle-rental-overlay/src/index.js
@@ -32,7 +32,7 @@ class VehicleRentalOverlay extends MapLayer {
   constructor(props) {
     super(props);
     this.state = {
-      zoom: props.leaflet.map.getZoom()
+      zoom: null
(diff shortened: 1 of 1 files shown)


miles-grant-ibigroup (approved):
I've been running into this bug for a while. Can't believe the fix is this easy! Is this perhaps related to the bike layer issues we've been having?

daniel-heppner-ibigroup (approved):
Nice work on this, the Leaflet state stuff can be so tricky!

PR #389, opened 2022-04-27
fix(ParkAndRideOverlay): Use i18n from FromToLocationPicker 2.1.0.

binh-dam-ibigroup:
This PR makes the P/R overlay to use i18n from `from-to-location-picker` 2.1.0.
Typescript is not included in this PR because of issues with react-leaflet's `MapLayer` class.
A new "fix" release will be created when merging this PR.

### DIFF
--- a/packages/park-and-ride-overlay/package.json
+++ b/packages/park-and-ride-overlay/package.json
@@ -19,7 +19,7 @@
   },
   "dependencies": {
     "@opentripplanner/core-utils": "^4.5.0",
-    "@opentripplanner/from-to-location-picker": "^1.3.0",
+    "@opentripplanner/from-to-location-picker": "^2.1.0",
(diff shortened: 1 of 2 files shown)


daniel-heppner-ibigroup (approved):
Works fine for me!

miles-grant-ibigroup (approved):
Not getting the popup to appear in the story, but that doesn't seem to be related to this PR.

binh-dam-ibigroup:
> Not getting the popup to appear in the story, but that doesn't seem to be related to this PR.

I see the popup when clicking on the "P" symbol in storybook. Are you referring to something else?

binh-dam-ibigroup:
> Yes when I click the P nothing happens. Might be a safari issue though

Indeed. The popup works with Chrome and Firefox. Filing a bug.

PR #390, opened 2022-04-27
Fare by Leg in `trip-details`

miles-grant-ibigroup:
This PR will add a new fare details component within `trip-details` that displays a table of fares for the legs in the itinerary.

### DIFF
--- a/packages/trip-details/src/TripDetails.story.tsx
+++ b/packages/trip-details/src/TripDetails.story.tsx
@@ -14,11 +14,11 @@ import * as TripDetailsClasses from "./styled";
 import {
   CaloriesDetailsProps,
   DepartureDetailsProps,
-  FareDetailsProps,
   TripDetailsProps
 } from "./types";
 
 import customMessages from "../__mocks__/custom-messages.yml";
+import { FareDetails } from "./fare-detail";
 
 // import mock itinaries. These are all trip plan outputs from OTP.
 const bikeOnlyItinerary = require("@opentripplanner/itinerary-body/src/__mocks__/itineraries/bike-only.json");
@@ -85,18 +85,6 @@ const CustomDepartureDetails = ({
   </>
 );
 
-const CustomFareDetails = ({
-  maxTNCFare,
-  minTNCFare,
-  transitFares
-}: FareDetailsProps): ReactElement => (
-  <>
-    Custom details about fares (transitFares: {JSON.stringify(transitFares)}{" "}
-    (cents), minTNCFare: {minTNCFare} and maxTNCFare: {maxTNCFare} can be
-    constructed dynamically using any markup.
-  </>
-);
-
 const CustomCaloriesDetails = ({
   bikeSeconds,
   calories,
@@ -122,7 +110,6 @@ function createTripDetailsTemplate(
       CaloriesDetails,
       defaultFare,
       DepartureDetails,
-      FareDetails,
       itinerary
     }: TripDetailsProps,
     { globals, parameters }: StoryContext
@@ -139,7 +126,6 @@ function createTripDetailsTemplate(
         CaloriesDetails={CaloriesDetails}
         defaultFare={defaultFare}
         DepartureDetails={DepartureDetails}
-        FareDetails={FareDetails}
         fareKeyNameMap={fareKeyNameMap}
         itinerary={itinerary}
       />
@@ -246,12 +232,11 @@ export const TncTransitItinerary = makeStory(
 );
 
 const flattenedMessages = flatten(customMessages);
-export const TncTransitItineraryWithCustomMessagesAndDetails = makeStory(
+export const TncTransitItineraryWithCustomMessages = makeStory(
   {
     CaloriesDetails: CustomCaloriesDetails,
     defaultFareKey: "electronicRegular",
     DepartureDetails: CustomDepartureDetails,
-    FareDetails: CustomFareDetails,
     itinerary: tncTransitTncItinerary
   },
   {
@@ -271,3 +256,11 @@ export const TncTransitItineraryWithCustomMessagesAndDetails = makeStory(
 export const FlexItinerary = makeStory({
   itinerary: fareComponentsItinerary
 });
+
+export const FareDetailsComponent = () => (
+  <FareDetails
+    transitFares={{}}
+    legs={[]}
+    prefixes={{ electronic: "ORCA", cash: "Cash" }}
+  />
+);
--- a/packages/trip-details/src/fare-detail.tsx
+++ b/packages/trip-details/src/fare-detail.tsx
@@ -1,0 +1,8 @@
+/* eslint-disable import/prefer-default-export */
+import React from "react";
+import { FareDetailsProps } from "./types";
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const FareDetails = (props: FareDetailsProps): JSX.Element => (
+  <b>hello</b>
+);
--- a/packages/trip-details/src/index.tsx
+++ b/packages/trip-details/src/index.tsx
@@ -23,6 +23,7 @@ import {
 
 // Load the default messages.
 import defaultEnglishMessages from "../i18n/en-US.yml";
+import { FareDetails } from "./fare-detail";
 
 // HACK: We should flatten the messages loaded above because
 // the YAML loaders behave differently between webpack and our version of jest:
@@ -149,7 +150,6 @@ export function TripDetails({
   className = "",
   DepartureDetails = null,
   defaultFareKey = "regular",
-  FareDetails = null,
   fareKeyNameMap = {},
   itinerary
 }: TripDetailsProps): ReactElement {
@@ -291,12 +291,13 @@ export function TripDetails({
         {fare && (
           <TripDetail
             // Any custom description for the transit fare needs to be handled by the slot.
+            //  TODO: add leg info
             description={
-              FareDetails && (
+              Object.keys(fareKeyNameMap).length > 0 && (
                 <FareDetails
-                  maxTNCFare={maxTNCFare}
-                  minTNCFare={minTNCFare}
                   transitFares={transitFares}
+                  legs={[]}
+                  prefixes={{}}
                 />
               )
             }
--- a/packages/trip-details/src/types.ts
+++ b/packages/trip-details/src/types.ts
@@ -1,6 +1,6 @@
 // Prettier does not recognize the import type syntax.
 // eslint-disable-next-line prettier/prettier
-import type { Itinerary } from "@opentripplanner/types";
+import type { Itinerary, Leg, Fare } from "@opentripplanner/types";
 import type { ReactElement } from "react";
 
 export interface CaloriesDetailsProps {
@@ -13,22 +13,12 @@ export interface DepartureDetailsProps {
   departureDate: Date;
 }
 
-export interface TransitFareData {
-  [key: string]: {
-    currency: {
-      currencyCode: string
-      defaultFractionDigits: number
-      currency: string
-      symbol: string
-    };
-    cents: number;
-  }
-}
 
 export interface FareDetailsProps {
-  maxTNCFare: number;
-  minTNCFare: number;
-  transitFares: TransitFareData;
+  transitFares: Fare;
+  legs: Leg[]
+  // To make this i18n friendly, set the string to be a react i18n key
+  prefixes: { [ prefix: string ]:  string }
 }
 
 export interface TransitFareProps {
@@ -37,7 +27,7 @@ export interface TransitFareProps {
   fareKeyNameMap: {
     [key: string]: string;
   };
-  transitFares: TransitFareData;
+  transitFares: Fare;
 }
 
 export interface TripDetailsProps {
@@ -57,10 +47,6 @@ export interface TripDetailsProps {
    * Slot for a custom component to render the expandable section for departure.
    */
   DepartureDetails?: React.ElementType<DepartureDetailsProps>;
-  /**
-   * Slot for a custom component to render the expandable section for fares.
-   */
-  FareDetails?: React.ElementType<FareDetailsProps>;
   /**
    * Mapping between fare keys and human-readable names for them.
    */
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -340,11 +340,12 @@ type ApplicableId = string | FeedScopedId;
  * not any bike rental or TNC rental fees.
  */
 export interface Fare {
-  details: {
+  details?: {
     [name: string]: {
-      fareId: ApplicableId;
+      fareId?: ApplicableId;
       price: Money;
-      routes: ApplicableId[];
+      legIndex?: number;
+      routes?: ApplicableId[];
     }[];
   };
   fare?: {
--- a/packages/itinerary-body/src/demos/index.tsx
+++ b/packages/itinerary-body/src/demos/index.tsx
@@ -65,7 +65,7 @@ export function CustomTransitLegSummary({
     /* eslint-disable jsx-a11y/no-static-element-interactions, jsx-a11y/click-events-have-key-events */
     <div onClick={onClick}>
       {leg.duration && (
-        <span>Ride {coreUtils.time.formatDuration(leg.duration)}</span>
+        <span>Ride {coreUtils.time.formatDuration(leg.duration, "en-US")}</span>
       )}
       {leg.intermediateStops && (
         <span>
--- a/packages/location-field/src/types.ts
+++ b/packages/location-field/src/types.ts
@@ -11,10 +11,10 @@ export type ResultType =
   | "STOP";
 export type Location = {
   id?: string;
-  lat?: number;
-  lon?: number;
+  lat: number;
+  lon: number;
   main?: string;
-  name?: string;
+  name: string;
   secondary?: string;
 };
 
--- a/packages/trip-details/package.json
+++ b/packages/trip-details/package.json
@@ -17,7 +17,7 @@
     "velocity-react": "^1.4.3"
   },
   "devDependencies": {
-    "@opentripplanner/types": "^1.1.0",
+    "@opentripplanner/types": "^1.2.2",
     "@types/flat": "^5.0.2"
   },
   "peerDependencies": {


philip-cline (approved):

packages/trip-details/src/types.ts
```diff
@@ -57,16 +53,16 @@ export interface TripDetailsProps {
...
   /**
    * Mapping between fare keys and human-readable names for them.
    */
   fareKeyNameMap?: {
     [name: string]: string;
   };
+  /**
+   * Column and table configuration for fare details/fare by leg table.
+   */
+  fareDetailsLayout?: FareDetailsLayout[];
```
binh-dam-ibigroup, line 65: Sort props

packages/trip-details/src/types.ts
```diff
@@ -13,22 +13,18 @@ export interface DepartureDetailsProps {
...
+    header: string;
+  }[];
 }
 
 export interface FareDetailsProps {
-  maxTNCFare: number;
-  minTNCFare: number;
-  transitFares: TransitFareData;
+  transitFares: Fare;
+  legs: Leg[];
```
binh-dam-ibigroup, line 26: Sort props

packages/types/src/index.ts
```diff
@@ -340,11 +340,13 @@ type ApplicableId = string | FeedScopedId;
...
  */
 export interface Fare {
-  details: {
+  details?: {
     [name: string]: {
-      fareId: ApplicableId;
+      fareId?: ApplicableId;
       price: Money;
-      routes: ApplicableId[];
+      legIndex?: number;
```
binh-dam-ibigroup, line 347: Sort props

packages/trip-details/src/index.tsx
```diff
@@ -291,12 +292,14 @@ export function TripDetails({
...
             description={
-              FareDetails && (
+              fareDetailsLayout &&
+              Object.keys(itinerary.fare.details).length > 0 && (
                 <FareDetails
-                  maxTNCFare={maxTNCFare}
-                  minTNCFare={minTNCFare}
-                  transitFares={transitFares}
+                  transitFares={itinerary.fare}
+                  legs={itinerary.legs}
```
binh-dam-ibigroup, line 301: Sort props

packages/trip-details/src/fare-detail.tsx
```diff
@@ -0,0 +1,154 @@
...
+      };
+    })
+    .filter(leg => leg.transitLeg);
+
+  return (
+    <div>
+      {layout.map(config => (
+        <FareTypeTable
+          header={config.header}
+          cols={config.cols}
```
binh-dam-ibigroup, line 146: Sort props

packages/trip-details/src/fare-detail.tsx
```diff
@@ -0,0 +1,154 @@
...
+        {cols.map(col => {
+          const fare = fareTotals[col.key];
+          return (
+            <td key={col.key}>
+              <FormattedNumber
+                value={(fare?.cents || 0) / 100}
+                // This isn't a "real" style prop
+                // eslint-disable-next-line react/style-prop-object
+                style="currency"
+                currency={fare?.currency?.currencyCode}
```
binh-dam-ibigroup, line 107: Sort props

packages/trip-details/src/fare-detail.tsx
```diff
@@ -0,0 +1,154 @@
...
+          {cols.map(col => {
+            const fare = leg.fares[col.key];
+            return (
+              <td key={col.key}>
+                <FormattedNumber
+                  value={fare?.price?.cents / 100 || 0}
+                  // This isn't a "real" style prop
+                  // eslint-disable-next-line react/style-prop-object
+                  style="currency"
+                  currency={
```
binh-dam-ibigroup, line 84: Sort props

packages/trip-details/src/fare-detail.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+const TotalRow = styled.tr`
+  border-top: 2px solid #333;
+  font-weight: 600;
+`;
+
+const Table = styled.table`
+  border-collapse: collapse;
+  display: block;
+  font-family: sans-serif;
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Could you try and see how the font size looks like with the default font in OTP-RR? (With the Hind font used in OTP-RR, you might need to set a 90% font size or so in the details)

packages/trip-details/src/fare-detail.tsx
```diff
@@ -0,0 +1,154 @@
...
+  td {
+    text-align: right;
+  }
+  td:nth-of-type(2n + 1) {
+    background: #cccccc22;
+  }
+  td.no-zebra {
+    background: transparent;
+  }
+  th:first-of-type {
```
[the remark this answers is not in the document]
binh-dam-ibigroup: I think I see the intent for the height... Do you think using tabs for different customer types (adult, youth, senior) would help? (not required)

packages/trip-details/src/fare-detail.tsx
```diff
@@ -0,0 +1,154 @@
...
+                    legs?.[0].fares?.[col.key]?.price?.currency.currencyCode
+                  }
+                  currencyDisplay="narrowSymbol"
+                />
+              </td>
+            );
+          })}
+        </tr>
+      ))}
+      <TotalRow>
```
binh-dam-ibigroup, line 96: What do you think of displaying the totals to in the header, underneath the payment mode, instead of having a totals row?

packages/trip-details/src/fare-detail.tsx
```diff
@@ -0,0 +1,154 @@
...
+      </TableHeader>
+      {legs.map((leg, index) => (
+        <tr key={index}>
+          <td className="no-zebra">{leg.routeShortName}</td>
+          {cols.map(col => {
+            const fare = leg.fares[col.key];
+            return (
+              <td key={col.key}>
+                <FormattedNumber
+                  value={fare?.price?.cents / 100 || 0}
```
binh-dam-ibigroup, line 80: What do you think about displaying "free" or something that indicates it is part of a free transfer?

packages/trip-details/src/fare-detail.tsx
binh-dam-ibigroup: You could use existing method `renderFare` to make this more concise.

packages/trip-details/src/TripDetails.story.tsx
```diff
@@ -271,3 +303,12 @@ export const TncTransitItineraryWithCustomMessagesAndDetails = makeStory(
 export const FlexItinerary = makeStory({
   itinerary: fareComponentsItinerary
 });
+
+const exampleItinerary = walkInterlinedTransitItinerary;
+export const FareDetailsComponent = () => (
+  <FareDetails
+    transitFares={exampleItinerary.fare}
```
binh-dam-ibigroup, line 310: Use `englishFareKeyMap` and `frenchFareKeyMap` to demo localization (see line 165 in this file).

packages/trip-details/src/fare-detail.tsx
```diff
@@ -0,0 +1,154 @@
...
+import React from "react";
+import styled from "styled-components";
+import { FormattedNumber } from "react-intl";
+import { FareDetailsProps } from "./types";
+
+interface FareTypeTableProps {
+  header: string;
+  cols: {
+    key: string;
+    header: string;
```
binh-dam-ibigroup, line 13: What is the mechanism you have in mind to localize these headers? The keys are slightly different than what is defined in `TripDetails.fareKeyNameMap` prop, so the existing prop cannot be reused here.

binh-dam-ibigroup (changes requested):
Needs a breaking change commit and a clarification on how the fare keys should be localized. While we are with breaking changes, I think we should get rid of the expanding feature on the "Transit fare" line and just show one static summary there, and add an expanding feature to the headers in `FareDetails`, but I'm open to discuss.

packages/trip-details/src/fare-detail.tsx
miles-grant-ibigroup: I think this could look good what do you think @daniel-heppner-ibigroup

packages/trip-details/src/fare-detail.tsx
miles-grant-ibigroup: I think this is a good idea what do you think @daniel-heppner-ibigroup

packages/trip-details/src/fare-detail.tsx
miles-grant-ibigroup: I think the best solution is to try to work on using `fareKeyNameMap` if possible -- @daniel-heppner-ibigroup do you have any thoughts on this or do you want to try to discuss offline

packages/trip-details/src/index.tsx
miles-grant-ibigroup: This could be cool, but I think the table is formatted to spec based on an outside request

binh-dam-ibigroup: Basically, there is duplication of information, so to me it makes sense to at least remove the exiting fare list ((keep the default fare total) in favor of the new fare table. If we wanted we could have a new condensed state in the fare table that would look like:

### Transit Fare: $5.50

| Adult | Cash<br>$5.50 | Orca<br>$3.25 | Orca Lift<br>$1.50 |
|-------|-------|-------|--------|
| Youth | Cash<br>$3.00 | Orca<br>$1.50 |       |
| Senior | Cash<br>$2.00 | Orca<br>$1.00 |       |

[Expand]

miles-grant-ibigroup: I see what you mean but I'm still in favor of both as they provide different utility. However, I'll defer to @daniel-heppner-ibigroup as the designer of this! I see your argument as well

packages/trip-details/src/fare-detail.tsx
daniel-heppner-ibigroup: Agree, I like this idea. It will need to be internationalized however.

daniel-heppner-ibigroup: We aren't actually using the `isTransfer` information that we get from OTP, so that would make a lot of sense to use. That would let us differentiate between a free service and a free transfer. @miles-grant-ibigroup

packages/trip-details/src/index.tsx
daniel-heppner-ibigroup: This is a tough one! I like the idea of reducing duplication, but I find the big table to be a bit confusing. I think when you're looking for your fare type it's easier to just look at a list rather than find two axes on a table. 

As Miles said, the other problem here is that the layout of the fare by leg information is very specific based on an outside request.

packages/trip-details/src/fare-detail.tsx
daniel-heppner-ibigroup: `fareKeyNameMap` is not a localization object, unless I'm missing something? It's just a config object with the names, but there's no way of providing it on a per locale basis. As of now I don't believe we support localizing fare type names (beyond agency-specific names). For things like "ORCA" this is fine, but "Cash" and "Youth" should be localized.

I think the solution here is to add a new section to the localization files for fare names and then move all this (including the `fareKeyNameMap`) into there, so it can be localized.

packages/trip-details/src/fare-detail.tsx
daniel-heppner-ibigroup: How about using an icon like `@styled-icons/boxicons-regular/Transfer` next to fares that use a transfer?

packages/trip-details/src/fare-detail.tsx
daniel-heppner-ibigroup: Yeah, I like this. I think we will need to either make it configurable or ask Jon to see if that would be an acceptable deviation.

packages/trip-details/src/fare-detail.tsx
miles-grant-ibigroup: I love that! Great idea

packages/trip-details/src/fare-detail.tsx
miles-grant-ibigroup: Could we create a map between `fareKeyNameMap` and language strings? So instead of putting the name in the map itself we instead use the key as a i18n key. If a result is returned, we override the name in the object. That would let us support localization here while also remaining backwards compatible

daniel-heppner-ibigroup: So basically use the same key from `fareKeyNameMap` with i18n, and use the old `fareKeyNameMap` as a fallback? That sounds great to me. Want me to implement it?

miles-grant-ibigroup: I can do it next week but if you have time before then that would be great!

miles-grant-ibigroup:
I like the idea of replacing the old listing with the new table if available -- 2 small things:

1) The case of `http://localhost:5555/?path=/story/tripdetails--walk-interlined-transit-itinerary` is not a bit awkward (with the table showing up twice)
2) I think the price shouldn't be bolded -- there is no visual hierarchy!
3) If the fare is 0, the `$` goes away! You see "Cash 0" 

Other than those small things this looks excellent thanks so much for refactoring all that!

daniel-heppner-ibigroup:
Oh yeah, actually the only reason I had it showing up twice in that story from your point (1) was to show what it looks like in each!

packages/trip-details/src/index.tsx
```diff
@@ -149,9 +150,9 @@ export function TripDetails({
   className = "",
   DepartureDetails = null,
   defaultFareKey = "regular",
-  FareDetails = null,
   fareKeyNameMap = {},
-  itinerary
+  itinerary,
+  fareDetailsLayout
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: Prop no longer removed.

packages/trip-details/src/fare-detail.tsx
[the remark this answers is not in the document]
daniel-heppner-ibigroup: Design is pretty set in stone by external request.

packages/trip-details/src/fare-detail.tsx
```diff
@@ -0,0 +1,154 @@
...
+          prev[key] = fareForKey;
+        }
+        return prev;
+      }, {});
+      return {
+        ...leg,
+        fares
+      };
+    })
+    .filter(leg => leg.transitLeg);
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: Can't, since we rely on the indexes in the map and filtering items out would mess up the indexes.

packages/types/src/index.ts
```diff
@@ -345,13 +354,7 @@ type ApplicableId = string | FeedScopedId;
...
-    [name: string]: {
-      fareId: ApplicableId;
-      price: Money;
-      routes: ApplicableId[];
-    }[];
-  };
+  details?: FareDetails;
   fare?: {
     [name: string]: Money;
   };
```
binh-dam-ibigroup, line 360: By the same token:
```suggestion
  fare?: Record<string, Money>;
```

packages/trip-details/src/types.ts
```diff
@@ -50,9 +67,13 @@ export interface TripDetailsProps {
    */
   DepartureDetails?: React.ElementType<DepartureDetailsProps>;
   /**
-   * Slot for a custom component to render the expandable section for fares.
-   */
+  * Slot for a custom component to render the expandable section for fares.
+  */
```
binh-dam-ibigroup, line 71: Fix indent.

binh-dam-ibigroup (approved):
A bit of formatting and type polishing comments, so I give a conditional approval subject to addressing those.

packages/trip-details/src/TripDetails.story.tsx
```diff
@@ -48,6 +54,52 @@ const StyledTripDetails = styled(TripDetails)`
   }
 `;
 
+const fareByLegLayout: FareTableLayout[] = [
+  {
+    header: "regular" as FareTableText,
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: Thank you! I wasn't sure if this would be considered too verbose when something much simpler would do, but I thought it was worth doing to make the fixed set of values more obvious and easier to maintain in the future.

packages/trip-details/src/fare-table.tsx
```diff
@@ -0,0 +1,219 @@
...
+          prev[key] = fareForKey;
+        }
+        return prev;
+      }, {});
+      return {
+        ...leg,
+        fares
+      };
+    })
+    .filter(leg => leg.transitLeg);
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: Cannot do this, as it will change the index of the legs, which we rely on to match legs up with their fares.

packages/trip-details/src/utils.tsx
```diff
@@ -0,0 +1,30 @@
...
+  return <strong>{contents}</strong>;
+}
+
+/**
+ * Render formatted fare.
+ * @param currencyCode The ISO currency code to use (USD, GBP, EUR).
+ * @param fare The fare value, in currency units, to be shown.
+ * @returns The formatted fare value according to the selected locale.
+ */
+export function renderFare(currencyCode: string, fare: number): ReactElement {
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: It's used in many places where a simple `number` is expected, such as in the value prop to `FormattedMessage`.

packages/trip-details/src/fare-table.tsx
```diff
@@ -0,0 +1,219 @@
...
+
+const TableHeader = styled.thead`
+  th {
+    min-width: 5ch;
+    padding: 0.75em 1.5em;
+    text-align: center;
+    font-weight: normal;
+  }
+  th:nth-of-type(2n + 1) {
+    background: #cccccc22;
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: Going to leave the transparency questions to @miles-grant-ibigroup since he did that, but I'm guessing the idea is to create a darker background even on gray backgrounds (which you can see in the trip-details stories).

packages/trip-details/src/types.ts
```diff
@@ -13,14 +13,31 @@ export interface DepartureDetailsProps {
...
+    header: FareTableText;
+    key: string;
+  }[];
+  header: FareTableText;
+}
 export interface TransitFareData {
   [key: string]: Money
 }
 
 export interface FareDetailsProps {
```
daniel-heppner-ibigroup, line 36: Since we're now displaying the fare by leg table in the dropdown rather than under the question mark, we could make a new type and revert this one back to original so that we don't need a breaking change (by the way, is there any easy way to see if we ever committed a breaking change?). Should we keep max/min TNC fare info available for the fare details or is that outdated now anyway?

binh-dam-ibigroup: Yes, you should revert the breaking change. Leave the min/max TNC fare (it's no relevant to this PR).

PR #392, opened 2022-04-29
feat(transit-vehicle-overlay): Add i18n support.

binh-dam-ibigroup:
This PR adds i18n support to the `TransitVehicleOverlay` component for those implementations that use the default markers, popups, and tooltip components from this package. The PR does not include TypeScript-ing because of issues with `MapLayer` types.

Note: It is not required by IBI's implementation of OTP-react-redux, which already passes localized content to the overlay.


packages/transit-vehicle-overlay/src/components/popups/VehiclePopup/index.js
```diff
@@ -1,77 +1,184 @@
...
-  vehicle: coreUtils.types.transitVehicleType,
-
   /** indicate if this vehicle is being tracked, */
   isTracked: PropTypes.bool,
 
   /** callback which forwards the vehicle and tracking status from track button */
-  setTracked: PropTypes.func
+  setTracked: PropTypes.func,
+
+  /** vehicle record - @see: core-utils/types/transitVehicleType */
```
philip-cline, line 174:
```suggestion
  /** vehicle record - @see: core-utils/src/types/transitVehicleType */
```
?

philip-cline (approved):
One tiny nitpick but that's it

philip-cline (approved):
Still got my approval!

miles-grant-ibigroup (approved):
This one looks like it was tricky. Thanks for the changes!

PR #393, opened 2022-04-29
fix(core-utils/query-params): don’t allow optimizing transfers

miles-grant-ibigroup:
`optimize=TRANSFERS` seems to cause problems even on OTP1. This PR removes the dropdown from OTP1 instances, as it was already removed from OTP2 instances.


miles-grant-ibigroup:
Good call binh! Some cruft removed

miles-grant-ibigroup:
Good call binh! Some cruft removed

packages/trip-form/i18n/en-US.yml
```diff
@@ -34,7 +34,6 @@ otpUi:
     optimizeBikeSpeed: Speed
     optimizeFor: Optimize for
     optimizeQuick: Speed
```
binh-dam-ibigroup, line 36: These two strings should be removed too, they are for the "Optimize Walk" selector.

binh-dam-ibigroup: Correction, only `optimizeQuick` should be removed here.

binh-dam-ibigroup (approved):
See https://github.com/opentripplanner/otp-ui/pull/393#discussion_r864112804 and it should be good to go.

philip-cline (approved):

PR #398, opened 2022-05-05
Deprecate TimeOptions

binh-dam-ibigroup:
No breaking changes in this PR that just removes superficial uses of `TimeOptions` and related code in the `itinerary-body` demo code.

### DIFF
--- a/packages/core-utils/src/deprecated.js
+++ b/packages/core-utils/src/deprecated.js
@@ -318,6 +318,8 @@ export function summarizeQuery(query, locations = []) {
 }
 
 export function getTimeZoneOffset(itinerary) {
+  logDeprecationWarning("getTimeZoneOffset");
+
   if (!itinerary.legs || !itinerary.legs.length) return 0;
 
   // Determine if there is a DST offset between now and the itinerary start date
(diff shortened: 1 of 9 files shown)


packages/core-utils/src/deprecated.js
```diff
@@ -316,3 +316,17 @@ export function summarizeQuery(query, locations = []) {
     : require("./itinerary").toSentenceCase(query.mode);
   return `${mode} from ${from} to ${to}`;
 }
+
+export function getTimeZoneOffset(itinerary) {
```
miles-grant-ibigroup, line 320: We should probably add a deprecation warning

```suggestion
export function getTimeZoneOffset(itinerary) {
logDeprecationWarning("getTimeZoneOffset");
```

miles-grant-ibigroup (approved):
Works great thanks for cleaning that one up! Just one small addition to fully deprecate the method

packages/core-utils/src/deprecated.js
binh-dam-ibigroup: Yes, good one, thank you (https://github.com/opentripplanner/otp-ui/pull/398/commits/57b99e9ca4c62ccf622b5d38a14be4d6b6754fc6).

daniel-heppner-ibigroup (approved):
LGTM

PR #401, opened 2022-05-11
fix(itinerary-body): grab mapillary URLs instead of just ID

miles-grant-ibigroup:
Using the ID was nice to be able to render the iframe with, but the Mapillary iframe does not work with chrome. 

This PR adjusts the Mapillary fetch to return an image URL instead.


daniel-heppner-ibigroup (approved):
:/

PR #402, opened 2022-05-17
fix(trip-form/i18n): Add missing French messages

binh-dam-ibigroup:
Adds missing French messages for walk reluctance.


miles-grant-ibigroup (approved):
Thanks for this!

PR #403, opened 2022-05-17
Add i18n support for default vehicle/station names

binh-dam-ibigroup:
This PR adds i18n support for the vehicle/stations names shown in the popups.

PR #367, opened 2022-04-01
Add Typescript to `core-utils`

miles-grant-ibigroup:
The big one...

This PR removes many `PropType`s throughout the packages. These should be replaced by their typescript alternatives

Removes all references to core-utils types. This make some of the propTypes in non typescripted packages a little more useless. This should be resolved by type-scripting them. 

Some stories are lost pending a new release of some of these packages. 

### Packages depending on updated `from-to-picker`
- park-and-ride-overlay
- stops-overlay
- vehicle-rental-overlay

I think my strategy here will be to split out the `core-utils` and `from-to-picker` parts of this PR into a new one, and then have that PR block this one.

PR #404, opened 2022-05-20
chore(trip-details:deps): Remove types pkg dep

binh-dam-ibigroup:
Dev-time references to other packages in the same mono repo are not necessary. This fixes the build that's reported broken after merging #403 (it probably became broken after merging #367).

### DIFF
--- a/packages/trip-details/package.json
+++ b/packages/trip-details/package.json
@@ -17,6 +17,7 @@
     "velocity-react": "^1.4.3"
   },
   "devDependencies": {
+    "@opentripplanner/types": "^2.0.0",
     "@types/flat": "^5.0.2"
   },
   "peerDependencies": {
--- a/packages/trip-details/package.json
+++ b/packages/trip-details/package.json
@@ -17,7 +17,6 @@
     "velocity-react": "^1.4.3"
   },
   "devDependencies": {
-    "@opentripplanner/types": "^1.1.0",
     "@types/flat": "^5.0.2"
   },
   "peerDependencies": {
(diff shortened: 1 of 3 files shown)


miles-grant-ibigroup (changes requested):
I was able to fix this like so: https://github.com/opentripplanner/otp-ui/pull/400/commits/3a177cba3a6f18d9eede322dd7f93440281c9519#diff-0d962559c234d86f1bdbc98f27ab0f840d7f64f1813d58835fd0443d1df3ae57L20

Does removing the dependency really not affect the build?

binh-dam-ibigroup:
> I was able to fix this like so: [3a177cb#diff-0d962559c234d86f1bdbc98f27ab0f840d7f64f1813d58835fd0443d1df3ae57L20](https://github.com/opentripplanner/otp-ui/commit/3a177cba3a6f18d9eede322dd7f93440281c9519#diff-0d962559c234d86f1bdbc98f27ab0f840d7f64f1813d58835fd0443d1df3ae57L20)
> 
> Does removing the dependency really not affect the build?

Maybe, but the same thing will happen at the next breaking change.

miles-grant-ibigroup (approved):
If this doesn't affect the build or the snapshots then I guess I approve!

miles-grant-ibigroup (changes requested):
I think upgrading the package is the way to go

binh-dam-ibigroup:
> 30d2845

Lol I readded the ref in https://github.com/opentripplanner/otp-ui/pull/404/commits/30d284545af5aeac8a125bc5232c4501459dcbac and that is not working!

miles-grant-ibigroup:
I thiiiink you might just have to update the snapshots! I had the same problem when I patched this in all my PRs but I'm not sure why the types package changes the snapshots I will be honest

miles-grant-ibigroup (approved):

daniel-heppner-ibigroup (approved):

PR #406, opened 2022-05-26
OTP2 Scooter Fixes new test itineraries across all packages

miles-grant-ibigroup:
Adds the new scooter test itinerary to all packages that use the test itineraries. Relies on https://github.com/opentripplanner/otp-ui/pull/400


packages/core-utils/src/map.ts
```diff
@@ -133,7 +133,7 @@ export function itineraryToTransitive(
           // OTP2 Scooter case
           leg.mode === "SCOOTER"
         ) {
-          fromPlaceId = `escooter_rent_station_${leg.from.bikeShareId}`;
+          fromPlaceId = `escooter_rent_station_${leg.from.name}`;
```
binh-dam-ibigroup, line 136: This change contradicts https://github.com/opentripplanner/otp-ui/pull/400/files#r882102376.

packages/trip-details/src/TripDetails.story.tsx
```diff
@@ -236,6 +237,10 @@ export const EScooterRentalTransitItinerary = makeStory({
   itinerary: eScooterRentalTransiteScooterRentalItinerary
 });
 
+export const OTP2EScooterRentalTransitItinerary = makeStory({
+  itinerary: otp2ScooterItinerary
+});
+
```
binh-dam-ibigroup, line 243: What is this story supposed to show?

binh-dam-ibigroup (commented):
Could you advise regarding my two comments, please?

packages/core-utils/src/map.ts
miles-grant-ibigroup: I forgot to add a comment here, which I will do. Essentially the reason that I moved back to `name` is that the fromPlace is used to generate the title of the e-scooter on the map. While `bikeShareId` is always a long yucky uuid, `name` has the potential to be a nicer title. 

The e-scooter naming is a little bugged right now, so this will likely be changed again in the future.

packages/trip-details/src/TripDetails.story.tsx
miles-grant-ibigroup: This is here for completeness, I guess the main thing I was checking for is that it didn't crash while parsing the slightly different OTP2 itinerary format. If you don't think that's enough of a reason to have a story, we can remove it!

binh-dam-ibigroup: Then add a comment saying what you stated.

packages/core-utils/src/map.ts
binh-dam-ibigroup: Ok, but there was another instance around line 158 so both need to be consistent.

miles-grant-ibigroup: I will undo the change for now and make a separate full e-scooter name fix PR

miles-grant-ibigroup: Undone in 2f4508cf91057c01c2bab6f0ec75666b49d60ecd

packages/trip-details/src/TripDetails.story.tsx
miles-grant-ibigroup: Comment added in 2f4508cf91057c01c2bab6f0ec75666b49d60ecd

binh-dam-ibigroup (approved):

philip-cline (approved):

PR #407, opened 2022-05-26
fix(location-field): correctly debounce autocomplete requests

miles-grant-ibigroup:
The debounce was using the `throttle` method which did the wrong thing! This PR corrects the behavior and results in actual debouncing.


philip-cline (approved):

packages/location-field/src/index.tsx
```diff
@@ -108,7 +108,7 @@ const LocationField = ({
     }
   }, [initialSearchResults]);
 
-  const geocodeAutocomplete = throttle(1000, text => {
+  const geocodeAutocomplete = debounce(300, text => {
```
binh-dam-ibigroup, line 111: Can you try wrapping the call to `debounce` with `useCallback`, please?
Otherwise, this function is being re-created and the debounce reset at each render (each time the user types a character into the search box), making debounce useless.

binh-dam-ibigroup (changes requested):

packages/location-field/src/index.tsx
miles-grant-ibigroup: Good catch thank you! Had to use `useMemo` instead but the debounce is now working in d37da4df0da1f791f31043c2a955ca0a1f4cb3a5

binh-dam-ibigroup: `useCallback` includes a `useMemo` thing, but as you have written  it is fine.

binh-dam-ibigroup (approved):

PR #409, opened 2022-06-07
Fix scripts package

binh-dam-ibigroup:
This PR is an attempt to fix the semantic release failure from merging #287 by adding the missing index file to the `scripts` package (that requires separating code that runs right away from the reusable code).


miles-grant-ibigroup (approved):
Didn't know this was needed! In the future we might want to make `index.ts` the entrypoint, but given how simple things are at the moment this looks good to me.

philip-cline (approved):

binh-dam-ibigroup:
Looks like I messed up and forgot a `fix` commit. Merging this PR did not trigger a package release. (But somehow, packages were released previously even if semantic release failed...)

PR #410, opened 2022-06-08
fix(packages/scripts): Add fix commit to release changes from #409.

binh-dam-ibigroup:
This PR adds that `fix` commit I forgot in #409!


miles-grant-ibigroup (approved):
Thanks for `fix`ing this!

philip-cline (approved):

PR #411, opened 2022-06-09
Fix vehicle rental UUID

daniel-heppner-ibigroup:
To fix the vehicle rental UUID, we needed to move the `itineraryToTransitive` code out of core-utils. 

The function `itineraryToTransitive` relies on the `getPlaceName` function, and we needed to switch to the one in the `itinerary-body` package rather than the old deprecated one in `core-utils`. Only problem is that `core-utils` can't depend on other packages without creating circular dependencies, which is why `itineraryToTransitive` needed to move. Besides that, I think it makes more sense in `transitive-overlay` as a utility relating to that.


packages/itinerary-body/src/util.ts
```diff
@@ -102,7 +102,10 @@ export function getPlaceName(
 ): string {
   // If address is provided (i.e. for carshare station, use it)
   if (place.address) return place.address.split(",")[0];
-  if (place.networks && place.vertexType === "VEHICLERENTAL") {
+  if (
+    place.networks &&
+    (place.vertexType === "VEHICLERENTAL" || place.vertexType === "BIKESHARE")
```
miles-grant-ibigroup, line 107: Is this the entirety of the patch beyond moving that method into transitive-overlay?

miles-grant-ibigroup (approved):
A bit worried about any regressive side effects, but all looks good in these stories and in my testing. I guess the otp-rr PR will contain the bulk of the work here!

daniel-heppner-ibigroup:
Please take another look @miles-grant-ibigroup, I made some changes to display "E-Scooter" wherever I could. It's a bit hacky because of how OTP2 returns the data.

packages/itinerary-body/src/ItineraryBody/place-row.tsx
```diff
@@ -54,7 +54,13 @@ export default function PlaceRow({
   // check the real-time arrival information for the next leg of their journey.
   const interline = !!(!isDestination && leg.interlineWithPreviousLeg);
   const hideBorder = interline || !legIndex;
-  const place = isDestination ? leg.to : leg.from;
+  const place = isDestination ? { ...leg.to } : { ...leg.from };
```
miles-grant-ibigroup, line 57: Why was this change made?

packages/itinerary-body/src/util.ts
```diff
@@ -102,11 +102,10 @@ export function getPlaceName(
 ): string {
   // If address is provided (i.e. for carshare station, use it)
   if (place.address) return place.address.split(",")[0];
-  if (place.networks && place.vertexType === "VEHICLERENTAL") {
-    // For vehicle rental pick up, do not use the place name. Rather, use
-    // company name + vehicle type (e.g., SPIN E-scooter). Place name is often just
-    // a UUID that has no relevance to the actual vehicle. For bikeshare, however,
-    // there are often hubs or bikes that have relevant names to the user.
+  if ((place.name.match(/-/g) || []).length > 3) {
```
miles-grant-ibigroup, line 105: Was it not possible to re-use the version of this match used elsewhere?

miles-grant-ibigroup (changes requested):
Thanks for figuring out this fix! From what I'm seeing this isn't as bad a hack as you think it is -- just a few small style nitpicks, but overall this works great!

packages/itinerary-body/src/util.ts
daniel-heppner-ibigroup: I really tried to use the other function that already does this, but that one is really oriented to the station overlay and not an itinerary. It would have needed a lot of changes to make it generic enough to use both places and probably would have been a lot more code overall.

packages/itinerary-body/src/ItineraryBody/place-row.tsx
daniel-heppner-ibigroup: We need a copy of the place so that when we edit the vertex type on the next line it doesn't cause any side effects.

miles-grant-ibigroup (approved):
Thanks for the explanations!

daniel-heppner-ibigroup:
The snapshots needs i18n

packages/transitive-overlay/src/TransitiveOverlay.story.js
```diff
@@ -333,13 +336,13 @@ export const FlexItinerary = () => (
...
     <TransitiveOverlay
-      transitiveData={itineraryToTransitive(flexItinerary, companies)}
+      transitiveData={itineraryToTransitive(flexItinerary, { companies })}
       visible
     />
   </BaseMap>
 );
 
-export const OTP2ScooterItinerary = () => (
+export const OTP2ScooterItinerary = injectIntl(() => (
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: Yeah, it was added so we could use `useIntl()` below.

packages/core-utils/src/map.ts
```diff
@@ -78,255 +64,6 @@ export function matchLatLon(location1: Location, location2: Location): boolean {
...
-  tdata.stops.push(...Object.values(stops));
-
-  // add the journey to the tdata journeys array
-  tdata.journeys.push(journey);
-
-  // console.log('derived tdata', tdata);
-  return tdata;
-}
-
 type TransitivePlaceRaw = {
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: Looks like it's still used in the functions below

packages/transitive-overlay/src/TransitiveOverlay.story.js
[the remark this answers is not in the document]
binh-dam-ibigroup: > Yeah, it was added so we could use `useIntl()` below.

For react-intl, you either use `useIntl` or `injectIntl` but not both in the same component.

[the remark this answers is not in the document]
daniel-heppner-ibigroup: After some back and forth, finally figured out a solution here! 
3ca5cc7

packages/transitive-overlay/src/util.ts
```diff
@@ -0,0 +1,276 @@
+import {
+  Company,
+  Itinerary,
+  Leg,
+  TransitiveData
+} from "@opentripplanner/types";
+import coreUtils from "@opentripplanner/core-utils";
+import { getPlaceName } from "@opentripplanner/itinerary-body/lib/util";
```
binh-dam-ibigroup, line 8: This is a new dependency to the itinerary-body package. Please move that dependency in the `package.json` for the `transitive-overlay` package from a dev to a full dependency.

binh-dam-ibigroup (changes requested):
See https://github.com/opentripplanner/otp-ui/pull/411/files#r901989177

packages/transitive-overlay/src/util.ts
binh-dam-ibigroup, line 8:
```suggestion
import { getPlaceName } from "@opentripplanner/itinerary-body";
```

daniel-heppner-ibigroup:
@binh-dam-ibigroup ready for another review here I believe!

packages/transitive-overlay/package.json
```diff
@@ -21,7 +21,7 @@
     "@opentripplanner/core-utils": "^5.0.1",
     "lodash.isequal": "^4.5.0",
     "transitive-js": "^0.14.1",
-    "@opentripplanner/itinerary-body": "^3.0.5"
+    "@opentripplanner/itinerary-body": "^3.1.0"
```
binh-dam-ibigroup, line 24: Sort dependencies.

binh-dam-ibigroup (changes requested):
Oops these stories broke under the transitive-overlay package:
- E Scooter Rental Itinerary
- E Scooter Rental Transit itinerary

Also, do these changes cover the OTP 2 Scooter Itinerary? (The location is showing a long uuid)

daniel-heppner-ibigroup:
OK, I think the dependencies are correct on this now. However I do see that the Portland scooter station shows as "EMAQ". I believe that is the name of the station and is printed on the station somewhere, in which case I think it makes sense to show it to the user. If they don't have it displayed anywhere in the physical station then maybe it would be worth hiding it in OTP, but I'm not sure how to approach that. Perhaps a regex that looks for 4 uppercase letters?

binh-dam-ibigroup:
> Also, do these changes cover the OTP 2 Scooter Itinerary? (The location is showing a long uuid)

This story http://localhost:5555/?path=/story/transitiveoverlay--otp-2-scooter-itinerary&globals=locale:en-US is still showing a long uuid. Is it covered with this PR?

daniel-heppner-ibigroup:
That one shows up correctly 

> > Also, do these changes cover the OTP 2 Scooter Itinerary? (The location is showing a long uuid)
> 
> This story http://localhost:5555/?path=/story/transitiveoverlay--otp-2-scooter-itinerary&globals=locale:en-US is still showing a long uuid. Is it covered with this PR?

It should be. It shows up correctly for me. Have you tried deleting your node_modules folders (including in packages/transitive-overlay)?
 
image

binh-dam-ibigroup:
> > > Also, do these changes cover the OTP 2 Scooter Itinerary? (The location is showing a long uuid)
> > 
> > 
> > This story http://localhost:5555/?path=/story/transitiveoverlay--otp-2-scooter-itinerary&globals=locale:en-US is still showing a long uuid. Is it covered with this PR?
> 
> It should be. It shows up correctly for me. Have you tried deleting your node_modules folders (including in packages/transitive-overlay)?
> 
> image

Ok thanks for clarifying.

binh-dam-ibigroup (approved):
Thanks for addressing my comments.

PR #413, opened 2022-06-10
fix(DateTimeSelector): Add and use timeZone prop to format current time

binh-dam-ibigroup:
This PR adds a `timeZone` prop to the `DateTimeSelector` component, so that when the user clicks "Leave now", the current time passed to the OTP query parameters is the local time at the specified timeZone (which is not necessarily the user time zone). This is useful when querying OTP for agencies in a different time zone than the user time zone.


miles-grant-ibigroup:
Oh, excellent! I misunderstood the description then. Thank you for clearing that up!

miles-grant-ibigroup:
Oh, excellent! I misunderstood the description then. Thank you for clearing that up!

miles-grant-ibigroup:
Oh, excellent! I misunderstood the description then. Thank you for clearing that up!

philip-cline (approved):

PR #416, opened 2022-06-23
Use booking rule messages from GTFS

miles-grant-ibigroup:
GTFS booking rule messages weren't being used in all parts of the UI where it makes sense to use them. This PR makes better use of booking rule messages, and adds a new demo flex itinerary that contains messages.


daniel-heppner-ibigroup (approved):
Code looks good. Does this mean that "extra" flex messages won't be translatable?

miles-grant-ibigroup:
Translations within GTFS strings have been an open issue for a while... I believe there's some proposals but nothing we've implemented yet.

packages/itinerary-body/src/util.ts
```diff
@@ -128,3 +129,22 @@ export function getPlaceName(
...
+export function getFlexMessageValues(
+  info: FlexBookingInfo
+): {
+  hasLeadTime: boolean;
+  hasPhone: boolean;
+  leadDays?: number;
+  phoneNumber?: string;
+} {
+  return {
+    hasLeadTime: coreUtils.itinerary.isAdvanceBookingRequired(info),
```
binh-dam-ibigroup, line 145: Potential improvement (to reduce the number of params in the language strings): could we remove `hasLeadTime` and use whether `leadDays` is non-zero instead? Same for `hasPhone` and `phoneNumber`.

packages/trip-details/i18n/en-US.yml
```diff
@@ -20,28 +20,11 @@ otpUi:
...
-      }{hasLeadTime, select,
-        true { at least {leadDays, plural, one {# day} other {# days}} in advance}
-        other {}
-      }.
     title: Trip Details
     tncFare: "{companies} Fare: <strong>{minTNCFare} - {maxTNCFare}</strong>"
     transitFare: Transit Fare
     transitFareEntry: "{name}: <strong>{value}</strong>"
-    tripIncludesFlex: This trip includes flexible routes.
+    tripIncludesFlex: This trip includes flexible routes. {extraMessage}
```
binh-dam-ibigroup, line 27: Can we just render the extra message separately? (as far as I know `extraMessage` is not subject to i18n.)

packages/trip-details/src/index.tsx
```diff
@@ -325,58 +309,20 @@ export function TripDetails({
                   }
                   description="Text stating that portions of the trip include a flex (on-demand) transit service."
                   id="otpUi.TripDetails.tripIncludesFlex"
+                  values={{
+                    extraMessage: [
```
binh-dam-ibigroup, line 313: Please render `extraMessage` as its own JSX after `<FormattedMessage>`. The reason is that it is content that is completely independent (although related) of the sentence "This trip includes flex routes."

binh-dam-ibigroup (changes requested):
Looks good overall, but I render the extra pickup/dropoff messages outside of `<FormattedMessage>`.

packages/trip-details/i18n/en-US.yml
miles-grant-ibigroup: The reason I did this is to support RTL languages or languages (such as german) where it would be customary to wrap this sort of message in quotes

packages/trip-details/src/index.tsx
miles-grant-ibigroup: See previous comment about RTL langauges

packages/itinerary-body/src/util.ts
miles-grant-ibigroup: `hasLeadTime` is removed. `hasPhone` is not, as react-intl can to my knowledge not check the "blank-ness" of a string

miles-grant-ibigroup:
Comments are addressed in ce688347ec9b573d85abdbeaa873be8e1fbb9c33

packages/trip-details/i18n/en-US.yml
binh-dam-ibigroup: Could you add a comment in `en-US.yml` regarding the quotes?

binh-dam-ibigroup (approved):
Don't want to be too hung up on the `extraMessages` concatenation, so please add a comment regarding the language-specific decoration.

miles-grant-ibigroup:
Comment is added in [da8b4c3](https://github.com/opentripplanner/otp-ui/pull/416/commits/da8b4c32baccb722c29e2f66b67ede8b8e2d5aa1)

PR #418, opened 2022-06-29
fix(itinerary-body): detect UUID in place name, fix scooter vertextype

daniel-heppner-ibigroup:
Sorry for the double PR, but I hadn't realized that I had a more updated version of this package in my node-modules, so I hadn't realized the need to push more updates to it.

This is all the changes necessary for itinerary-body package. It detects the UUID based place names, and also rewrites the vertextype on places to improve compatibility between OTP1 and OTP2. It's a bit ugly but that's just because OTP2 reports everything as BIKESHARE even when it's not BIKESHARE, and the only way of knowing is looking ahead to the next leg. When we don't have access to the next leg (say deep in a component), we will just say "Vehicle" so it isn't wrong.


packages/itinerary-body/src/defaults/access-leg-description.tsx
```diff
@@ -22,6 +22,13 @@ export default function AccessLegDescription({
   style
 }: Props): ReactElement {
   const intl = useIntl();
+  // Replace the Vertex Type for BIKESHARE with VEHICLE as we do not know that
```
miles-grant-ibigroup, line 25: This seems to not cause problems with OTP1 but are we sure? Do we know that OTP1 will never return "BIKESHARE"?

miles-grant-ibigroup (approved):
Haven't tested this integrated into otp-rr (can't release an alpha build until maplibre work is finished) but in the stories at least this looks to work really well! Thanks for the changes!

packages/itinerary-body/src/defaults/access-leg-description.tsx
daniel-heppner-ibigroup: OTP1 can return BIKESHARE, and this will replace it with VEHICLE. However that will cause the behavior to be consistent between OTP1 and OTP2. It's true that with OTP1 `BIKESHARE` actually does mean bike share, but we have no way (inside this function) of knowing whether it's OTP1 bikeshare or OTP2 vehicle rental.

PR #421, opened 2022-07-12
fix(trip-details): Revert accidental breaking change

daniel-heppner-ibigroup:
The fare-by-leg PR got merged before I had pushed this commit to it, causing a breaking change that was not marked properly. In this PR I revert the breaking change since it is not necessary.

### DIFF
--- a/packages/trip-details/src/fare-table.tsx
+++ b/packages/trip-details/src/fare-table.tsx
@@ -7,7 +7,7 @@ import { Transfer } from "@styled-icons/boxicons-regular/Transfer";
 
 import { boldText, renderFare } from "./utils";
 
-import { FareDetailsProps, FareTableLayout, FareTableText } from "./types";
+import { FareLegTableProps, FareTableLayout, FareTableText } from "./types";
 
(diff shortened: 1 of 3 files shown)


packages/trip-details/src/index.tsx
```diff
@@ -276,7 +276,13 @@ export function TripDetails({
           <TripDetail
             // Any custom description for the transit fare needs to be handled by the slot.
             description={
-              FareDetails && <FareDetails transitFares={transitFares} />
+              FareDetails && (
+                <FareDetails
+                  transitFares={transitFares}
```
binh-dam-ibigroup, line 281: Sort props.

packages/trip-details/src/fare-table.tsx
```diff
@@ -183,7 +183,7 @@ const FareLegDetails = ({
   transitFareDetails,
   transitFares,
   legs
```
binh-dam-ibigroup, line 185: Sort props

binh-dam-ibigroup (approved):
Sort props before merging.

PR #423, opened 2022-07-19
fix(location-field): remove zip code from previews

miles-grant-ibigroup:
The zip codes were distracting more than they were helping, so this PR removes them.


daniel-heppner-ibigroup (approved):
LGTM

philip-cline (approved):
Great to see the correct spelling of neighbourhood being used

PR #426, opened 2022-07-25
Vector tiles PR tweaks

binh-dam-ibigroup:
This PR suggests a way to remove the somewhat "artificial" `forceMaxHeight` and `passedRef` props from the `BaseMap` component from #420, see if these changes are useful to you. (does not cover propagating changes to other packages, snapshots, and other refactors such as having a storybook-wide decorator for the various map overlays).


miles-grant-ibigroup:
Thanks Binh for figuring this out! Took a bit of trickery to get the new approach working in all the overlays but I think the end result is much cleaner. Thanks again!

PR #427, opened 2022-07-28
Remove momentjs (2/2)

### LINKED ISSUE #359: An alert set for 24 hours after a timestamp that is "today" should be reported as "tomorrow"
This may come from how the moment library treats time differences, per https://momentjs.com/docs/#/displaying/difference/ (In our case, it will compute the fractional day difference and truncate that down (e.g., 0.85 days is truncated down to 0 days).

Steps to reproduce:

1. Checkout commit https://github.com/opentripplanner/otp-ui/pull/357/commits/e9a6833e56605b34089c2ff5bab282e900f7bf04
2. Run Storybook and open http://localhost:5555/?path=/story/itinerarybody-otp-react-redux--walk-transit-walk-itinerary&globals=locale:en-US
3. Expand the transit alerts. Observe that the first alert is labeled as "Tomorrow" 
4. Change [this line](https://github.com/opentripplanner/otp-ui/blob/e9a6833e56605b34089c2ff5bab282e900f7bf04/packages/itinerary-body/src/stories/OtpRrItineraryBody.story.js#L41) so that it adds 24 hours to the `now` variable, instead of 36 hours.
5. In the same story, observe that the first transit alert is labeled as "Today", which is incorrect.

binh-dam-ibigroup:
This PR replaces the momentjs dependency with date-fns on the itinerary-body and trip-form packages.
Minor release versions should be issued as a result.

Fix #359 by determining what constitutes "today", "tomorrow" or "yesterday" by comparing calendar days instead of 24-hour periods in the AlertsBody.

### DIFF
--- a/packages/itinerary-body/package.json
+++ b/packages/itinerary-body/package.json
@@ -16,8 +16,8 @@
     "@opentripplanner/location-icon": "^1.4.0",
     "@styled-icons/fa-solid": "^10.34.0",
     "@styled-icons/foundation": "^10.34.0",
+    "date-fns": "^2.28.0",
     "flat": "^5.0.2",
-    "moment": "^2.24.0",
     "react-resize-detector": "^4.2.1",
     "velocity-react": "^1.4.3"
   },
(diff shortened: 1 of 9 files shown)


binh-dam-ibigroup:
Superseded by #428.

PR #429, opened 2022-08-03
Remove velocity-react

daniel-heppner-ibigroup:
Velocity-react is a deprecated library that has been removed from OTP-RR. It is a large animation library that we only used for one thing, so this PR replaces it with a simple library that uses CSS animations instead. 
Removing it from OTP-UI will allow it to be completely eliminated from the bundle.


miles-grant-ibigroup (changes requested):
Looks like you might have to update snapshots -- otherwise seems to be working for me thanks for removing that old library!

binh-dam-ibigroup:
@daniel-heppner-ibigroup Could you please check the test please?

daniel-heppner-ibigroup:
> @daniel-heppner-ibigroup Could you please check the test please?

I've been told it was failing because of one of your PRs needing to be merged? I'll double check.

binh-dam-ibigroup:
> > @daniel-heppner-ibigroup Could you please check the test please?
> 
> I've been told it was failing because of one of your PRs needing to be merged? I'll double check.

That's right. Cherry-pick this commit into this PR: https://github.com/opentripplanner/otp-ui/pull/430/commits/032b6b3b8801e52e9e9da96818ca5c3a912fcbf1.

binh-dam-ibigroup:
@daniel-heppner-ibigroup Could you please update snapshots?

daniel-heppner-ibigroup:
> @daniel-heppner-ibigroup Could you please update snapshots?

Hmm, locally my tests are failing on master as well. I don't think it's a snapshot issue. Can you have a look?

packages/itinerary-body/src/TransitLegBody/index.tsx
```diff
@@ -210,18 +210,13 @@ class TransitLegBody extends Component<Props, State> {
...
-            {expandAlerts && (
-              <AlertsBody
-                alerts={leg.alerts}
-                AlertIcon={AlertBodyIcon}
-                timeZone={timeZone}
-              />
-            )}
-          </VelocityTransitionGroup>
+          <AnimateHeight duration={500} height={expandAlerts ? "auto" : 0}>
+            <AlertsBody
```
binh-dam-ibigroup, line 214: Previously, `<AlertsBody>` was rendered conditionally. If rendering it unconditionally going forward, you'll need to check in that component's render function for props that could be null, such as `alerts` where some of the snapshot tests are failing.

packages/trip-details/src/trip-detail.tsx
```diff
@@ -51,19 +51,14 @@ export default class TripDetail extends Component<Props, State> {
...
-            <S.TripDetailDescription>
-              <S.HideButton onClick={this.onHideClick}>
-                <TimesCircle size="0.92em" />
-              </S.HideButton>
-              {description}
-            </S.TripDetailDescription>
-          )}
-        </VelocityTransitionGroup>
+        <AnimateHeight duration={300} height={expanded ? "auto" : 0}>
+          <S.TripDetailDescription>
```
binh-dam-ibigroup, line 55: Previously, the contents in trip details was rendered conditionally. From my local test log, it looks there is still a reference to momentjs that is lurking... could you try remove it?

binh-dam-ibigroup (changes requested):
Could you fix the itinerary test failures please?

daniel-heppner-ibigroup:
I am not sure why this is failing on CI, it passes just fine locally. It looks like styled-components classnames are slightly different. The ID is the same but on CI it is adding a `-sc` part to the classname.

binh-dam-ibigroup:
> I am not sure why this is failing on CI, it passes just fine locally. It looks like styled-components classnames are slightly different. The ID is the same but on CI it is adding a `-sc` part to the classname.

Yeah, I had to do the same thing with https://github.com/opentripplanner/otp-ui/pull/428/commits/5933a58b7085c060d36ef84af4bb060ad4b05c23. Try deleting the `.jest-cache` folder in your repo folder.

daniel-heppner-ibigroup:
I've never had more trouble getting tests to pass than with this PR...

daniel-heppner-ibigroup:
No cache, branch in sync with origin- I really don't know why they're failing on here and passing locally.

daniel-heppner-ibigroup:
Success!

binh-dam-ibigroup (approved):
LGTM!

packages/itinerary-body/src/TransitLegBody/alerts-body.tsx
```diff
@@ -21,6 +21,7 @@ export default function AlertsBody({
   AlertIcon = S.DefaultAlertBodyIcon,
   timeZone = getUserTimezone()
 }: Props): ReactElement {
+  if (typeof alerts !== "object") return null;
```
miles-grant-ibigroup, line 24: It looks like `alerts` is an array below? If for some reason we need to check for 'object' anyway, I'd add a comment about what's going on

yarn.lock
```diff
@@ -16998,16 +16998,6 @@ react-textarea-autosize@^8.3.0:
...
-  version "2.9.0"
-  resolved "https://registry.yarnpkg.com/react-transition-group/-/react-transition-group-2.9.0.tgz#df9cdb025796211151a436c69a8f3b97b5b07c8d"
-  integrity sha512-+HzNTCHpeQyl4MJ/bdE0u6XRMe9+XG/+aL4mCxVN4DnPBQ0/5bfHWPDuOZUzYdMj94daZaZdCCc1Dzt9R/xSSg==
-  dependencies:
-    dom-helpers "^3.4.0"
-    loose-envify "^1.4.0"
-    prop-types "^15.6.2"
-    react-lifecycles-compat "^3.0.4"
-
 react-transition-group@^4.3.0:
```
miles-grant-ibigroup, line 17001: It looks like `react-transition-group` is still here because of something else. Might be a cyclical thing where it's an older version of an otp-ui package we just need to update, but I'd still do an analysis just to double check! I believe I've used `npm why` in this repo before and it worked at least sort of

miles-grant-ibigroup: There might be a `yarn why` or similar as well

miles-grant-ibigroup (changes requested):
Looking good and working good, just a few small things to double check before we merge!

packages/itinerary-body/src/TransitLegBody/alerts-body.tsx
daniel-heppner-ibigroup: `typeof` on an array is `object`! I found that out when I wrote this

yarn.lock
daniel-heppner-ibigroup: Good catch

daniel-heppner-ibigroup: Actually, React Transition Group is a separate library from React that just helps you do CSS animations by applying and removing CSS classes to your elements. I wouldn't even be surprised if Animate Height package is depending on it. So I think it's fine that it's in here, as long as Velocity is gone.

packages/itinerary-body/src/TransitLegBody/alerts-body.tsx
miles-grant-ibigroup: Javascript is very cool

yarn.lock
miles-grant-ibigroup: Thanks for double checking!

miles-grant-ibigroup (approved):

PR #430, opened 2022-08-03
Update to coreutils 7

binh-dam-ibigroup:
This PR updates packages to use core-utils 7.0.0 and types 3.0.0. Patch releases should occur for all affected packages, except for core-utils that should not have a release

(The types dev dependency for core-utils is updated, but it is identical to the released one, so that does not change `.d.ts` files generated for core-utils.)

### DIFF
--- a/packages/core-utils/package.json
+++ b/packages/core-utils/package.json
@@ -28,6 +28,6 @@
     "tsc": "tsc"
   },
   "devDependencies": {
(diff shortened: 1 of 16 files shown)


miles-grant-ibigroup (approved):
Looks good let's hope this entire process works out...

daniel-heppner-ibigroup (approved):

PR #432, opened 2022-08-08
Micromobility name override

daniel-heppner-ibigroup:
I think this should allow us to add a setting to the config to override all the station names for a given company. It shouldn't even require an OTP-RR change other than package updates. I think we'll probably need to split this into two PRs though, right? Or use the alpha branch?


miles-grant-ibigroup (changes requested):
I've tested this and it works! Thanks for the change. We should be able to avoid having to update the types package as the new version of `itinerary-body` should still build even if the types package is out of date (as in the ci types get imported from the files instead of the package itself... for some reason).

This has my approval once 1) master is merged into here, which I think is causing the ci to fail and 2) we add a fix commit! That way a new version will be created.

daniel-heppner-ibigroup:
Apparently adding 190k lines to the snapshot file fixed it. Good to go boss
image

miles-grant-ibigroup:
I've taken a look at the snapshots and it looks like they're just rearranging all the snapshots? Very strange...

miles-grant-ibigroup (approved):
Just tested with otp-rr and this works great thanks for the simple fix!

philip-cline (approved):

PR #433, opened 2022-08-08
fix(printable-itinerary): Update itinerary-body version

binh-dam-ibigroup:
This PR will publish a new version of printable itinerary that does not rely on momentjs.


miles-grant-ibigroup (approved):
I'm guessing these huge snapshot changes are a result of some of the master merging we've been doing?

philip-cline (approved):

PR #434, opened 2022-08-16
fix(endpoints-overlay): Fix undefined longitude arg on endpoint drag.

binh-dam-ibigroup:
Previous code was sending an undefined longitude after dragging an origin/destination marker.


miles-grant-ibigroup (approved):
Works great! Thanks for finding the fix and cleaning things up in the process

daniel-heppner-ibigroup (approved):
Excellent! This is a bug that I've seen a lot in dev but not in production for some reason. Anyway, very happy that it's fixed, thank you!

PR #436, opened 2022-09-02
Update Internal Dependencies (Sep 2022)

miles-grant-ibigroup:
As part of the MapLibreGL upgrade, we need to update all internal dependencies. This PR also fully removes moment.


daniel-heppner-ibigroup (approved):
"This PR also fully removes moment." I feel like I've seen this one before...

PR #443, opened 2022-09-29
Maplibre transitive route labels

binh-dam-ibigroup:
Blocked by/requires:
- #450
- #451

This PR improves route labels in the transitive overlay:
- horizontal rendering
- better background rendering
- better spacing of route labels when mutliple are rendered by maplibre-gl

### DIFF
--- a/packages/transitive-overlay/src/index.tsx
+++ b/packages/transitive-overlay/src/index.tsx
@@ -157,7 +157,9 @@ const TransitiveCanvasOverlay = ({
         filter={routeFilter}
         id="routes-labels"
         layout={getRouteLayerLayout("name")}
-        paint={{ "text-color": "#eee" }}
+        paint={{
+          "text-color": ["get", "textColor"]
+        }}
         type="symbol"
       />
       <Layer filter={["==", "type", "place"]} id="places" type="circle" />
--- a/packages/transitive-overlay/src/index.tsx
+++ b/packages/transitive-overlay/src/index.tsx
@@ -33,8 +33,6 @@ const TransitiveCanvasOverlay = ({
   transitiveData
 }: Props): JSX.Element => {
   const { current: map } = useMap();
-  // console.log(transitiveData);
-
   const geojson: GeoJSON.FeatureCollection<
     GeoJSON.Geometry,
     Record<any, any>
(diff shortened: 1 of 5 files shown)


packages/transitive-overlay/src/index.tsx
```diff
@@ -160,22 +141,25 @@ const TransitiveCanvasOverlay = ({
...
+          "text-halo-width": 4 // Max value is 1/4 of text size per maplibre docs.
         }}
         type="symbol"
       />
+      <Layer
+        // This layer renders transit route names (foreground).
+        filter={routeFilter}
+        id="routes-labels"
+        layout={getRouteLayerLayout("name")}
+        paint={{ "text-color": "#eee" }}
```
miles-grant-ibigroup, line 160: The text-color here should be imported from the route geojson object

packages/transitive-overlay/src/route-layers.ts
```diff
@@ -0,0 +1,56 @@
...
+  // spaces are replaced with '!' (~same width as space) to create a background with a uniform height.
+  // Also, ensure there is a minimum background width (3 characters).
+  const routeNameUpper = (routeName.length < 3 ? "000" : routeName)
+    .toUpperCase()
+    .replace(/\s/g, "!");
+
+  return {
+    geometry: polyline.toGeoJSON(polyline.encode(concatenatedLines)),
+    properties: {
+      color: `#${route.route_color || "000080"}`,
```
miles-grant-ibigroup, line 35: I would add the text-color here too

miles-grant-ibigroup (changes requested):
Looks really good overall! Just a few minor tweaks, most notably using a standardized text color generation process (which will come in an upcoming core-utils update)

packages/transitive-overlay/src/index.tsx
binh-dam-ibigroup: Yep, now included with https://github.com/opentripplanner/otp-ui/pull/443/commits/470b95aff6343e6124d13d37698502fe4916ebc2.

packages/transitive-overlay/src/route-layers.ts
binh-dam-ibigroup: Included with https://github.com/opentripplanner/otp-ui/pull/443/commits/470b95aff6343e6124d13d37698502fe4916ebc2.

packages/transitive-overlay/src/route-layers.ts
```diff
@@ -0,0 +1,56 @@
...
+  const route = routes.find(r => r.route_id === pattern.route_id);
+  // Concatenate geometries (arrays of coordinates) to help maplibre spread out labels (not perfect).
+  const concatenatedLines = pattern.stops
+    .map(stop => stop.geometry)
+    .filter(geometry => !!geometry)
+    .reduce((result, geom, index) => {
+      const coords = polyline.decode(geom);
+      // Remove the first element (except for the first array) because it is a duplicate
+      // of the last element of the previous array.
+      if (index !== 0) coords.shift();
```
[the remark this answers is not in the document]
binh-dam-ibigroup: If you leave segments of the same route separate, I think maplibre will _try_ to draw a label for each segment. Or at least, that's the clue I get from the screenshots below.

image

miles-grant-ibigroup (approved):
Based on what we discussed about this being a temporary fix, I don't think it makes much sense going through the details of this implementation. It works well across different situations and the labels are definitely more usable than the previous snaky labels. Thanks for the changes!

demory:
Agree with the above comment about this being a temp fix -- the repeated route labels aren't ideal but I don't think it's worth a lot of additional effort. Approving...

demory (approved):

PR #435, opened 2022-08-23
feat(itinerary-body): add optional param to itinerary for collapsing alerts

mcleinman:
This adds an optional param to itineraries, which would force alerts to always open in collapsed form, even if there are 1 or 2 alerts. It also turns a magic number in the code into a named variable, and (hopefully) increases the readability of this code by adding some booleans into some not-immediately-intuitive logic. Also added some stories to storybook to cover the various permutations of this new feature.

Once this goes through review, we'd like to use this functionality at TriMet.

PR #445, opened 2022-09-30
For Raster Tiles: feat(itinerary-body): add optional param to itinerary for collapsing alerts

mcleinman:
This is identical to https://github.com/opentripplanner/otp-ui/pull/435, but branches off the new `raster-tiles` branch so that we're able to use it. (At the time that 435 was reviewed/merged, I didn't fully understand the branching strategy of this.)

----
From original PR:

This adds an optional param to itineraries, which would force alerts to always open in collapsed form, even if there are 1 or 2 alerts. It also turns a magic number in the code into a named variable, and (hopefully) increases the readability of this code by adding some booleans into some not-immediately-intuitive logic. Also added some stories to storybook to cover the various permutations of this new feature.

Once this goes through review, we'd like to use this functionality at TriMet.

### DIFF
--- a/packages/itinerary-body/src/ItineraryBody/index.tsx
+++ b/packages/itinerary-body/src/ItineraryBody/index.tsx
@@ -15,6 +15,7 @@ const ItineraryBody = ({
   accessibilityScoreGradationMap,
   AlertBodyIcon,
   AlertToggleIcon,
+  alwaysCollapseAlerts = false,
   className,
   config,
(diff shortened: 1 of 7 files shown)


mcleinman:
Closing this, as https://github.com/opentripplanner/otp-ui/pull/435 got this work into the main branch. And per discussion elsewhere, TriMet is going to update to the main branch and not continue using raster-tiles version.

PR #446, opened 2022-09-30
Maplibre transitive paths

binh-dam-ibigroup:
This PR improves rendering of transitive paths:
- Walk paths are rendered as dots (or beans).
- A thinner line is applied to bus paths vs a thicker line for other transit modes.
- Map center is fixed for some stories.


miles-grant-ibigroup (approved):
Works great thanks for the changes and isolating the PRs!

daniel-heppner-ibigroup (approved):
Thank you! Good changes. It would be very cool if we could override the line colors and thickness in the config, say for BRT or something.

PR #444, opened 2022-09-30
chore(deps): Unpin icons version in transit vehicle overlay.

binh-dam-ibigroup:
This PR unpins the icons version in the transit-vehicle-overlay package (prevents duplicate icons package downloads).

PR #448, opened 2022-10-03
refactor(itinerary-body): remove direct import of fa-solid

daniel-heppner-ibigroup:
This should fix the problem of the entirety of fa-solid being loaded into the otp-rr bundle.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2832,14 +2832,6 @@
   resolved "https://registry.yarnpkg.com/@open-draft/until/-/until-1.0.3.tgz#db9cc719191a62e7d9200f6e7bab21c5b848adca"
   integrity sha512-Aq58f5HiWdyDlFffbbSjAlv596h/cOnt2DO1w3DOC7OJ5EHs0hd/nycJfiu9RJbT6Yk6F1knnRRXNSpxoIVZ9Q==
 
-"@opentripplanner/icons@1.2.4":
-  version "1.2.4"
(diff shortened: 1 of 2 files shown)


daniel-heppner-ibigroup:
I'm not sure why the yarn.lock changed, but I did use it as the fix commit. Is there something else to investigate there?

yarn.lock
```diff
@@ -2832,6 +2832,14 @@
   resolved "https://registry.yarnpkg.com/@open-draft/until/-/until-1.0.3.tgz#db9cc719191a62e7d9200f6e7bab21c5b848adca"
   integrity sha512-Aq58f5HiWdyDlFffbbSjAlv596h/cOnt2DO1w3DOC7OJ5EHs0hd/nycJfiu9RJbT6Yk6F1knnRRXNSpxoIVZ9Q==
 
+"@opentripplanner/icons@1.2.4":
```
binh-dam-ibigroup, line 2835: You can remove the changes to this file from the PR. The underlying issue is handled by #444.

binh-dam-ibigroup (approved):
LGTM

yarn.lock
daniel-heppner-ibigroup: Sounds good, I will revert it

miles-grant-ibigroup (approved):
Change looks good, thanks for finding that. Approved provided merging master into this before merging this into master (the revert commit makes me nervous -- it may delete the `yarn.lock` changes!)

PR #449, opened 2022-10-04
feat(printable-itinerary): Add Stop ID to printable itinerary

mcleinman:
Adds Stop ID to the printable itineraries


mcleinman:
Closing in favor of https://github.com/opentripplanner/otp-ui/pull/457. Per discussion elsewhere, TriMet is planning on updating so we don't need to use raster-tiles branch.

PR #450, opened 2022-10-05
Add type support for route text color in itinerary leg and transitive

binh-dam-ibigroup:
This PR updates types to include route text color attributes provided from OTP itinerary legs into transitive.


miles-grant-ibigroup (approved):
Looks good thanks for the changes!

daniel-heppner-ibigroup (approved):

PR #451, opened 2022-10-05
Core Utils: add text color generation

miles-grant-ibigroup:
This PR adds a new method to determine a readable text color for a background color. Rather than use standard benchmarks or methods, this method instead uses a carefully calibrated luminance based approach to "approve" color combinations that real agencies use (and that look appealing). The goal is to approve GTFS-based color combinations when they make sense, while being able to supply reasonable alternatives when the GTFS data is missing or of poor quality.

#### Looking for feedback on:

Use of Chroma. It's much smaller than tinycolor, but since otp-rr still uses tinycolor even after using this perhaps it makes more sense to use that larger library here? It's a 14kb pre-compression difference, so not sure if it's the biggest thing to worry about.

### DIFF
--- a/packages/core-utils/src/__tests__/route.js
+++ b/packages/core-utils/src/__tests__/route.js
@@ -193,6 +193,10 @@ describe("util > route", () => {
       expect(getMostReadableTextColor("#00FF00", "#ffffff")).toBe("#000000");
       expect(getMostReadableTextColor("#321b26", "#000000")).toBe("#ffffff");
       expect(getMostReadableTextColor("#321b26", "9e4141")).toBe("#ffffff");
+
+      expect(getMostReadableTextColor("#f00", "#f00")).toBe("#ffffff");
+      expect(getMostReadableTextColor("#fff", "#fff")).toBe("#000000");
+      expect(getMostReadableTextColor("#000", "#000")).toBe("#ffffff");
     });
   });
 });
--- a/packages/core-utils/src/__tests__/route.js
+++ b/packages/core-utils/src/__tests__/route.js
@@ -1,4 +1,5 @@
 import {
+  getMostReadableTextColor,
   getTransitOperatorFromLeg,
   getTransitOperatorFromOtpRoute,
   makeRouteComparator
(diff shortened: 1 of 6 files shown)


packages/core-utils/src/__tests__/route.js
```diff
@@ -135,4 +136,60 @@ describe("util > route", () => {
       expect(sortRoutes(route12, route13)).toMatchSnapshot();
     });
   });
+
+  describe("route text color generation", () => {
```
binh-dam-ibigroup, line 140: Maybe add a comment about where these color combinations and expected values come from.

packages/core-utils/src/__tests__/route.js
```diff
@@ -135,4 +136,60 @@ describe("util > route", () => {
...
+    it("should not modify appropriate color pairings", () => {
+      expect(getMostReadableTextColor("#EE352E", "#ffffff")).toBe("#ffffff");
+      expect(getMostReadableTextColor("#00933C", "#ffffff")).toBe("#ffffff");
+      expect(getMostReadableTextColor("#b933ad", "#ffffff")).toBe("#ffffff");
+      expect(getMostReadableTextColor("#808183", "#ffffff")).toBe("#ffffff");
+      expect(getMostReadableTextColor("#0039a6", "#ffffff")).toBe("#ffffff");
+
+      // Both work
```
binh-dam-ibigroup, line 148: Maybe say instead that the function is indifferent with respect to the proposed text color.

packages/core-utils/src/__tests__/route.js
```diff
@@ -135,4 +136,60 @@ describe("util > route", () => {
...
+      expect(getMostReadableTextColor("#fccc0a", "#000000")).toBe("#000000");
+
+      expect(getMostReadableTextColor("#0075B2", "#ffffff")).toBe("#ffffff");
+      expect(getMostReadableTextColor("#D4A723", "#000000")).toBe("#000000");
+      expect(getMostReadableTextColor("#009D4B", "#ffffff")).toBe("#ffffff");
+      expect(getMostReadableTextColor("#CE242B", "#ffffff")).toBe("#ffffff");
+
+      expect(getMostReadableTextColor("#00FF00", "#000000")).toBe("#000000");
+      expect(getMostReadableTextColor("#009999", "#ffffff")).toBe("#ffffff");
+      expect(getMostReadableTextColor("#009999", "#000000")).toBe("#000000");
```
binh-dam-ibigroup, line 168: Move this one with the "indifferent" colors.

packages/core-utils/src/__tests__/route.js
```diff
@@ -135,4 +136,60 @@ describe("util > route", () => {
...
+      expect(getMostReadableTextColor("#FF00FF", "#000000")).toBe("#000000");
+
+      expect(getMostReadableTextColor("#996633", "#ffffff")).toBe("#ffffff");
+      expect(getMostReadableTextColor("#fccc0a", "#000000")).toBe("#000000");
+
+      expect(getMostReadableTextColor("#0075B2", "#ffffff")).toBe("#ffffff");
+      expect(getMostReadableTextColor("#D4A723", "#000000")).toBe("#000000");
+      expect(getMostReadableTextColor("#009D4B", "#ffffff")).toBe("#ffffff");
+      expect(getMostReadableTextColor("#CE242B", "#ffffff")).toBe("#ffffff");
+
```
binh-dam-ibigroup, line 165: Add some tests with a proposed color that is not black or white.

packages/core-utils/src/route.ts
```diff
@@ -425,3 +426,39 @@ export function makeRouteComparator(
     makeStringValueComparator(obj => obj.longName)
   );
 }
+
+/**
+ * Tests if a pair of colors is readable. If it is, that readble color is returned.
```
binh-dam-ibigroup, line 431: Typo "readable" on second occurrence.

packages/core-utils/src/route.ts
```diff
@@ -425,3 +426,39 @@ export function makeRouteComparator(
...
+ * If it is not, a more appropriate alternative is returned.
+ *
+ * Uses algorithm based on combined luminance. Values have been derived from
+ * looking at real agency color pairings. These pairings are difficult to
+ * generate for, as some colors see both white and black used by different agencies.
```
binh-dam-ibigroup, line 436: Move second and third sentence to tests (see corresponding remark there).

packages/core-utils/src/route.ts
```diff
@@ -425,3 +426,39 @@ export function makeRouteComparator(
...
+ *
+ * This method therefore can accept black and white for the same background color.
```
binh-dam-ibigroup, line 438: How about any other color?

packages/core-utils/src/route.ts
```diff
@@ -425,3 +426,39 @@ export function makeRouteComparator(
...
+ *
+ * When generating colors, white is preferred.
```
binh-dam-ibigroup, line 440: Move this sentence to after line 462.

packages/core-utils/src/route.ts
```diff
@@ -425,3 +426,39 @@ export function makeRouteComparator(
...
+  proposedTextColor = "#ffffff"
+): string {
+  if (!backgroundColor.includes("#")) {
+    backgroundColor = `#${backgroundColor}`;
+  }
+  if (!proposedTextColor.includes("#")) {
+    proposedTextColor = `#${proposedTextColor}`;
+  }
+
+  // Check if proposed color is readable
```
binh-dam-ibigroup, line 455: Do you have the reference for this logic?

packages/core-utils/src/route.ts
```diff
@@ -425,3 +426,39 @@ export function makeRouteComparator(
...
+ *
+ * When generating colors, white is preferred.
+ * @param backgroundColor     A hex string, usually the "routeColor"
+ * @param proposedTextColor   A hex string, usually the "routeTextColor"
+ */
+export function getMostReadableTextColor(
+  backgroundColor: string,
+  proposedTextColor = "#ffffff"
+): string {
+  if (!backgroundColor.includes("#")) {
```
binh-dam-ibigroup, line 448: I think it should be `.startsWith("#")`?

packages/core-utils/src/__tests__/route.js
binh-dam-ibigroup: Better idea, could you replace this test suite into a visual in storybook? (Do a "before" and "after" comparison).

binh-dam-ibigroup (approved):
Approving subject to cleaning up a little bit, but function seems to work. I'm good with using the smaller library. A nice-to-have would be to turn the tests into visuals in storybook.

packages/core-utils/src/__tests__/route.js
miles-grant-ibigroup: I've added an interactive way to test the method.

packages/core-utils/src/__tests__/route.js
miles-grant-ibigroup: Comment expanded

packages/core-utils/src/__tests__/route.js
miles-grant-ibigroup: Good catch thanks

packages/core-utils/src/__tests__/route.js
miles-grant-ibigroup: Couldn't find any agencies with non-black/white pairings. I have one at the end that has a non-bw color. I've also added an interactive way to test the method which allows you to test any color.

packages/core-utils/src/route.ts
miles-grant-ibigroup: Good catch!

packages/core-utils/src/route.ts
miles-grant-ibigroup: Good idea thanks

packages/core-utils/src/route.ts
miles-grant-ibigroup: These aren't common, but I've updated the comment anyway

packages/core-utils/src/route.ts
miles-grant-ibigroup: Done

packages/core-utils/src/route.ts
miles-grant-ibigroup: Comment is updated

packages/core-utils/src/route.ts
miles-grant-ibigroup: Very good catch thank you! Updated

miles-grant-ibigroup:
@philip-cline looks like the new interactive story is causing some trouble in the ci. Should we get rid of it? I'll leave it up to you

packages/core-utils/src/core-utils.story.tsx
```diff
@@ -0,0 +1,64 @@
...
+      </ColorBlock>
+      <ColorBlock fg={getMostReadableTextColor(bg, fg)} bg={bg}>
+        Corrected color pair
+      </ColorBlock>
+    </ColorBlockWrapper>
+  );
+};
+
+export const RouteColorTester = (): JSX.Element => {
+  const [fg, setFg] = useState("#333333");
```
binh-dam-ibigroup, line 35: I would be in favor of adding the colors from the unit test too.

miles-grant-ibigroup: I just updated my response -- unless I used wild hacks I couldn't get these to save to the stories

miles-grant-ibigroup: It doesn't feel like a proper use of stories to me since it's not a component we're testing? Don't stories exist since you can't easily unit test a component? I would think a real unit test is preferred to a story-based solution no?

packages/core-utils/src/core-utils.story.tsx
```diff
@@ -0,0 +1,64 @@
...
+  return (
+    <>
+      <ColorPair bg={bg} fg={fg}></ColorPair>
+      {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
+      <label>
+        Foreground Color
+        <input
+          type="color"
+          value={fg}
+          onChange={e => setFg(e.target.value)}
```
binh-dam-ibigroup, line 46: Sort props (2)

packages/core-utils/src/core-utils.story.tsx
```diff
@@ -0,0 +1,64 @@
...
+        ></input>
```
binh-dam-ibigroup, line 47: Close tag with `/>` (2)

demory:
Testing out this branch in storybook and found at least one case where I didn't get the expected result. Setting both FG and BG to "pure" red (255,0,0) does not correct the foreground color:

image

Trying the same for blue works as expected:

image

demory (approved):

PR #453, opened 2022-10-07
fix(scripts): Separate language columns

binh-dam-ibigroup:
This PR fixes the output of the `collect-i18n-messages.ts` script so that contents for each language appears in its own column (as opposed to concatenated).


philip-cline (approved):

daniel-heppner-ibigroup (approved):
Thank you!

PR #454, opened 2022-10-07
fix(core-utils): Fix itinerary.getLegBounds return type.

binh-dam-ibigroup:
This PR fixes the return type for `itinerary.getLegBounds`.


miles-grant-ibigroup (approved):
Thanks for catching this! This makes sense.

daniel-heppner-ibigroup (approved):
Thank you!

PR #442, opened 2022-09-16
Allow per-vehicle color in transit-vehicle-overlay

mcleinman:
Allow optional per-vehicle colors in transit-vehicle-overlay. This is not a breaking change; it will continue to default to the color provided to the entire object if a vehicle does not have a valid color. I also updated one of the stories to show this new functionality.

This work was forked off transit-vehicle-overlay v2.4.2 (commit 34705faf32635d502e106c30174c7ea525f7a0c5), as it is for the pre-MapLibreGL version of transit-vehicle-overlay. It will need to based off a new branch. Per discussion elsewhere, the new branch will be created after this PR is put up.

PR #458, opened 2022-10-12
feat(transit-vehicle-overlay): Add optional per-vehicle color

mcleinman:
Allow optional per-vehicle colors in transit-vehicle-overlay. This is not a breaking change; it will continue to default to the color provided to the entire object if a vehicle does not have a valid color. I also updated one of the stories to show this new functionality.

This is similar to https://github.com/opentripplanner/otp-ui/pull/442. We're likely going to update to a non-raster-tiles version, so we need this functionality in the main branch as well.


binh-dam-ibigroup:
> Allow optional per-vehicle colors in transit-vehicle-overlay. 

Hi @mcleinman, thanks for this PR. By curiosity, could you provide a clarification of what the highlight color on one vehicle, but not others on the same transit route, is intended to mean? Thank you!

PR #461, opened 2022-10-14
Trip-details use CO2 from Itinerary

daniel-heppner-ibigroup:
This makes a change to the trip details package to use the co2 value included in the itinerary that is passed in from OTP-RR. That way this value only gets calculated once, higher up. The co2 property is added to the type to accommodate this change.

### DIFF
--- a/packages/location-field/src/types.ts
+++ b/packages/location-field/src/types.ts
@@ -155,7 +155,7 @@ export interface LocationFieldProps {
   locationType: LocationType;
   /**
    * A list of stopIds of the stops that should be shown as being nearby. These
-   * must be referencable in the stopsIndex prop.
+   * must be referenceable in the stopsIndex prop.
    */
   nearbyStops?: string[];
   /**
--- a/packages/trip-details/src/index.tsx
+++ b/packages/trip-details/src/index.tsx
@@ -232,12 +232,6 @@ export function TripDetails({
     walkDuration
   } = coreUtils.itinerary.calculatePhysicalActivity(itinerary);
 
-  const co2 = coreUtils.itinerary.calculateEmissions(
-    itinerary,
-    co2Config?.carbonIntensity,
-    co2Config?.units
-  );
-
   // Parse flex info and generate appropriate strings
   const containsFlex = itinerary.legs.some(coreUtils.itinerary.isFlex);
   const pickupBookingInfo = itinerary.legs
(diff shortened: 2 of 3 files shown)


packages/location-field/src/types.ts
```diff
@@ -155,7 +155,7 @@ export interface LocationFieldProps {
   locationType: LocationType;
   /**
    * A list of stopIds of the stops that should be shown as being nearby. These
-   * must be referencable in the stopsIndex prop.
+   * must be referenceable in the stopsIndex prop.
```
daniel-heppner-ibigroup, line 158: codespell has once again woken from its slumber to find spelling problems in completely separate parts of the code from the one i edited

packages/trip-details/src/index.tsx
```diff
@@ -232,12 +232,6 @@ export function TripDetails({
     walkDuration
   } = coreUtils.itinerary.calculatePhysicalActivity(itinerary);
 
-  const co2 = coreUtils.itinerary.calculateEmissions(
-    itinerary,
-    co2Config?.carbonIntensity,
-    co2Config?.units
-  );
-
```
miles-grant-ibigroup, line 240: Do you think it's a good idea to still do co2 calculation if `carbonIntensity` is defined but `itinerary.co2` isn't?

packages/types/src/index.ts
```diff
@@ -381,6 +381,8 @@ export type Fare = {
  * http://otp-docs.ibi-transit.com/api/json_Itinerary.html
  */
 export type Itinerary = {
+  co2?: number;
+  co2VsBaseline?: number;
```
miles-grant-ibigroup, line 385: Is this used anywhere?

miles-grant-ibigroup (approved):
Works for me, just a few small nitpicks

packages/types/src/index.ts
daniel-heppner-ibigroup: The value is used in OTP-RR in an un-typescripted file, so at some point if we typescript that file we'll be grateful for this property being here.

packages/trip-details/src/index.tsx
daniel-heppner-ibigroup: Sure! I can add this back in with a check.

amy-corson-ibigroup (approved):

PR #463, opened 2022-10-18
Adjust endpoint icon size

binh-dam-ibigroup:
This PR adjusts the size of the endpoint icons to match the transit stop and rental place markers in the original transitive overlay, and cleans up the related CSS.

### DIFF
--- a/packages/endpoints-overlay/src/index.tsx
+++ b/packages/endpoints-overlay/src/index.tsx
@@ -85,7 +85,7 @@ function DefaultMapMarkerIcon({
   switch (type) {
     case "to": {
       // The fa-solid's "location-dot" icon used here has a width-height ratio of 3/4,
-      // so the desired width for the outline/"stacked" element is 4/3 of the pixels prop.
+      // so the desired width for the outline/"stacked" element is 4/3 of the "apparent" width.
       const toPixels = pixels * 1.3;
       inner = (
         <>
--- a/packages/endpoints-overlay/src/index.tsx
+++ b/packages/endpoints-overlay/src/index.tsx
@@ -80,17 +80,17 @@ function DefaultMapMarkerIcon({
   location,
   type
 }: UserLocationAndType): ReactElement {
-  const pixels = 20;
+  const PIXELS = 20;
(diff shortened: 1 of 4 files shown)


packages/endpoints-overlay/src/index.tsx
```diff
@@ -80,22 +80,27 @@ function DefaultMapMarkerIcon({
...
   type
 }: UserLocationAndType): ReactElement {
+  const pixels = 20;
   let inner;
   switch (type) {
-    case "to":
+    case "to": {
+      // The fa-solid's "location-dot" icon used here has a width-height ratio of 3/4,
+      // so the desired width for the outline/"stacked" element is 4/3 of the "apparent" width.
+      const toPixels = pixels * 1.3;
```
miles-grant-ibigroup, line 89: Screen Shot 2022-10-20 at 2 26 49 PM
I think this is causing some strange behavior in safari

packages/endpoints-overlay/src/index.tsx
```diff
@@ -80,22 +80,27 @@ function DefaultMapMarkerIcon({
   location,
   type
 }: UserLocationAndType): ReactElement {
+  const pixels = 20;
```
miles-grant-ibigroup, line 83: Should this be an all caps constant?

miles-grant-ibigroup (changes requested):
Seeing some weird behavior on webkit. Otherwise seems to all be working...

packages/endpoints-overlay/src/index.tsx
binh-dam-ibigroup: Safari doesn't round CSS px units like the other browsers! Fixed in https://github.com/opentripplanner/otp-ui/pull/463/commits/e52e9233d55d2480c62822f7328fa912a0114728.

miles-grant-ibigroup (approved):
Looks great! Thanks for the changes.

daniel-heppner-ibigroup (approved):
Looks good to me (on Webkit)
Kinda silly that we have to calculate the size like that.

PR #447, opened 2022-10-03
Support Specifying Pelias Layers

miles-grant-ibigroup:
Pelias requests by default include things like continents and nations. These are not always wanted. This PR adds support for adding a list of wanted layers to the geocoder configuration.

PR #464, opened 2022-10-18
fix(geocoder): remove invalid layers from default layers list

miles-grant-ibigroup:
#447 contained a bug which sent invalid requests to the geocoder. This PR resolves this oversight.


daniel-heppner-ibigroup (approved):
oops

PR #467, opened 2022-10-25
Simplify itinerary translations

binh-dam-ibigroup:
This PR attempts to remove some conditional statements from the language files, so that translations are a bit easier (at the expense of some extra code).

### DIFF
--- a/packages/itinerary-body/src/defaults/line-column-content.tsx
+++ b/packages/itinerary-body/src/defaults/line-column-content.tsx
@@ -1,11 +1,49 @@
 import LocationIcon from "@opentripplanner/location-icon";
 import React, { ReactElement } from "react";
-import { useIntl } from "react-intl";
+import { IntlShape, useIntl } from "react-intl";
 
 import RouteBadge from "../RouteBadge";
 import * as S from "../styled";
 import { LineColumnContentProps } from "../types";
 import { defaultMessages } from "../util";
+
+/**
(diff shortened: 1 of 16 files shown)


packages/itinerary-body/src/defaults/line-column-content.tsx
```diff
@@ -1,12 +1,50 @@
...
+      return intl.formatMessage({
+        defaultMessage:
+          defaultMessages["otpUi.ItineraryBody.travelByMode.walk"],
+        description: "Travel by walking",
+        id: "otpUi.ItineraryBody.travelByMode.walk"
+      });
+    default:
+      return modeId;
+  }
+}
```
daniel-heppner-ibigroup, line 46: Confirming this is different from the other switch in the access leg?

daniel-heppner-ibigroup (approved):
LGTM. I think this should help a lot with our translation push, since those complicated formatjs strings can be really confusing when you're looking at them on their own (and even when you're looking at them with the code!)

packages/itinerary-body/src/defaults/line-column-content.tsx
binh-dam-ibigroup: Yes, unfortunately this one has to be separate from the vehicle type because it contains a preposition that can change with the mode (see the French entries for instance).

amy-corson-ibigroup (approved):

PR #468, opened 2022-10-25
fix: support bounds updating on small screens

miles-grant-ibigroup:
The 200 units padding caused some problems on smaller screens where maplibre refused to update the bounds. This PR checks the screen width and only adds the padding if the window width is large enough to accommodate it.

### DIFF
--- a/packages/route-viewer-overlay/src/index.tsx
+++ b/packages/route-viewer-overlay/src/index.tsx
@@ -76,7 +76,7 @@ const RouteViewerOverlay = (props: Props): JSX.Element => {
 
         current?.fitBounds(bounds, {
           duration: 500,
-          padding: 200
+          padding: window.innerWidth > 500 ? 200 : undefined
         });
 
         if (props.mapCenterCallback) {
--- a/packages/transitive-overlay/tsconfig.json
+++ b/packages/transitive-overlay/tsconfig.json
@@ -2,7 +2,7 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     "composite": true,
-    "lib": ["es2019"],
+    "lib": ["es2019", "dom"],
     "outDir": "./lib",
     "rootDir": "./src",
     "skipLibCheck": true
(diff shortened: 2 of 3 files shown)


packages/route-viewer-overlay/src/index.tsx
```diff
@@ -76,7 +76,7 @@ const RouteViewerOverlay = (props: Props): JSX.Element => {
 
         current?.fitBounds(bounds, {
           duration: 500,
-          padding: 200
+          padding: window.innerWidth > 500 ? 200 : undefined
```
daniel-heppner-ibigroup, line 79: Is there some clever way we should think about doing this using a formula? Maybe a percentage of screen width? I'm also fine with this.

daniel-heppner-ibigroup (approved):

packages/route-viewer-overlay/src/index.tsx
miles-grant-ibigroup: Yeah we could try to somehow set the padding dynamically based on how much free room there is. For now though even this is still causing some problems in a few cases. 

Once we've resolved all the mobile view weirdnesses we should definitely do this!

binh-dam-ibigroup, line 79: Do you think these could become a ratio of, I guess both the width and height of the window?

miles-grant-ibigroup: It's possible, but I'd want to debug the issue more. I can't quite figure out what about the map size makes maplibregl fail to set the bounds. The 500 value seems to cover most failures. I don't think screen height affects the bug.

packages/transitive-overlay/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     "composite": true,
-    "lib": ["es2019"],
+    "lib": ["es2019", "dom"],
```
binh-dam-ibigroup, line 5: What is this change for?

miles-grant-ibigroup: `window` gets marked as an unknown variable without it

binh-dam-ibigroup (approved):
The improvements on the transitive stories are nice, let's go with them for now.

PR #469, opened 2022-10-25
fix(base-map): support onContextMenu in touch env

miles-grant-ibigroup:
Right clicking is not possible with a touch environment, so we fake it by handling long presses. The 600ms timeout is based on https://developer.android.com/reference/android/view/ViewConfiguration.html#getLongPressTimeout%28%29


daniel-heppner-ibigroup (approved):
Looks good, but how does it work with multitouch? I'm not sure how to test on a mobile device.

miles-grant-ibigroup:
@daniel-heppner-ibigroup I tested it on a touch device and it handles multitouch fine. The second finger triggers `onMove` which cancels the timer.

amy-corson-ibigroup (approved):

PR #441, opened 2022-09-15
(transit-vehicle-overlay): support ambient vehicle rendering

miles-grant-ibigroup:
This PR adds a new mode to render transit vehicles without a hover effect.

PR #470, opened 2022-10-27
Transit vehicle heading

binh-dam-ibigroup:
Requires/blocked by #441.

This PR enhances the package to further customize rendering of transit vehicle symbols.
Let me know if prop names, method names, default values need to be tweaked.

The following behaviors are possible and can be customized using the provided or outside components:
- Render route short names only, with auto-fit inside the icon container/shape component
- Render route symbols, optionally using custom `ModeIcon` components (with a default mode and fallback to route names if a mode is not found)
- Use of route color as a fixed background or on hover only
- Rotate symbols, or use a caret that indicates the orientation/heading of a vehicle
- Customizable symbol size/padding


packages/transit-vehicle-overlay/src/BasicModeIcon.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Fill-in type for the icons package.
+ */
+export interface ModeIconProps {
+  mode?: string;
+}
+
+/**
+ * Default, minimal mode icon renderer.
```
miles-grant-ibigroup, line 20: Is this not already present within the icons package?

packages/transit-vehicle-overlay/src/styled.ts
```diff
@@ -0,0 +1,136 @@
...
+  &::before {
+    border-bottom: 5px solid #333;
+    border-left: 6px solid transparent;
+    border-right: 6px solid transparent;
+    content: "";
+    height: 0;
+    left: 50%;
+    margin-left: -6px;
+    position: absolute;
+    top: 0px;
```
miles-grant-ibigroup, line 91: Screen Shot 2022-11-01 at 8 23 53 AM
With this change, it's possible to get the arrow to cleanly "attach" to the border!

```suggestion
    top: -6px;
```

miles-grant-ibigroup: The background color would also have to be set to match the border.

packages/transit-vehicle-overlay/src/styled.ts
```diff
@@ -0,0 +1,136 @@
...
+    width: 0;
+  }
+`;
+
+const routeColorBackgroundCss = css<ColorProps>`
+  background: ${props => props.backgroundColor};
```
miles-grant-ibigroup, line 97: Is it possible to add some transparency here? The extreme color blocks are quite jarring and some transparency helps the icons blend in better.
```suggestion
  background: ${props => props.backgroundColor}aa;
```

packages/transit-vehicle-overlay/src/styled.ts
```diff
@@ -0,0 +1,136 @@
...
+  return styled(Container).attrs(getColorProps(defaultColor))`
+    ${routeColorBackgroundCss}
+  `;
+}
+
+/**
+ * Applies the vehicle's route color a background on hovering a component
+ * and a foreground color that is contrast compatible with that color.
+ */
+export function withRouteColorBackgroundOnHover(
```
miles-grant-ibigroup, line 127: Could the `:hover` not be added dynamically within `withRouteColorBackground` based on an argument in that method?

packages/transit-vehicle-overlay/src/styled.ts
```diff
@@ -0,0 +1,136 @@
...
+  background: #eee;
+  border: 2px solid #333;
+  border-radius: ${getPixels}px;
+  cursor: default;
+  height: ${getPixels}px;
+  line-height: ${getPixels}px;
+  padding: ${getPadding}px;
+  position: relative;
+  text-align: center;
+  transition: all 0.1s ease-in-out;
```
miles-grant-ibigroup, line 60: Could the transition be extended just a little bit to perhaps .15 or .2 seconds? Especially since both text color and background color change, the rapid transition can be a bit extreme

packages/transit-vehicle-overlay/src/index.tsx
```diff
@@ -37,29 +79,39 @@ const TransitVehicleOverlay = ({
...
 export default TransitVehicleOverlay;
+
+// Export the other subcomponents
+export {
+  CircleWithCaret,
+  DefaultVehicleIcon,
+  RotatingCircle,
+  RouteNumberIcon,
+  withRouteColorBackground,
+  withRouteColorBackgroundOnHover
```
miles-grant-ibigroup, line 116: Does all of this need to be exported? Are we planning on using it elsewhere?

miles-grant-ibigroup (changes requested):
I like the new configuration options, I think they make this overlay much more mature. Just some cosmetic changes for now.

packages/transit-vehicle-overlay/src/index.tsx
binh-dam-ibigroup: Yes, that's for use in OTP-RR!

packages/transit-vehicle-overlay/src/BasicModeIcon.tsx
binh-dam-ibigroup: Should we make the `ModeIcon` prop mandatory then?

packages/transit-vehicle-overlay/src/styled.ts
binh-dam-ibigroup: Sure, changed in https://github.com/opentripplanner/otp-ui/pull/470/commits/abd1d3fc03f9a6430464461e3c852aa127b9022b.

packages/transit-vehicle-overlay/src/styled.ts
binh-dam-ibigroup: Good idea, thanks, updated in https://github.com/opentripplanner/otp-ui/pull/470/commits/abd1d3fc03f9a6430464461e3c852aa127b9022b.

packages/transit-vehicle-overlay/src/styled.ts
binh-dam-ibigroup: Added the caret position option in https://github.com/opentripplanner/otp-ui/pull/470/commits/fd50aeab09828df5e4317249b698e852d787b10f.

binh-dam-ibigroup: With that change, only inner carets get their foreground matched with the route number/icon.

packages/transit-vehicle-overlay/src/BasicModeIcon.tsx
binh-dam-ibigroup: https://github.com/opentripplanner/otp-ui/pull/470/commits/28312b082dc609ced2e367e583908d4463e4036e removes `BasicModeIcon` and makes the `ModeIcon` prop mandatory, but I'm realizing that it is should not be for cases where users use the `RouteNumberIcon` component instead...

packages/transit-vehicle-overlay/src/WithCaret.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const offset = options?.offset || 0;
+  const RawCaret = isInner ? InnerCaret : OuterCaret;
+  const SizedCaret = height
+    ? styled(RawCaret)`
+        &::before {
+          border-bottom-width: ${Math.max(0, height)}px;
+          border-left-width: ${halfWidth}px;
+          border-right-width: ${halfWidth}px;
+          margin-left: -${halfWidth}px;
+          top: -${offset + (isInner ? 0 : height)}px;
```
miles-grant-ibigroup, line 37: I think this might be causing some issues with custom sized outer carets -- with inner carets things work fine. Screen Shot 2022-11-07 at 8 41 42 AM

packages/transit-vehicle-overlay/src/styled.ts
```diff
@@ -0,0 +1,154 @@
...
+import { FC } from "react";
+import styled, { css } from "styled-components";
+
+import { VehicleComponentProps } from "./types";
+
+interface RouteColorBackgroundOptions {
+  /**
+   * The alpha component of a color in hexadecimal.
+   */
+  alphaHex?: string;
```
miles-grant-ibigroup, line 12: Could we possibly use alphaHex in a story?

miles-grant-ibigroup (approved):
Looks good! A few bugs that arise when using some of the configuration options in tandem, but nothing blocking I don't feel.

Overall I'm a bit weary of adding too many configuration options but again I don't see that as blocking and defaults are something we can fix in future PRs.

packages/transit-vehicle-overlay/src/WithCaret.tsx
binh-dam-ibigroup: Correct, the implementer has to choose an offset and size that look good together.

packages/transit-vehicle-overlay/src/styled.ts
binh-dam-ibigroup: It is used in the route color on-hover background. Should there be a separate story?

miles-grant-ibigroup: Oh I missed that sorry! Yes then we're good to go.

demory:
Overall this looks great. One design/styling question. Personally, I prefer the caret outside the circle, with less padding around the icon. For one thing, this eliminates the possibility of collision between the icon/label and caret:

image

I was able to accomplish the above by changing the height/width/top/left values for `Caret` as follows:

image

My question is, could this be accomplished by an implementing user solely via CSS (and/or other configuration options)? I still don't fully understand how styled-components works vis a vis overriding styles using classnames.

miles-grant-ibigroup:
@demory Binh and I had some back and forth about caret placement: https://github.com/opentripplanner/otp-ui/pull/470#discussion_r1015509167

It looks like the default position is not ideal, but it can be configured to achieve the look of being "attached" to the border

binh-dam-ibigroup:
> [...] Personally, I prefer the caret outside the circle, with less padding around the icon. For one thing, this eliminates the possibility of collision between the icon/label and caret:
> 
> image
> 
> [...] 
> My question is, could this be accomplished by an implementing user solely via CSS (and/or other configuration options)? I still don't fully understand how styled-components works vis a vis overriding styles using classnames.

I like how this one looks too, let me make that the default. 

Note that the caret size and position can be requested at will from the overlay component without using CSS or styled-components (an implementer would create the caret based on some config parameters). Also, the icon size and padding are also props of the overlay component, see this fragment below:
```
<TransitVehicleOverlay
  IconContainer={withCaret(Circle, { height: 15, offset: 4, width: 10 })}
  iconPixels={20}
  iconPadding={2}
  ...
/>
```

binh-dam-ibigroup:
> Overall this looks great. One design/styling question. Personally, I prefer the caret outside the circle, with less padding around the icon. For one thing, this eliminates the possibility of collision between the icon/label and caret:
> 
> image
> 

This is now the default as of https://github.com/opentripplanner/otp-ui/pull/470/commits/e068fc4cad4f3af949056886f931d88da9d61b6c.

PR #471, opened 2022-10-28
Improve popup style

binh-dam-ibigroup:
This PR improves the map popup component with shading and rounded border per https://github.com/opentripplanner/otp-react-redux/pull/667#discussion_r1000960712.

### DIFF
--- a/packages/base-map/src/index.tsx
+++ b/packages/base-map/src/index.tsx
@@ -190,4 +190,6 @@ const LayerWrapper = (props: LayerProps): JSX.Element => {
   return <>{visible && children}</>;
 };
 
+export const Popup = Styled.Popup;
+
 export { LayerWrapper, MarkerWithPopup, Styled };
--- a/packages/base-map/src/styled.ts
+++ b/packages/base-map/src/styled.ts
@@ -8,7 +8,7 @@ export const Popup = styled(MapGlPopup)`
   & > .maplibregl-popup-content,
   & > .mapboxgl-popup-content {
     border-radius: 10px;
-    box-shadow: 0 3px 14px 4px rgb(0 0 0 / 40%);
+    box-shadow: 0 3px 14px 4px rgb(0 0 0 / 20%);
   }
 `;
 
--- a/packages/base-map/src/styled.ts
+++ b/packages/base-map/src/styled.ts
@@ -7,8 +7,8 @@ import styled from "styled-components";
 export const Popup = styled(MapGlPopup)`
   & > .maplibregl-popup-content,
   & > .mapboxgl-popup-content {
-    border-radius: 6px;
-    box-shadow: 0 3px 14px rgb(0 0 0 / 40%);
+    border-radius: 10px;
+    box-shadow: 0 3px 14px 4px rgb(0 0 0 / 40%);
   }
 `;
 
--- a/packages/base-map/src/styled.ts
+++ b/packages/base-map/src/styled.ts
@@ -1,4 +1,16 @@
+import { Popup as MapGlPopup } from "react-map-gl";
 import styled from "styled-components";
+
+/**
+ * Adds a box shadow and tweaks border radius to make popups easier to read.
+ */
+export const Popup = styled(MapGlPopup)`
+  & > .maplibregl-popup-content,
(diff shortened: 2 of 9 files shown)


packages/base-map/src/styled.ts
```diff
@@ -1,5 +1,17 @@
...
 import styled from "styled-components";
 
+/**
+ * Adds a box shadow and tweaks border radius to make popups easier to read.
+ */
+export const Popup = styled(MapGlPopup)`
+  & > .maplibregl-popup-content,
+  & > .mapboxgl-popup-content {
+    border-radius: 6px;
+    box-shadow: 0 3px 14px rgb(0 0 0 / 40%);
```
miles-grant-ibigroup, line 11:
```suggestion
    box-shadow: 0 3px 14px 4px rgb(0 0 0 / 40%);
```
Could we add some spread to this?

packages/base-map/src/styled.ts
```diff
@@ -1,5 +1,17 @@
+import { Popup as MapGlPopup } from "react-map-gl";
 import styled from "styled-components";
 
+/**
+ * Adds a box shadow and tweaks border radius to make popups easier to read.
+ */
+export const Popup = styled(MapGlPopup)`
+  & > .maplibregl-popup-content,
+  & > .mapboxgl-popup-content {
+    border-radius: 6px;
```
miles-grant-ibigroup, line 10:
```suggestion
    border-radius: 10px;
```
I think if we bump this up we should commit to a more round look

packages/base-map/src/index.tsx
```diff
@@ -190,4 +190,6 @@ const LayerWrapper = (props: LayerProps): JSX.Element => {
   return <>{visible && children}</>;
 };
 
+export const Popup = Styled.Popup;
```
miles-grant-ibigroup, line 193: I'm not seeing this in the `stop-overlay` or `vehicle-rental-overlay`

miles-grant-ibigroup (changes requested):
I like the look but I'd like to see it used in all applicable packages!

packages/base-map/src/index.tsx
binh-dam-ibigroup: These packages need a new version number before they can use the popup with the new style (see `index.tsx` in this PR from the `base-map` package).

binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-ui/pull/471/commits/96fe48e52e433946a26d2b7199136f20574e3443 for storybook viewing only.

packages/base-map/src/styled.ts
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-ui/pull/471/commits/45ca046b59d4c46d94a43292084bf17a1e50cec0.

packages/base-map/src/styled.ts
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-ui/pull/471/commits/45ca046b59d4c46d94a43292084bf17a1e50cec0.

miles-grant-ibigroup: Sorry I forgot to add one more change I made which is to change 40% to 20%! which I think looks a bit less jarring.

miles-grant-ibigroup (approved):
With the drop shadow color change this is perfect and the non fix/feat commit on the other packages is the perfect way to get those changes through!

Thanks so much

packages/base-map/src/styled.ts
binh-dam-ibigroup: In that case could we revert to the initial shadow which has more contrast and a narrower span?

miles-grant-ibigroup: I think with 20% instead of 40% the new one looks much more defined! Just needs to be a little more transparent

binh-dam-ibigroup: Ok, see https://github.com/opentripplanner/otp-ui/pull/471/commits/539a6ecb056e69a35cbafe6db03d6024676b30ee.

amy-corson-ibigroup (approved):

PR #472, opened 2022-10-28
OTP2 Tile Overlay

miles-grant-ibigroup:
Adds a new overlay which allows rendering of tiles rendered by OTP2 (http://docs.opentripplanner.org/en/latest/sandbox/MapboxVectorTilesApi/)

Render is an array of elements, as the `base-map` can only show/hide layers that are unique components. 

I'm not sure what the best way to test this is -- we could attempt to store some mocked OTP2 tile overlay data, or perhaps unit test some of the individual components. Curious what others think.

Also adds a new `map-popup` package to centralize the code for every both types of popup (stop and rental vehicle). MapPopup is used in `StopOverlay`, `VehicleRentalOverlay`, and the new `Otp2TileOverlay`


packages/otp2-tile-overlay/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "tsc": "tsc"
+  },
+  "peerDependencies": {
+    "react": "^16.14.0",
+    "react-map-gl": "^7.0.15"
+  },
+  "dependencies": {
+    "@opentripplanner/stops-overlay": "^5.0.3-alpha.1",
+    "@opentripplanner/vehicle-rental-overlay": "^2.1.0-alpha.1"
+  },
```
binh-dam-ibigroup, line 21: Hmm I am not sure I like the dependency to these overlays. Intuitively, it looks like it should be the other way around instead?

packages/otp2-tile-overlay/src/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+        filter={network ? ["all", ["==", "network", network]] : ["all"]}
+        id={id}
+        key={id}
+        paint={LAYER_PAINT[type]}
+        source="otp2-tiles"
+        source-layer={type}
+        type="circle"
+      />
+      {clickedEntity && (
+        <Popup
```
binh-dam-ibigroup, line 105: I feel strongly about having this package not display popups at all. Actually, we had a related team discussion about removing popups from the `endpoint-overlay` package that goes in the same direction. 

The reason for that was to not pollute the component with handlers and contents that are implementer-specific, such as `setLocation` or the popups below.

We should instead publish an event such as `onEntityClick`, and demo handling these events in storybook using popups.

packages/otp2-tile-overlay/src/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+  /**
+   * A method fired when the stop viewer is opened in the popup
+   */
+  setViewedStop: ({ stopId }: { stopId: string }) => void
+  type: string
+}): JSX.Element => {
+  const { current: map } = useMap()
+
+  // TODO: handle this complex type: it can be a stop, a station, and some extra fields too
+  const [clickedEntity, setClickedEntity] = useState<any>(null)
```
binh-dam-ibigroup, line 46: With the recommendation of not displaying popups, this state should be removed and the setter becomes the callback function for the published entity click event.

packages/otp2-tile-overlay/src/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+            synthesizedEntity.formFactor === "SCOOTER",
+          // OTP1 compatibility -- will get overwritten if possible
+          networks: [synthesizedEntity.network],
+          ...synthesizedEntity,
+          // OTP1 compatibility
+          bikesAvailable: synthesizedEntity.vehiclesAvailable,
+          // OTP1 compatibility
+          x: synthesizedEntity.lon,
+          // OTP1 compatibility
+          y: synthesizedEntity.lat
```
binh-dam-ibigroup, line 87: Depending on how the overlay is intended to be used, this block might be better suited as a prop.

packages/otp2-tile-overlay/src/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+      }
+    })
+  }, [id, map])
+
+  return (
+    <>
+      <Layer
+        filter={network ? ["all", ["==", "network", network]] : ["all"]}
+        id={id}
+        key={id}
```
binh-dam-ibigroup, line 98: Do we need a `key` prop?

binh-dam-ibigroup (commented):
I think we'll have to discuss how these overlays are meant to be used individually.

packages/otp2-tile-overlay/package.json
miles-grant-ibigroup: I've added a comment that the overlays should be deprecated once OTP1 support is removed, and the overlays merged into a new one that lives inside the OTP2 tile layer.

packages/otp2-tile-overlay/src/index.tsx
miles-grant-ibigroup: I've made the new props optional so that the layer can operate in a "minimal" mode where only the title is displayed.

packages/otp2-tile-overlay/src/index.tsx
miles-grant-ibigroup: Have added a comment to refactor the popups into a single coherent popup with no prop-rewriting

packages/otp2-tile-overlay/src/index.tsx
miles-grant-ibigroup: I've added a comment that it would be best to remove it

packages/otp2-tile-overlay/src/index.tsx
miles-grant-ibigroup: Yes. `base-map` removes the fragment

packages/otp2-tile-overlay/src/Otp2TileOverlay.story.tsx
```diff
@@ -0,0 +1,35 @@
...
+  title: "OTP2 Tile Layer"
+};
+
+export const OtpTileLayer = (): JSX.Element => {
+  const [endpoint, setEndpoint] = useState("");
+  return (
+    <>
+      {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
+      <label>
+        OTP2 Server with stops layer enabled
```
binh-dam-ibigroup, line 15: Instead of/in addition to using a server, I recommend including a mock of a simple tile layer with a few entities, as we already do with the overlay packages to illustrate the inputs.

packages/vehicle-rental-overlay/src/StationPopup.tsx
```diff
@@ -1,146 +0,0 @@
-import { Styled as BaseMapStyled } from "@opentripplanner/base-map";
```
binh-dam-ibigroup, line 1: This file was involved in commit https://github.com/opentripplanner/otp-ui/pull/472/commits/671c6aeeafe3c0bddebf80cb1f43c96d4e45f2a8 which is a `feat` commit, are we expecting to issue a feature release for the vehicle rental overlay package?

binh-dam-ibigroup (commented):
It doesn't feel like `vehicle-rental-overlay` should receive a `feature` release, what do you think? (see comment below)

.storybook/react-intl.js
```diff
@@ -20,6 +20,7 @@ const packages = [
   "itinerary-body",
   "location-field",
   "printable-itinerary",
+  'map-popup',
```
binh-dam-ibigroup, line 23: Use double quotes.

packages/map-popup/src/MapPopup.story.tsx
```diff
@@ -0,0 +1,79 @@
...
+export default {
+  title: "Map Popup"
+};
+
+const STOP = {
+  id: "9526",
+  name: "W Burnside & SW 2nd",
+  lat: 45.523009,
+  lon: -122.672529,
+  flex: false
```
binh-dam-ibigroup, line 14: Sort attributes (multiple instances).

packages/map-popup/src/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+// HACK: We should flatten the messages loaded above because
+// the YAML loaders behave differently between webpack and our version of jest:
+// - the yaml loader for webpack returns a nested object,
+// - the yaml loader for jest returns messages with flattened ids.
+export const defaultMessages: { [key:string]:string } = flatten(defaultEnglishMessages);
+
+type Entity = Stop | Station
+type Props = {
+  configCompanies?: ConfiguredCompany[];
```
binh-dam-ibigroup, line 25: Sort props

packages/map-popup/src/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+  setViewedStop?: ({ stopId }: { stopId: string }) => void;
+
+  entity: Entity
+};
+
+/**
+ * Renders a map popup for a stop, scooter, or shared bike
+ */
+export function MapPopup({ configCompanies, entity, getEntityName, setLocation, setViewedStop, }: Props): JSX.Element {
+  if (!entity) return <></>
```
binh-dam-ibigroup, line 38: Can the return be null instead?

packages/map-popup/src/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+  const intl = useIntl()
+
+  const getNameFunc = getEntityName || makeDefaultGetEntityName(intl, defaultMessages);
+  const name = getNameFunc(entity, configCompanies);
+
+  const generateLocation = () => {
```
binh-dam-ibigroup, line 44: Move method outside and let it accept `entity` and `name` as params.

packages/map-popup/src/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+  const entityIsStop = !("bikesAvailable" in entity)
+  // @ts-expect-error ts doesn't understand entityIsStop
+  const stopId = entityIsStop && entity?.code || entity.id.split(":")[1] || entity.id
+
+  return (
+    <BaseMapStyled.MapOverlayPopup>
+      <BaseMapStyled.PopupTitle>{name}</BaseMapStyled.PopupTitle>
+      {/* render dock info if it is available */}
+      {entityIsStationHub && (
+        <BaseMapStyled.PopupRow>
```
binh-dam-ibigroup, line 65: Each of these sub cases should be a component that renders the specific type of entity.

packages/map-popup/src/MapPopup.story.tsx
```diff
@@ -0,0 +1,79 @@
...
+  isFloatingBike: true,
+  isCarStation: false,
+  realTimeData: true,
+  "stroke-width": 1,
+  color: "#f00"
+};
+
+export const StopEntity = (): JSX.Element => {
+  return (
+    <MapPopupContents
```
binh-dam-ibigroup, line 53: These should be rendering different components, one for each entity type.

packages/map-popup/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@opentripplanner/map-popup",
+  "version": "0.0.1",
```
binh-dam-ibigroup, line 3: This could be `1.0.0` I suppose?

packages/otp2-tile-overlay/src/index.tsx
```diff
@@ -0,0 +1,182 @@
+import {
+  ConfiguredCompany,
+  MapLocationActionArg,
+} from "@opentripplanner/types"
+import { Layer, Popup,  Source, useMap } from "react-map-gl"
```
binh-dam-ibigroup, line 5: es-lint didn't catch the extra space?

packages/otp2-tile-overlay/src/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+import EntityPopup from "@opentripplanner/map-popup"
```
binh-dam-ibigroup, line 6: Sort imports by package name.

packages/otp2-tile-overlay/src/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+import React, { useEffect, useState } from "react"
+// eslint-disable-next-line prettier/prettier
+import type { EventData } from "mapbox-gl"
+import { LAYER_PAINT } from "./util"
+
+const OTP2TileLayerWithPopup = ({
+  configCompanies,
```
binh-dam-ibigroup, line 13: Sort props (multiple instances)

packages/otp2-tile-overlay/src/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+  useEffect(() => {
+    map?.on("mouseenter", id, () => {
+      map.getCanvas().style.cursor = "pointer"
+    })
+    map?.on("mouseleave", id, () => {
+      map.getCanvas().style.cursor = ""
+    })
+    map?.on("click", id, onMapClick || defaultClickHandler)
+  }, [id, map])
```
binh-dam-ibigroup, line 109: Remove event handlers when this component unmounts.

packages/otp2-tile-overlay/src/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+          onClose={() => setClickedEntity(null)}
+        >
+            <EntityPopup
+              configCompanies={configCompanies}
+              setViewedStop={setViewedStop}
+              setLocation={setLocation ? (location) => {
+                setClickedEntity(null)
+                setLocation(location)
+              } : null}
+              entity={{ ...clickedEntity, id: clickedEntity?.id || clickedEntity?.gtfsId }}
```
binh-dam-ibigroup, line 137: Sort props

binh-dam-ibigroup (changes requested):
I need to review more, but her is my first round of comments.

packages/map-popup/package.json
miles-grant-ibigroup: Will add a "breaking change"

packages/map-popup/src/index.tsx
miles-grant-ibigroup: Nope, this causes errors.

packages/vehicle-rental-overlay/src/StationPopup.tsx
miles-grant-ibigroup: I think we should -- this is quite a change as it brings in new dependencies and no longer exports the popup contents itself.

packages/otp2-tile-overlay/src/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+        filter={network ? ["all", ["==", "network", network]] : ["all"]}
+        id={id}
+        key={id}
+        paint={LAYER_PAINT[type]}
+        source="otp2-tiles"
+        source-layer={type}
+        type="circle"
+      />
+      {clickedEntity && (
+        <Popup
```
binh-dam-ibigroup, line 123: I am wondering if `<Popup>` could be nested inside `<Layer>`, that would clean up the fragment tags.

binh-dam-ibigroup (changes requested):
Thanks for starting the leg work on this. My main requests are:
1. Convert `generateOTP2TileLayers` into a component,
2. Accept some tilejson/mock pbf file to make testing easier,
3. Separate rendering of entity details by entity type,
4. CSS a11y tweaks on the "View Stop" button.

packages/otp2-tile-overlay/src/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+  /**
+   * An optional method to override the map click handler. If a method is passed, NO POPUPS
+   * WILL APPEAR ON CLICK. The implementer will be responsible for handling all click events
+   * in accordance with the MapLibreGL api.
+   */
+  onMapClick?: (event: EventData) => void
+  /**
+   * Determines which layer of the OTP2 tile data to display. Also determines icon color.
+   */
+  type: string
```
binh-dam-ibigroup, line 53: Should this a list of permitted values instead, i.e. all the known OTP layer types?

packages/otp2-tile-overlay/src/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+          setLocation={setLocation}
+          setViewedStop={setViewedStop}
+          type={type}
+        />
+      )
+    })
+  ]
+}
+
+export default generateOTP2TileLayers
```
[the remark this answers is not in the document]
miles-grant-ibigroup: This sounds great, but unfortunately breaks the layer selector. The layer selector relies on each `Source` having a unique id and being a direct child of `BaseMap`. I would have done it like this if it weren't for this restriction.

packages/otp2-tile-overlay/src/index.tsx
miles-grant-ibigroup: I've tried this and it doesn't work.

packages/otp2-tile-overlay/src/index.tsx
miles-grant-ibigroup: See my comment on the main comment about this. This annoying complexity has to live somewhere -- the only way to avoid using this method is to re-write the base map layer selector to require additional configuration which I am very opposed to.

packages/otp2-tile-overlay/src/index.tsx
miles-grant-ibigroup: How would you detect this?

packages/map-popup/src/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+              defaultMessage={defaultMessages["otpUi.MapPopup.stopId"]}
+              description="Displays the stop id"
+              id="otpUi.MapPopup.stopId"
+              values={{
+                stopId
+              }}
+            />
+          </strong>
+        </span>
+        <S.ViewStopButton onClick={() => setViewedStop(stopId)}>
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Seems not to have caused any problems...

packages/map-popup/src/index.tsx
```diff
@@ -0,0 +1,129 @@
+import React from "react";
+
+import { Styled as BaseMapStyled } from "@opentripplanner/base-map";
+import FromToLocationPicker from "@opentripplanner/from-to-location-picker";
+// eslint-disable-next-line prettier/prettier
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Prettier doesn't understand `type`

packages/map-popup/src/index.tsx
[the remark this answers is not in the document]
miles-grant-ibigroup: Not sure what the benefit here would be? These popups are getting created and destroyed so rapidly I don't see what benefit caching serves

packages/otp2-tile-overlay/src/Otp2TileOverlay.story.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export default {
+  title: "OTP2 Tile Layer"
+};
+
+export const OtpTileLayer = (): JSX.Element => {
+  const [endpoint, setEndpoint] = useState("");
+  return (
+    <>
+      {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
```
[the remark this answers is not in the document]
miles-grant-ibigroup: This would be out of line with how we do things in other packages. Maybe this can be part of a greater a11y refactor?

packages/otp2-tile-overlay/src/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+  configCompanies?: ConfiguredCompany[]
+): JSX.Element[] => {
+  return [
+    <Source
+      // @ts-expect-error we use a nonstandard prop
+      alwaysShow
+      id="otp2-tiles"
+      key="otp2-tiles"
+      type="vector"
+      url={`${endpoint}/${layers.map((l) => l.type).join(",")}/tilejson.json`}
```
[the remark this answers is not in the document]
miles-grant-ibigroup: I will pass mock data via `msw` the way we do for `location-field`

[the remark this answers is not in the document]
miles-grant-ibigroup: That way we can avoid new props

packages/otp2-tile-overlay/src/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+  }, [id, map])
+
+  return (
+    <>
+      <Layer
+        filter={network ? ["all", ["==", "network", network]] : ["all"]}
+        id={id}
+        key={id}
+        paint={LAYER_PAINT[type]}
+        source="otp2-tiles"
```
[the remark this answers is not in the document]
miles-grant-ibigroup: That's correct, but this isn't possible without a significant `base-map` refactor

packages/otp2-tile-overlay/src/index.tsx
miles-grant-ibigroup: No, as these are fully configurable in the OTP config

packages/otp2-tile-overlay/src/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+            />
+         
+        </Popup>
+      )}
+    </>
+  )
+}
+
+const generateOTP2TileLayers = (
+  layers: { name?: string; network?: string; type: string }[],
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Layer type names are freely definable in OTP

packages/otp2-tile-overlay/src/Otp2TileOverlay.story.tsx
miles-grant-ibigroup: This was much more work than expected but we now mock tile server data using `msw`

binh-dam-ibigroup (commented):
Additional comments for now. I think it would make sense to do a separate PR for the map popup stuff so there can be a release number that you can use for the tiles overlay.

packages/base-map/src/index.tsx
```diff
@@ -91,8 +91,14 @@ const BaseMap = ({
...
     ? children
-        .flat()
-        .filter(child => child?.props?.id !== undefined)
+        .flat(10)
+        .filter(
+          child =>
+            child?.props?.id !== undefined &&
+            // Some sources will not have layers as children, and should be ignored
+            // from the list.
+            child?.props?.alwaysShow !== true
```
binh-dam-ibigroup, line 100: Rename `alwaysShow` to `notListed`? Or, if this condition is targeting `<Source>` elements, could we check `child?.type.name` and `child?.props?.children` directly?

Screenshot 2022-12-21 at 9 24 04 AM

packages/otp2-tile-overlay/src/index.tsx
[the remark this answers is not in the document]
binh-dam-ibigroup: I give in on this one.

binh-dam-ibigroup (changes requested):
See various comments. I think the main request now is to splitting the PR so that changes to `base-map` and `map-popup` get new version numbers for the tile layer package to reference.

packages/base-map/src/index.tsx
miles-grant-ibigroup: It's more common to want to have a layer off by default than on so I did it this way to avoid writing out too many props. 

It's also used on `Layer`s that shouldn't be ridable!

packages/map-popup/src/index.tsx
```diff
@@ -0,0 +1,136 @@
+import React from "react";
+
+import { Styled as BaseMapStyled } from "@opentripplanner/base-map";
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Someone would have to be doing something really strange to use this package without `base-map`. I don't support duplicating these styles, especially as they're being imported by a lot of other layers.

packages/map-popup/src/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+  setLocation?: ({ location, locationType }: { location: Location, locationType: string }) => void;
+  setViewedStop?: ({ stopId }: { stopId: string }) => void;
+};
+
+/**
+ * Renders a map popup for a stop, scooter, or shared bike
+ */
+export function MapPopup({ configCompanies, entity, getEntityName, setLocation, setViewedStop, }: Props): JSX.Element {
+  if (!entity) return <></>
+  const intl = useIntl()
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Good catch thanks!

.storybook/react-intl.js
```diff
@@ -20,6 +20,7 @@ const packages = [
   "itinerary-body",
   "location-field",
   "printable-itinerary",
+  "map-popup",
   "stops-overlay",
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Even if it's using this package doesn't `StopsOverlay` still render internationalized content?

miles-grant-ibigroup:
"prerequisites" PR is open #518. Will merge that with your single approval and then update this PR.

binh-dam-ibigroup (approved):
Thanks for the heavy lifting and changes on this PR!

amy-corson-ibigroup (approved):
One small suggestion for the storybook map, but overall LGTM!

packages/otp2-tile-overlay/src/Otp2TileOverlay.story.tsx
```diff
@@ -0,0 +1,41 @@
...
+          value={endpoint}
+        />
+      </label>
+      <pre style={{ display: "inline", paddingLeft: 5 }}>
+        https://fake-otp-server.com/otp
+      </pre>{" "}
+      mocks downtown Philadelphia stop tile data. Check stops in the layer
+      selector and zoom in on downtown Philadelphia (near city hall) to see some
+      stops loaded from OTP tiles.
+      <BaseMap center={[0, 0]} zoom={3} style={{ height: "80vh" }}>
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Yes, please, thanks!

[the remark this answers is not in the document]
miles-grant-ibigroup: Unfortunately the mocked tiles only work when zooming in from the 0,0 origin :( 

I tried doing this initially, but the tiles become harder to mock as the requests get less predictable. Maybe this can be a future improvement.

PR #474, opened 2022-10-31
fix(trip-details): repair co2 regression

miles-grant-ibigroup:
`trip-details` was crashing when `co2config` was missing. This PR repairs this issue.


daniel-heppner-ibigroup (approved):
Thanks for the fix!

amy-corson-ibigroup (approved):

PR #475, opened 2022-11-02
Update base-map in pkgs that use its Popup or MarkerWithPopup.

binh-dam-ibigroup:
This PR updates the base-map version for packages that use it's new `Popup` component that displays a shadow.
(Other packages that use the map but not the popups are not touched.)


miles-grant-ibigroup (approved):
Works great thanks for handling the logistics of this

PR #480, opened 2022-11-10
Fix flex indicator conditions

philip-cline:
This PR:
- Adds phone-ahead drop-off types to the `isReservationRequired` method. 
- Refactors `isContinuousDropoff` to be `isCoordinationRequired` which is closer to the true behaviour of service using these fields.
- Adds condition for coordination with driver on pickup. 

This is a breaking change since the isContinuousDropoff will no longer work. A companion PR will update this in OTP-RR.


miles-grant-ibigroup:
@philip-cline this PR is blocking https://github.com/opentripplanner/otp-react-redux/pull/661

What's the status of this PR? I am available for review and @daniel-heppner-ibigroup who is reviewing the stop reviewer refresh can likely also review this

philip-cline:
This PR is finished and is ready for review. I can assign you as first reviewer :).

miles-grant-ibigroup:
I just realized this should probably be a breaking change -- the flex method names change!

amy-corson-ibigroup (approved):

PR #482, opened 2022-11-15
fix(core-utils): prefer 24 hour time parsing

miles-grant-ibigroup:
This PR resolves the infamous "am/pm" bug we've been seeing. Times such as `8:24` were being parsed as 8pm instead of 8am when the processing happened after 12pm because 12 hour time was being assumed. This PR corrects this.

### DIFF
--- a/packages/core-utils/src/query.js
+++ b/packages/core-utils/src/query.js
@@ -34,6 +34,8 @@ export const defaultParams = [
  */
 const TIME_FORMATS = [
   "HH:mm:ss",
+  "HH:mm",
+  "H:mm",
   "h:mm:ss a",
   "h:mm:ssa",
   "h:mm a",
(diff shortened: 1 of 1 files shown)


packages/core-utils/src/query.js
```diff
@@ -34,6 +34,8 @@ export const defaultParams = [
  */
 const TIME_FORMATS = [
   "HH:mm:ss",
+  "HH:mm",
```
binh-dam-ibigroup, line 37: Move `OTP_DATE_TIME_FORMAT` here or delete that last entry.

binh-dam-ibigroup (approved):
Good find, thanks!

packages/core-utils/src/query.js
miles-grant-ibigroup: Good catch thanks! Have opted to remove as the constant is used as a fallback anyway.

PR #485, opened 2022-11-18
Fix route overlay rendering

binh-dam-ibigroup:
Partial fix for https://github.com/opentripplanner/otp-react-redux/issues/703.

With this PR, the calculation of the geometry for `map.fitBounds` is improved so that the "inner" region for rendering a route is based on the size of the map canvas and directly related to the size of the map canvas. This should reduce instances of "empty" render regions which previously prevented re-rendering when new route data is provided.

Also, explicit map updates are requested after `map.fitBounds` is called (otherwise, logging map bounds would often show the correct coordinates before the canvas is updated).

### DIFF
--- a/packages/route-viewer-overlay/src/RouteViewerOverlay.story.js
+++ b/packages/route-viewer-overlay/src/RouteViewerOverlay.story.js
@@ -46,10 +46,14 @@ const WithChangingRoute = () => {
   const [rtData, setRouteData] = useState(routeData);
 
   useEffect(() => {
-    setTimeout(() => {
-      setRouteData(routeData2);
+    const interval = setInterval(() => {
+      setRouteData(rtData === routeData2 ? routeData : routeData2);
     }, 5000);
-  }, []);
+
+    return () => {
+      clearInterval(interval);
+    };
+  }, [rtData]);
 
   return <RouteViewerOverlay routeData={rtData} />;
 };
--- a/packages/route-viewer-overlay/src/RouteViewerOverlay.story.js
+++ b/packages/route-viewer-overlay/src/RouteViewerOverlay.story.js
@@ -1,6 +1,8 @@
-import React from "react";
+import React, { useEffect, useState } from "react";
 
 import routeData from "../__mocks__/mock-route.json";
+import routeData2 from "../__mocks__/mock-route2.json";
+import routeDataOtp2 from "../__mocks__/mock-route-otp2.json";
 import flexRouteData from "../__mocks__/mock-flex-route.json";
 
 import StopsOverlay from "../../stops-overlay/src";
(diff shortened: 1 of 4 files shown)


packages/route-viewer-overlay/src/RouteViewerOverlay.story.js
```diff
@@ -38,6 +40,20 @@ const Template = args => (
   </>
 );
 
+/** Should zoom to pattern. */
+
+const WithChangingRoute = () => {
+  const [rtData, setRouteData] = useState(routeData);
+
+  useEffect(() => {
+    setTimeout(() => {
```
miles-grant-ibigroup, line 49: Could we possibly swap this to an interval that switches back and forth?

packages/route-viewer-overlay/src/index.tsx
```diff
@@ -58,33 +58,51 @@ const RouteViewerOverlay = (props: Props): JSX.Element => {
...
+    }
 
-        current?.fitBounds(bounds, {
-          duration: 500,
-          padding: window.innerWidth > 500 ? 200 : undefined
-        });
+    // if pattern geometry updated, update the map points
+    if (bounds && current) {
+      const canvas = current.getCanvas();
+      // @ts-expect-error getOixelRatio not defeined in MapRef type.
```
miles-grant-ibigroup, line 83: Typo!

packages/route-viewer-overlay/src/index.tsx
```diff
@@ -58,33 +58,51 @@ const RouteViewerOverlay = (props: Props): JSX.Element => {
...
+
+      // Often times, the map is not updated right away, so try to force an update.
+      current.triggerRepaint();
+
+      if (props.mapCenterCallback) {
+        props.mapCenterCallback();
       }
     }
-  }, [routeData.patterns]);
+  }, [routeData, routeData.patterns, current]);
```
miles-grant-ibigroup, line 105: Why did we add `routeData.patterns` here?

miles-grant-ibigroup (changes requested):
Clever hack! Have tested it in otp-rr and it seems to function more consistently. Thanks for your research!

packages/route-viewer-overlay/src/index.tsx
binh-dam-ibigroup: You would want to refit if the selected pattern from the pattern selector changes. That changes the value in `routeData.patterns`, but not `routeData`.

packages/route-viewer-overlay/src/index.tsx
binh-dam-ibigroup: Fixed in https://github.com/opentripplanner/otp-ui/pull/485/commits/fb1c25df7b28b974859562e58e13e0e468377953.

packages/route-viewer-overlay/src/RouteViewerOverlay.story.js
binh-dam-ibigroup: Sure, updated in https://github.com/opentripplanner/otp-ui/pull/485/commits/fb1c25df7b28b974859562e58e13e0e468377953.

binh-dam-ibigroup:
> Clever hack! Have tested it in otp-rr and it seems to function more consistently. Thanks for your research!

It still misses some renderings on mine, but it is definitely better than before!

miles-grant-ibigroup (approved):
Tested it again and it works great! Thanks for the changes

binh-dam-ibigroup:
Blocking this PR for #488, so that the map fitting code can be reused in multiple packages.

binh-dam-ibigroup:
Going to merge this PR in a bit by internal request. When #488 merges I'll issue another fix PR.

PR #486, opened 2022-11-18
fix(stops-overlay): Remove event handlers on component unmount.

binh-dam-ibigroup:
In this PR, event handlers that are set up when mounting the stops overlay are removed when the component unmounts.


daniel-heppner-ibigroup (approved):
Thank you for this fix!

amy-corson-ibigroup (approved):

PR #488, opened 2022-11-21
fix(base-map): Add helper methods to fit map to bounds

binh-dam-ibigroup:
This fix-release PR adds helper methods to compute padding of a mapbox `MapRef` and fit bounds on a map using the padding parameters and attempt forcing a redraw. Such methods are used in the route viewer (I want to reuse this with #485), trip viewer, and transitive overlays.

### DIFF
--- a/packages/base-map/src/index.tsx
+++ b/packages/base-map/src/index.tsx
@@ -3,7 +3,7 @@ import { Map, MapProps } from "react-map-gl";
 import maplibregl, { Event } from "maplibre-gl";
 
 import * as Styled from "./styled";
-import callIfValid from "./util";
+import * as util from "./util";
 import MarkerWithPopup from "./MarkerWithPopup";
 
 /**
@@ -38,13 +38,11 @@ type Props = React.ComponentPropsWithoutRef<React.ElementType> & {
   // Unknown is used here because of a maplibre/mapbox issue with the true type, MapLayerMouseEvent
   onContextMenu?: (e: unknown) => void;
   /** A callback method which is fired when the map zoom or map bounds change */
-  // TODO: does this cause integration issues?
-  onViewportChanged?: (e: maplibregl.MapLibreEvent) => void;
+  onViewportChanged?: (e: State) => void;
   /** An initial zoom value for the map */
   zoom?: number;
 };
 type State = {
-  fitBoundsOptions?: Record<string, number | string | boolean>;
   latitude: number;
   longitude: number;
   zoom: number;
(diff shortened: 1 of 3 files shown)


packages/base-map/src/util.ts
```diff
@@ -1,9 +1,38 @@
...
+
+/**
+ * Helper function used in several packages to fit a map to the given bounds,
+ * using padding based on map canvas size.
+ * @param map The map where the bounds fitting is to occur.
+ * @param bounds The bounds to be fit.
+ */
+export function fitMapBounds(map: MapRef, bounds: LngLatBoundsLike): void {
+  map.fitBounds(bounds, {
+    duration: 500,
```
miles-grant-ibigroup, line 32: Can we re-add `duration: 300,
       essential: false,
       maxDuration: 600,`?

miles-grant-ibigroup: This allows the duration length to flex a little depending on how far the map has to pan.

packages/base-map/src/index.tsx
```diff
@@ -38,13 +38,11 @@ type Props = React.ComponentPropsWithoutRef<React.ElementType> & {
   // Unknown is used here because of a maplibre/mapbox issue with the true type, MapLayerMouseEvent
   onContextMenu?: (e: unknown) => void;
   /** A callback method which is fired when the map zoom or map bounds change */
-  // TODO: does this cause integration issues?
-  onViewportChanged?: (e: maplibregl.MapLibreEvent) => void;
+  onViewportChanged?: (e: State) => void;
```
miles-grant-ibigroup, line 41: Why was this change made?

miles-grant-ibigroup (changes requested):
Looks good on first glance, but I'm worried about regressions on this so will continue to look more carefully

packages/base-map/src/util.ts
binh-dam-ibigroup: I did not have the impression that these defaults, also set somewhere, were used previously, were they?

packages/base-map/src/index.tsx
binh-dam-ibigroup: The params we pass when calling our `onViewportChanged` event were effectively of type `State`.

packages/base-map/src/util.ts
binh-dam-ibigroup: In the only non-story places where we call `fitBounds` (route overlay, trip overlay, transitive overlay), we have been passing `duration: 500` and have not used the other options mentioned above.

miles-grant-ibigroup: It's possible? I'm not sure. 500ms is fine for now.

packages/base-map/src/index.tsx
miles-grant-ibigroup: Sounds good

miles-grant-ibigroup (approved):
Have tested this a bit more and things are working well! Thanks for cleaning things up

daniel-heppner-ibigroup (approved):
Thanks! Works well for me and looks good.

PR #492, opened 2022-11-28
fix(route-viewer-overlay): Fix auto-fitting of OTP2 flex routes.

### LINKED ISSUE #440: Route Viewer fails to pan to flex-only route
If a route has no stops defined, the route viewer will fail to automatically pan to that route when it is loaded. The solution is to check for a location-only route and if one is detected to pan to that location.

binh-dam-ibigroup:
Attempt to fix #440.
I am missing mock data for the kind of route described in the linked issue, feel free to suggest if you have that handy.


miles-grant-ibigroup (approved):
Very clean and works great. Thanks for a great fix!

amy-corson-ibigroup (approved):

PR #452, opened 2022-10-06
Maplibre fix place labels

binh-dam-ibigroup:
Now requires #454.

This PR improves label placement and cleans up the code for origin/destination, transit and rental leg places. What leg extremities get rendered (label and place circle) should be the same as what is at https://opentripplanner.org/otp-ui, with the addition of the labels for the origin and destination of an itinerary.

PR #496, opened 2022-11-28
Support empty transitive data

binh-dam-ibigroup:
Fix a regression introduced by #452, and future proof the handling of null transitive data input.

### DIFF
--- a/packages/transitive-overlay/src/index.tsx
+++ b/packages/transitive-overlay/src/index.tsx
@@ -86,7 +86,7 @@ const accessLegFilter = [
 
 type Props = {
   activeLeg?: Leg;
-  transitiveData: TransitiveData;
+  transitiveData?: TransitiveData;
 };
 
 const TransitiveCanvasOverlay = ({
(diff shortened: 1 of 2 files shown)


packages/transitive-overlay/src/index.tsx
```diff
@@ -199,6 +201,10 @@ const TransitiveCanvasOverlay = ({
     zoomToGeoJSON(polyline.toGeoJSON(activeLeg.legGeometry.points));
   }, [activeLeg]);
 
+  if (!transitiveData) return <></>;
```
daniel-heppner-ibigroup, line 204: Would it be possible to just put this at the top, along with anything that uses `activeLeg`, so that it's not necessary to have the big ternary above?

daniel-heppner-ibigroup (approved):
Looks good, just had a thought on making it cleaner, but maybe not possible.

packages/transitive-overlay/src/index.tsx
binh-dam-ibigroup: Good question! Unfortunately, [all hooks have to be rendered outside conditions, before `return` etc](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level).

amy-corson-ibigroup (approved):

PR #497, opened 2022-11-29
Restore arc rendering for flex segments

miles-grant-ibigroup:
The temporary move away from `transitive.js` included the loss of the transitive.js arc drawing algorithm. This PR restores that functionality by using a new algorithm based on turf methods.

In a dream world the arc follows the original path somewhat, but I'd need some help figuring out the trigonometry on that one... It's likely some sort of interpolated smoothing operation? Not even sure which textbook to reach for there


daniel-heppner-ibigroup (approved):
It looks good to me in the storybook! Thank you for figuring this out!

packages/transitive-overlay/src/route-layers.ts
```diff
@@ -32,16 +39,58 @@ export function patternToRouteFeature(
...
+    textColor: `#${route.route_text_color || "eee"}`,
+    type: "route"
+  };
+
+  const isFlex = pattern.stops[pattern.stops.length - 1].stop_id.includes(
+    "flexed_to"
+  );
+  const straight = polyline.toGeoJSON(polyline.encode(concatenatedLines));
+
+  if (isFlex) {
```
binh-dam-ibigroup, line 56: (non-blocking) I think TNC itineraries used to be drawn using arcs too. Could you add a TNC condition (at the right spot) too? Or create an issue specific to TNCs.

packages/transitive-overlay/src/route-layers.ts
```diff
@@ -32,16 +39,58 @@ export function patternToRouteFeature(
...
+  const properties = {
+    color: `#${route.route_color || "000080"}`,
+    name: routeName,
+    nameUpper: routeNameUpper,
+    routeType: route.route_type,
+    textColor: `#${route.route_text_color || "eee"}`,
+    type: "route"
+  };
+
+  const isFlex = pattern.stops[pattern.stops.length - 1].stop_id.includes(
```
binh-dam-ibigroup, line 51: Per the code, this can probably be `.endsWith`.

packages/transitive-overlay/src/route-layers.ts
```diff
@@ -32,16 +39,58 @@ export function patternToRouteFeature(
...
+    "flexed_to"
+  );
+  const straight = polyline.toGeoJSON(polyline.encode(concatenatedLines));
+
+  if (isFlex) {
+    // Create clone of plain route that only includes first and last point
+    const extraStraight = { ...straight };
```
binh-dam-ibigroup, line 58: This variable may not be necessary. You are modifying the coordinates of the original one below already, and `straight` is not used past this `if` block.

packages/transitive-overlay/src/route-layers.ts
```diff
@@ -32,16 +39,58 @@ export function patternToRouteFeature(
...
+    extraStraight.coordinates = [
+      extraStraight.coordinates[0],
+      extraStraight.coordinates[extraStraight.coordinates.length - 1]
```
binh-dam-ibigroup, line 61: Maybe extract `const`s for these two coordinates, they are used often by themselves.

binh-dam-ibigroup (approved):
Looks neat! See suggested tweaks, and should we include TNCs too in this PR?

packages/transitive-overlay/src/route-layers.ts
miles-grant-ibigroup: Have added a comment and an issue

packages/transitive-overlay/src/route-layers.ts
miles-grant-ibigroup: Good call! This was a remnant of my testing

packages/transitive-overlay/src/route-layers.ts
miles-grant-ibigroup: Much cleaner, thanks!

PR #498, opened 2022-11-29
Common fit bounds

binh-dam-ibigroup:
This PR updates the route, trip and transitive overlay packages to use the fit bounds methods from #488. This will also fix bounds fitting on mobile for the trip and transitive overlay packages.


miles-grant-ibigroup (approved):
Works great in testing and much cleaner. Thanks for cleaning this up!

daniel-heppner-ibigroup (approved):
Great work! I like how this cleaned up.

PR #499, opened 2022-12-02
Support Multiple Base Layers in BaseMap

miles-grant-ibigroup:
This PR adds support for multiple base layers. Configuration is done by allowing `baseLayer` to accept an array of strings instead of just a string.

A new prop `baseLayerNames` is added to allow the control to display pretty names. An alternative might be to use internationalization with the url as a key, but I think it's easier to push that responsibility up a level? Curious for others' thoughts.
Screen Shot 2022-12-02 at 11 33 10 AM

### DIFF
--- a/packages/base-map/src/index.story.tsx
+++ b/packages/base-map/src/index.story.tsx
@@ -185,20 +185,20 @@ export const withMultipleBaseLayers = () => {
   return (
     <Styled.StoryMapContainer>
       <input
+        onChange={e => setMapTilerKey(e.target.value)}
+        placeholder="MapTiler API Key"
         type="text"
         value={mapTilerKey}
-        placeholder="MapTiler API Key"
-        onChange={e => setMapTilerKey(e.target.value)}
       />
       {mapTilerKey && (
         <BaseMap
-          center={center}
           baseLayer={[
             `https://api.maptiler.com/maps/streets/style.json?key=${mapTilerKey}`,
             `https://api.maptiler.com/maps/ocean/style.json?key=${mapTilerKey}`,
             `https://api.maptiler.com/maps/hybrid/style.json?key=${mapTilerKey}`
           ]}
           baseLayerNames={["Streets", "Ocean", "Hybrid"]}
+          center={center}
         />
       )}
     </Styled.StoryMapContainer>
@@ -210,20 +210,20 @@ export const withMultipleBaseLayersAndOptionalLayers = () => {
   return (
     <Styled.StoryMapContainer>
       <input
+        onChange={e => setMapTilerKey(e.target.value)}
+        placeholder="MapTiler API Key"
         type="text"
         value={mapTilerKey}
-        placeholder="MapTiler API Key"
-        onChange={e => setMapTilerKey(e.target.value)}
       />
       {mapTilerKey && (
         <BaseMap
-          center={center}
           baseLayer={[
             `https://api.maptiler.com/maps/streets/style.json?key=${mapTilerKey}`,
             `https://api.maptiler.com/maps/ocean/style.json?key=${mapTilerKey}`,
             `https://api.maptiler.com/maps/hybrid/style.json?key=${mapTilerKey}`
           ]}
           baseLayerNames={["Streets", "Ocean", "Hybrid"]}
+          center={center}
         >
           <AllVehiclesOverlay id="layer-1" />
           <LayerWrapper id="layer-2">
--- a/packages/base-map/src/index.story.tsx
+++ b/packages/base-map/src/index.story.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable react/button-has-type */
-import React from "react";
+import React, { useState } from "react";
 import {
   AttributionControl,
   MapProvider,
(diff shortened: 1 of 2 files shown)


demory (approved):

demory:
I approved this but will add one suggestion: maybe add a second item to storybook that demos both the base layer selector and one or more overlays together?

packages/base-map/src/index.story.tsx
```diff
@@ -179,3 +179,28 @@ export const withOptionalControls = () => (
...
 withOptionalControls.decorators = [withMap()];
+
+export const withMultipleBaseLayers = () => {
+  const [mapTilerKey, setMapTilerKey] = useState("");
+  return (
+    <Styled.StoryMapContainer>
+      <input
+        type="text"
+        value={mapTilerKey}
+        placeholder="MapTiler API Key"
```
binh-dam-ibigroup, line 190: Sort props (multiple instances).

packages/base-map/src/index.tsx
```diff
@@ -23,7 +23,9 @@ import MarkerWithPopup from "./MarkerWithPopup";
  */
 type Props = React.ComponentPropsWithoutRef<React.ElementType> & {
   /** A URL pointing to the vector tile specification which should be used as the main map.  */
```
binh-dam-ibigroup, line 25: Update the JSDoc for arrays of URLs.

packages/base-map/src/index.tsx
```diff
@@ -102,6 +105,9 @@ const BaseMap = ({
   const [hiddenLayers, setHiddenLayers] = useState(
     toggleableLayers.filter(layer => !layer?.visible).map(layer => layer.id)
   );
+  const [activeBaseLayer, setActiveBaseLayer] = useState(
+    typeof baseLayer === "object" ? baseLayer?.[0] : baseLayer
```
binh-dam-ibigroup, line 109: From the type definitions, there is an assumption that `baseLayer` can be `null`. Is that correct? Also, `typeof null === 'object'` is true. So, a null check is probably needed on lines 142 and 154. Alternately, create a variable for whether `baseLayer` is a non-null array and use that throughout this component.

packages/base-map/src/index.tsx
```diff
@@ -143,6 +150,24 @@ const BaseMap = ({
           <ul
             className={`layers-list ${fakeMobileHover && "fake-mobile-hover"}`}
           >
+            {typeof baseLayer === "object" &&
+              baseLayer.map((layer: string, index: number) => {
+                return (
+                  <li key={index}>
+                    {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
+                    <label>
```
binh-dam-ibigroup, line 158: Go ahead and add the `htmlFor` prop to the label, even if it already contains the input control, and delete the a11y exception.

packages/base-map/src/index.tsx
```diff
@@ -143,6 +150,24 @@ const BaseMap = ({
...
+                      <input
+                        type="radio"
+                        checked={activeBaseLayer === layer}
```
binh-dam-ibigroup, line 161: Sort props.

packages/base-map/src/index.tsx
binh-dam-ibigroup: (Might do that for the checkbox labels below too)

packages/base-map/src/index.tsx
```diff
@@ -143,6 +150,24 @@ const BaseMap = ({
           <ul
             className={`layers-list ${fakeMobileHover && "fake-mobile-hover"}`}
           >
+            {typeof baseLayer === "object" &&
+              baseLayer.map((layer: string, index: number) => {
+                return (
+                  <li key={index}>
```
binh-dam-ibigroup, line 156: We can probably use `layer` as the key instead of the index.

packages/base-map/src/index.tsx
```diff
@@ -23,7 +23,9 @@ import MarkerWithPopup from "./MarkerWithPopup";
  */
 type Props = React.ComponentPropsWithoutRef<React.ElementType> & {
   /** A URL pointing to the vector tile specification which should be used as the main map.  */
-  baseLayer?: string;
+  baseLayer?: string | string[];
+  /** A list of names to match onto the base layers */
```
binh-dam-ibigroup, line 27: Also say in the JSDoc that `baseLayerNames` is used only if multiple entries are defined for `baseLayer`.

binh-dam-ibigroup (approved):
Approving subject to some code cleanup as I indicated and subject to @demory's suggestion to add a storybook example showing multiple base layers with overlays.

packages/base-map/src/index.tsx
miles-grant-ibigroup: Very good catch thank you! Forgot `null` is an object how we love javascript

packages/base-map/src/index.tsx
miles-grant-ibigroup: Why are we changing this now? We've been using this pattern everywhere in the project so far

packages/base-map/src/index.tsx
miles-grant-ibigroup: I had this but key `0` caused problems

PR #501, opened 2022-12-02
fix(trip-details): add export for fare by leg table

daniel-heppner-ibigroup:
Community Transit wants to use the fare by leg table in their custom webapp, so we need this fare by leg table to be exported. (I thought it was already exported and I remember there being a story for it, any idea where that went?)

### DIFF
--- a/packages/trip-details/src/TripDetails.story.tsx
+++ b/packages/trip-details/src/TripDetails.story.tsx
@@ -10,7 +10,7 @@ import {
 import styled from "styled-components";
 // The below eslint-disable is due to https://github.com/storybookjs/storybook/issues/13408
 // eslint-disable-next-line import/no-named-as-default
-import TripDetails from ".";
+import TripDetails, { FareLegTable } from ".";
 import * as TripDetailsClasses from "./styled";
 import {
   CaloriesDetailsProps,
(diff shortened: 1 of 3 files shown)


packages/trip-details/src/TripDetails.story.tsx
```diff
@@ -348,3 +348,14 @@ export const FareComponentsItinerary = makeStory({
...
 
 export const OTP2FlexItinerary = makeStory({ itinerary: flexItinerary });
+
+export const FareLegTableStory = (): ReactElement => {
+  return (
+    <FareLegTable
+      layout={fareByLegLayout}
+      legs={walkInterlinedTransitItinerary.legs}
+      transitFareDetails={walkInterlinedTransitItinerary.fare?.details}
+      transitFares={walkInterlinedTransitItinerary.fare?.fare}
```
miles-grant-ibigroup, line 358: Oy vey...

miles-grant-ibigroup (approved):
Looks great! Hope my test snapshot updates fix the ci...

packages/trip-details/src/TripDetails.story.tsx
daniel-heppner-ibigroup: I have no idea why I did it like this... It definitely seems like it could have just accepted the itinerary. I could make a bug for that

PR #502, opened 2022-12-02
fix(route-viewer-overlay): Support OTP1 route patterns with stops.

binh-dam-ibigroup:
This fixes a bug introduced by #492 that crashes the route viewer overlay when rendering patterns with stops without geometries.

### DIFF
--- a/packages/route-viewer-overlay/src/index.tsx
+++ b/packages/route-viewer-overlay/src/index.tsx
@@ -117,16 +117,17 @@ const RouteViewerOverlay = (props: Props): JSX.Element => {
 
     patterns.forEach(ptn => {
       ptn.stops?.forEach(stop => {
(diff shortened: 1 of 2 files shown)


packages/route-viewer-overlay/src/index.tsx
```diff
@@ -117,16 +117,17 @@ const RouteViewerOverlay = (props: Props): JSX.Element => {
 
     patterns.forEach(ptn => {
       ptn.stops?.forEach(stop => {
-        if (stop.geometries?.geoJson?.type === "Polygon") {
```
miles-grant-ibigroup, line 120: Could this not have been resolved by just adding a question mark here? I guess it's cleaner now...

miles-grant-ibigroup (approved):
Thanks for catching this! Fix works for me.

amy-corson-ibigroup (approved):

PR #503, opened 2022-12-05
Add stop code next to stop names in printed narratives

binh-dam-ibigroup:
Intended to replace #457. May be blocked/enhanced by #504.

Will trigger a fix release on printable itineraries.
For itinerary-body, will not trigger a fix release at this time (there is no change in behavior).


amy-corson-ibigroup (approved):
Looks good to me!

daniel-heppner-ibigroup (approved):
Looks good, thanks for the changes.

PR #504, opened 2022-12-05
fix(core-utils): Extract utility method to compute a displayed stop id.

binh-dam-ibigroup:
New method `itinerary.getDisplayedStopId` in core-utils, intended to be used in itinerary, printable itinerary packages, and in OTP-RR.


daniel-heppner-ibigroup (approved):
The tests make reviewing easy, thanks

PR #507, opened 2022-12-06
fix(route-viewer-overlay): Use timeout before fitting bounds

binh-dam-ibigroup:
Bounds fitting on the route viewer overlay still chokes if there is a large number of features rendered on the map, so in this PR, I am using a timeout before bounds fitting is triggered.

### DIFF
--- a/packages/route-viewer-overlay/src/index.tsx
+++ b/packages/route-viewer-overlay/src/index.tsx
@@ -137,8 +137,8 @@ const RouteViewerOverlay = (props: Props): JSX.Element => {
     });
 
     if (bounds && current) {
-      // Use a timeout to wait for other map rendering to finish
-      // before fitting to the route or pattern shape.
+      // Try to fit the map to route bounds immediately. If other overlays are still populating contents
+      // and/or the map skips/aborts fitting for any reason, try fitting bounds again after a short delay.
       const fitBounds = () => util.fitMapBounds(current, bounds);
       fitBounds();
       timeout = setTimeout(fitBounds, 250);
--- a/packages/route-viewer-overlay/src/index.tsx
+++ b/packages/route-viewer-overlay/src/index.tsx
@@ -139,7 +139,9 @@ const RouteViewerOverlay = (props: Props): JSX.Element => {
     if (bounds && current) {
       // Use a timeout to wait for other map rendering to finish
       // before fitting to the route or pattern shape.
-      timeout = setTimeout(() => util.fitMapBounds(current, bounds), 500);
+      const fitBounds = () => util.fitMapBounds(current, bounds);
+      fitBounds();
+      timeout = setTimeout(fitBounds, 250);
       if (props.mapCenterCallback) {
         props.mapCenterCallback();
       }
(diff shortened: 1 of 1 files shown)


packages/route-viewer-overlay/src/index.tsx
```diff
@@ -136,11 +137,16 @@ const RouteViewerOverlay = (props: Props): JSX.Element => {
     });
 
     if (bounds && current) {
-      util.fitMapBounds(current, bounds);
+      // Use a timeout to wait for other map rendering to finish
+      // before fitting to the route or pattern shape.
+      timeout = setTimeout(() => util.fitMapBounds(current, bounds), 500);
```
miles-grant-ibigroup, line 142: I've noticed this happening but I'm not sure adding a delay is the best UX. Could we possibly do one `fitBounds` right away and then a second one after the delay?

miles-grant-ibigroup (changes requested):
I'm not crazy about the delay, but think I may have an idea how to solve the problem. 

We can even try to measure if the map moved and if it did we can cancel the timeout? (to avoid the annoyance of dragging the map and it snapping back)

The other alternative of course is to adjust the `useEffect` triggers to fire when the geometry changes and that way the entire method will fire again when the data is fully downloaded -- no timeout needed.

packages/route-viewer-overlay/src/index.tsx
binh-dam-ibigroup: A shorter delay of 250 ms seemed to work if the concern is that someone could pan the map before the timeout expires.

binh-dam-ibigroup: > The other alternative of course is to adjust the useEffect triggers to fire when the geometry changes and that way the entire method will fire again when the data is fully downloaded -- no timeout needed.

That is mostly the existing implementation (except that we don't compute a value or hash for the entire geometry which can be a large chunk of data).

miles-grant-ibigroup: Do you think the hash idea is reasonable? Would prefer that to any timeouts

binh-dam-ibigroup: With 250ms, do we still need to have an extra trigger right-away?

miles-grant-ibigroup: I think so. 250ms is still a noticeable delay

binh-dam-ibigroup: Changes per above added in https://github.com/opentripplanner/otp-ui/pull/507/commits/984899b77a674839bf614a0f57a73e67bb77218a.

miles-grant-ibigroup (approved):
Works great! Thanks for the fix

packages/route-viewer-overlay/src/index.tsx
```diff
@@ -136,11 +137,18 @@ const RouteViewerOverlay = (props: Props): JSX.Element => {
     });
 
     if (bounds && current) {
-      util.fitMapBounds(current, bounds);
+      // Use a timeout to wait for other map rendering to finish
+      // before fitting to the route or pattern shape.
```
daniel-heppner-ibigroup, line 141: I think this comment could be adjusted to explain why the function is being called twice. (reducing perceived latency)

daniel-heppner-ibigroup (approved):
LGTM. Just maybe update the comment from the changes earlier in the PR review.

PR #510, opened 2022-12-11
Fix Reverse Geocoding (follow up #487)

amenk:
Reverse Geocoding with Photon does not show proper results when used with otp-react-redux (still coordinates are shown).

This should fix this issue - the returned point contains a field named "lon" not "lng", so lon was undefined here.

### DIFF
--- a/packages/geocoder/src/geocoders/photon.ts
+++ b/packages/geocoder/src/geocoders/photon.ts
@@ -82,8 +82,7 @@ export default class PhotonGeocoder extends Geocoder {
       return response as MultiGeocoderResponse;
     }
 
-    const lat = response.point.lat;
-    const lon = response.point.lng;
+    const { lat, lon } = response.point;
 
     const firstFeature = response.features[0];
     return {


miles-grant-ibigroup (approved):
Thanks for fixing this, works for me!

binh-dam-ibigroup (approved):
LGTM, but please add a commitizen `fix` commit so that a fix release of this package can be created.

amenk:
Thanks - 

I tried 

```
git reset --soft HEAD^1
git cz
```

and was expecting it allows me to select the commit type like here:

https://github.com/commitizen/cz-cli#using-the-command-line-tool

But just the commit message editor opened.

Is there some more documentation about the commitzen workflow in this repo?

Or is it enough to prefix the commit message with "fix: " ?

binh-dam-ibigroup:
> Or is it enough to prefix the commit message with "fix: " ?

That should be fine. It's okay to make a trivial code change before adding the fix commit to this PR.

amenk:
@binh-dam-ibigroup I amended the commit message & force-pushed.

PR #511, opened 2022-12-13
feature(itinerary-body): Add shouldAnimate param on all use of AnimateHeight

noletorious:
The intent behind conditionally animating collapsing content is to help address accessibility concerns around when the user wants to toggle motion off via their operating systems settings. More information on [prefers-reduce-motion](https://web.dev/prefers-reduced-motion/).

### DIFF
--- a/packages/itinerary-body/src/TransitLegBody/index.tsx
+++ b/packages/itinerary-body/src/TransitLegBody/index.tsx
@@ -44,6 +44,7 @@ interface Props {
   setViewedTrip: SetViewedTripFunction;
   showAgencyInfo: boolean;
   showViewTripButton: boolean;
+  shouldAnimate: boolean;
   timeZone: string;
   TransitLegSubheader?: FunctionComponent<TransitLegSubheaderProps>;
   TransitLegSummary: FunctionComponent<TransitLegSummaryProps>;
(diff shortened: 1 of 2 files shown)


miles-grant-ibigroup (changes requested):
Thanks for this change, I agree that we should respect `prefers-reduce-motion`. Do you think it's possible to do this inside of this component instead of exposing a new prop?

packages/itinerary-body/src/TransitLegBody/index.tsx
```diff
@@ -287,7 +292,10 @@ class TransitLegBody extends Component<Props, State> {
                 )}
               </S.TransitLegDetailsHeader>
               {/* IntermediateStops expanded body */}
-              <AnimateHeight duration={500} height={stopsExpanded ? "auto" : 0}>
+              <AnimateHeight
+                duration={shouldAnimate ? 500 : 0}
```
grant-humphries, line 296: since you're doing this check multiple times, you could assign the value to variable outside the mark-up and then re-use as a prop for the instances of `AnimateHeight`, something like:

```js
const animationDuration = shouldAnimate ? 500 : 0;

<AnimateHeight duration={animationDuration} />
```

PR #512, opened 2022-12-13
fix(SSR): isInputTypeSupported() and use of 'document' breaks TORA build

fpurcell:
this hides a use of 'document' from our SSR / yarn build, which otherwise complains about the use of an unknown (server-side) variable.

### DIFF
--- a/packages/trip-form/src/DateTimeSelector/index.tsx
+++ b/packages/trip-form/src/DateTimeSelector/index.tsx
@@ -90,14 +90,10 @@ interface DepartArriveOption {
  */
 function isInputTypeSupported(type: string): boolean {
   let retVal = false;
-  try {
-    if (typeof document !== "undefined") {
-      const input = document.createElement("input");
-      input.setAttribute("type", type);
-      retVal = input.type === type;
-    }
-  } catch {
-    retVal = false;
+  if (typeof document !== "undefined") {
+    const input = document.createElement("input");
+    input.setAttribute("type", type);
+    retVal = input.type === type;
   }
   return retVal;
 }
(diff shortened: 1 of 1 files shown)


packages/trip-form/src/DateTimeSelector/index.tsx
```diff
@@ -89,9 +89,17 @@ interface DepartArriveOption {
  * @param {*} type One of the HTML5 input types.
  */
 function isInputTypeSupported(type: string): boolean {
-  const input = document.createElement("input");
-  input.setAttribute("type", type);
-  return input.type === type;
+  let retVal = false;
+  try {
```
binh-dam-ibigroup, line 93: If `typeof document !== "undefined"` is the differentiator between server-side rendering and browser rendering, then is the try/catch block necessary?

fpurcell: You're right Binh .. it was a bit extra, and latest push has removed it.

binh-dam-ibigroup: Cool! With the latest changes, you should be able to replace `retVal` with return statements.

binh-dam-ibigroup (approved):
LGTM. Inlining `retVal` would be nice.

packages/trip-form/src/DateTimeSelector/index.tsx
```diff
@@ -89,9 +89,13 @@ interface DepartArriveOption {
  * @param {*} type One of the HTML5 input types.
  */
 function isInputTypeSupported(type: string): boolean {
-  const input = document.createElement("input");
-  input.setAttribute("type", type);
-  return input.type === type;
+  let retVal = false;
+  if (typeof document !== "undefined") {
```
binh-dam-ibigroup, line 93: @fpurcell Could you also please insert a comment explaining that the check is for server-side rendering.

miles-grant-ibigroup (approved):
Looks good to me. Glad the fix is this easy!

PR #514, opened 2022-12-15
feat(itinerary-body): add support for displaying only route_headsign

daniel-heppner-ibigroup:
This PR is mostly prop drilling for a small change that allows the trip_headsign to be displayed on its own by default in itinerary-body. With this enabled, the itinerary body will match what is displayed on the bus's headsign, unless trip_headsign is blank, in which case only the route_long_name will be displayed (same as before).

### DIFF
--- a/packages/itinerary-body/src/ItineraryBody/index.tsx
+++ b/packages/itinerary-body/src/ItineraryBody/index.tsx
@@ -25,6 +25,7 @@ const ItineraryBody = ({
   LineColumnContent,
   mapillaryCallback,
   mapillaryKey,
+  onlyShowTripHeadsign,
   PlaceName,
   RouteDescription,
   routingType = "ITINERARY",
@@ -78,6 +79,7 @@ const ItineraryBody = ({
           LineColumnContent={LineColumnContent}
           mapillaryCallback={mapillaryCallback}
           mapillaryKey={mapillaryKey}
+          onlyShowTripHeadsign={onlyShowTripHeadsign}
           PlaceName={PlaceName}
           RouteDescription={RouteDescription}
           routingType={routingType}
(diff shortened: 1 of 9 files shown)


amy-corson-ibigroup (approved):

miles-grant-ibigroup (changes requested):
This is a lot a lot of prop drilling... is there any way to reach the desired effect using some sort of heuristic? Or better yet is there a way override the fields at a higher level?

daniel-heppner-ibigroup:
> This is a lot a lot of prop drilling... is there any way to reach the desired effect using some sort of heuristic? Or better yet is there a way override the fields at a higher level?

Hm. This change was a request from Community Transit to show only the headsign if it exists (otherwise display only route_long_name). I am not sure specifically which components they're using. If they are only using the `RouteLongName` component, then maybe all the drilling isn't necessary since it would only need to be supported at that level. I don't believe that to be the case, though. Therefore, the only two alternatives I see are:
- simply rewrite the leg to remove the `route_long_name` if you only want the headsign. The issue is that then if there is no headsign, you won't have anything to display. But I guess you could include some logic to only remove that property if the headsign exists. That is starting to feel quite complicated, though. We could make another util function that handles that, which Community Transit could pass their legs through before it arrives at whatever OTP-UI components they're using. 
- just remove support for displaying both headsign and route_long_name for everyone, making it default to headsign only. I don't think this would be a viable path since it would affect everyone.

miles-grant-ibigroup:
Well the underlying issue we're addressing here is the same thing appearing twice right? In that the headsign and route long name contain the same thing. I think it's safe to assume that this is an issue across multiple gtfs feeds, which is why I was thinking of doing some sort of filtering inside the render method here! Maybe some kind of percentage overlap? If the strings are 80% the same or above only display the longer one? Or something like that

What do you think?

PR #516, opened 2022-12-16
fix(trip-form): Give a default selection to all TripOptions categories

grant-humphries:
A bug exists in the `TripOptions` component for which if there is a category that has mixed modes (e.g. an option with mode "CAR_HAIL" and an option with mode "CAR_PARK") and the first option has non-null company attribute value the category is not given a default selection (i.e. there is no selection on the category when it is initially opened).  The change made to `ModeRow` resolves this issue.

I've also added a story to with mode data containing a category that reproduces this bug (when the fix is not applied) and did some clean up to TripOptions stories by creating and re-using a template and moving some default config into the default export

### DIFF
--- a/packages/trip-form/src/TripOptions/ModeRow.tsx
+++ b/packages/trip-form/src/TripOptions/ModeRow.tsx
@@ -108,10 +108,15 @@ const ModeRow = ({
           }
 
           let mode = getCategoryPrimaryMode(category);
+
+          // If the category has a single mode select companies from all
+          // options, if the category has mixed modes select the company
+          // (if any) associated with first option
           const companies =
-            typeof category.mode === "undefined"
-              ? ""
-              : category.options?.map(o => o.company).join(",") || "";
+            (category.mode
+              ? category.options?.map(o => o.company).join(",")
+              : category.options?.[0]?.company) || "";
+
           if (category.type === "access") {
             mode = isChecked
               ? selectedTransitString
--- a/packages/trip-form/src/__mocks__/modes-en.ts
+++ b/packages/trip-form/src/__mocks__/modes-en.ts
@@ -1,3 +1,5 @@
+import cloneDeep from "lodash.clonedeep";
+
 import bikeImage from "./static/bike.svg";
 import carImage from "./static/car.svg";
 import scooterImage from "./static/scooter.svg";
@@ -196,25 +198,17 @@ const commonModes = {
 
 // This variant of the data adds an option with a company as the first
 // option in the "CAR" category that has mixed modes ("CAR_PARK", "CAR_HAIL")
-export const modesWithCompanyFirstMixedCategory = {
-  ...commonModes,
-  categories: commonModes.categories.map(c => {
-    const nextCategory = { ...c };
-
-    if (c.id === "CAR") {
-      nextCategory.options = [
-        {
-          company: "LYFT",
-          mode: "CAR_HAIL",
-          label: "Lyft",
-          url: "https://www.lyft.com"
-        },
-        ...nextCategory.options
-      ];
-    }
-
-    return nextCategory;
-  })
-};
+const modesWithCompanyFirstMixedCategory = cloneDeep(commonModes);
+modesWithCompanyFirstMixedCategory.categories.forEach(c => {
+  if (c.id === "CAR") {
+    c.options.unshift({
+      company: "LYFT",
+      mode: "CAR_HAIL",
+      label: "Lyft",
+      url: "https://www.lyft.com"
+    });
+  }
+});
 
 export default commonModes;
+export { modesWithCompanyFirstMixedCategory };
--- a/packages/trip-form/src/__mocks__/modes-en.ts
+++ b/packages/trip-form/src/__mocks__/modes-en.ts
@@ -194,4 +194,27 @@ const commonModes = {
   ]
 };
 
+// This variant of the data adds an option with a company as the first
+// option in the "CAR" category that has mixed modes ("CAR_PARK", "CAR_HAIL")
+export const modesWithCompanyFirstMixedCategory = {
+  ...commonModes,
+  categories: commonModes.categories.map(c => {
+    const nextCategory = { ...c };
+
+    if (c.id === "CAR") {
+      nextCategory.options = [
+        {
+          company: "LYFT",
+          mode: "CAR_HAIL",
+          label: "Lyft",
+          url: "https://www.lyft.com"
+        },
+        ...nextCategory.options
+      ];
+    }
+
+    return nextCategory;
+  })
+};
+
 export default commonModes;
--- a/packages/trip-form/src/__mocks__/modes-en.ts
+++ b/packages/trip-form/src/__mocks__/modes-en.ts
@@ -199,16 +199,14 @@ const commonModes = {
 // This variant of the data adds an option with a company as the first
 // option in the "CAR" category that has mixed modes ("CAR_PARK", "CAR_HAIL")
 const modesWithCompanyFirstMixedCategory = cloneDeep(commonModes);
-modesWithCompanyFirstMixedCategory.categories.forEach(c => {
-  if (c.id === "CAR") {
-    c.options.unshift({
-      company: "LYFT",
-      mode: "CAR_HAIL",
-      label: "Lyft",
-      url: "https://www.lyft.com"
-    });
-  }
-});
+modesWithCompanyFirstMixedCategory.categories
+  .find(c => c.id === "CAR")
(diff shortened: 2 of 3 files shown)


packages/trip-form/src/__mocks__/modes-en.ts
```diff
@@ -194,4 +194,27 @@ const commonModes = {
...
+  categories: commonModes.categories.map(c => {
+    const nextCategory = { ...c };
+
+    if (c.id === "CAR") {
+      nextCategory.options = [
+        {
+          company: "LYFT",
+          mode: "CAR_HAIL",
+          label: "Lyft",
+          url: "https://www.lyft.com"
```
binh-dam-ibigroup, line 210: Is `url` important for this PR? If so, add an optional `url` to type `Category` in `src/TripOptions/types.d.ts`, otherwise delete this `url` attribute.

packages/trip-form/src/__mocks__/modes-en.ts
```diff
@@ -194,4 +194,27 @@ const commonModes = {
   ]
 };
 
+// This variant of the data adds an option with a company as the first
+// option in the "CAR" category that has mixed modes ("CAR_PARK", "CAR_HAIL")
+export const modesWithCompanyFirstMixedCategory = {
+  ...commonModes,
+  categories: commonModes.categories.map(c => {
+    const nextCategory = { ...c };
```
binh-dam-ibigroup, line 202: I think using a cloning method (e.g. `lodash.cloneDeep` used elsewhere in this repo) would be better and would better manipulate story-specific data. (see other comment about array below).

packages/trip-form/src/__mocks__/modes-en.ts
```diff
@@ -194,4 +194,27 @@ const commonModes = {
...
+    if (c.id === "CAR") {
+      nextCategory.options = [
+        {
+          company: "LYFT",
+          mode: "CAR_HAIL",
+          label: "Lyft",
+          url: "https://www.lyft.com"
+        },
+        ...nextCategory.options
+      ];
```
binh-dam-ibigroup, line 213: It looks like array spreading does not play well with TypeScript... You could try (assuming `nextCategory` is a deep clone of argument `c`):
```js
      nextCategory.options.splice(0, 0, {
          company: "LYFT",
          label: "Lyft",
          mode: "CAR_HAIL",
          url: "https://www.lyft.com"
      });

```

packages/trip-form/src/__mocks__/modes-en.ts
binh-dam-ibigroup: Move the cloning/spreading inside the if block, so that if the car condition is not met, just return the original `c` object.

binh-dam-ibigroup (changes requested):
@grant-humphries Please consider comments based on code before you added the exceptions to type checking. I think enforcing types even on stories helps keeping the mock data and the various data types in sync.

packages/trip-form/src/__mocks__/modes-en.ts
grant-humphries: > Please consider comments based on code before you added the exceptions to type checking. I think enforcing types even on stories helps keeping the mock data and the various data types in sync.

Hi @binh-dam-ibigroup, thank you for reviewing this, I hadn't seen your comments at the time I adjusted the typescript config and I'm glad to revert that change if you think that's the way to go, but let me explain a little further what I've found:

With respect to this comment, you're looking at the wrong type definition, it's `ModeOption` that being validated against, not `Category` and the former already has an [optional `url` property](https://github.com/opentripplanner/otp-ui/blob/master/packages/trip-form/src/TripOptions/types.d.ts#L12).  The problem is that the typescript checks seem to be applying the types from `trip-form/src/types.ts` (which has similar, but not identical types defined) instead of `trip-form/src/TripOptions/types.d.ts` and this mock data is designed to work with the latter.  I'm a bit confused about how typescript is matching the mock data to types, its seems to kind of be making sort of an educated guess, but that doesn't work out in this instance.

Also, when the mock data is declared explicitly like `commonModes` [here](https://github.com/opentripplanner/otp-ui/blob/master/packages/trip-form/src/__mocks__/modes-en.ts#L6)  typescript doesn't seem to be doing any checks on it, if it were [this](https://github.com/opentripplanner/otp-ui/blob/master/packages/trip-form/src/__mocks__/modes-en.ts#L64) options section should raise the same violations as the new mock object, as it's in the same format as the Lyft option that I add in the variant.  It seems that type checking is only being done on the mocks when properties are *re*assigned

packages/trip-form/src/__mocks__/modes-en.ts
grant-humphries: Good call on the deep clone, will make that adjustment, the way I've set this up I'm re-using some parts of the `commonModes` tree and cloning some parts so probably best to disassociate them and clone the whole thing as you suggest.  Will look at adjusting the content of the `if` block as well

packages/trip-form/src/__mocks__/modes-en.ts
grant-humphries: typescript raises the same error when I switch to the `splice` syntax, I think the problem is that it's validating the option object against [`FullModeOption`](https://github.com/opentripplanner/otp-ui/blob/master/packages/trip-form/src/types.ts#L25) (from `src/types.ts`) instead of [`ModeOption`](https://github.com/opentripplanner/otp-ui/blob/master/packages/trip-form/src/TripOptions/types.d.ts#L3) (from `src/TripOptions/types.d.ts`).  Do you know of a way to control the association from the data to the type in this context?

packages/trip-form/src/__mocks__/modes-en.ts
grant-humphries: Adjusted the creation of the new mock data to use `cloneDeep` and adjusted the `if` condition to work better with that approach

grant-humphries:
@binh-dam-ibigroup it looks like the conversation around whether it makes sense to type check the mocks is a moot point now.  Since I switched to using `cloneDeep` to create the mock data variant typescript doesn't complain about the approach I use to inject the Lyft category option. I removed the change from the ts config that was ignoring mocks in type checking and the pipeline is passing without it

binh-dam-ibigroup:
> @binh-dam-ibigroup it looks like the conversation around whether it makes sense to type check the mocks is a moot point now. Since I switched to using `cloneDeep` to create the mock data variant typescript doesn't complain about the approach I use to inject the Lyft category option. I removed the change from the ts config that was ignoring mocks in type checking and the pipeline is passing without it

That's great news!

binh-dam-ibigroup (approved):
Thanks for the changes!

packages/trip-form/src/__mocks__/modes-en.ts
```diff
@@ -194,4 +196,19 @@ const commonModes = {
   ]
 };
 
+// This variant of the data adds an option with a company as the first
+// option in the "CAR" category that has mixed modes ("CAR_PARK", "CAR_HAIL")
+const modesWithCompanyFirstMixedCategory = cloneDeep(commonModes);
+modesWithCompanyFirstMixedCategory.categories.forEach(c => {
```
miles-grant-ibigroup, line 202: A bit of a nitpick, but I'm a little confused as to why we're using `forEach` here instead of `find` or similar. I guess this might be fewer lines of code so I think a comment at the very least explaining why `forEach` is used would be helpful!

packages/trip-form/src/TripOptions/ModeRow.tsx
```diff
@@ -108,10 +108,15 @@ const ModeRow = ({
...
+          // If the category has a single mode select companies from all
+          // options, if the category has mixed modes select the company
+          // (if any) associated with first option
           const companies =
-            typeof category.mode === "undefined"
-              ? ""
-              : category.options?.map(o => o.company).join(",") || "";
+            (category.mode
+              ? category.options?.map(o => o.company).join(",")
+              : category.options?.[0]?.company) || "";
```
miles-grant-ibigroup, line 118: This is really clean thanks for this!

miles-grant-ibigroup (approved):
This looks great! Thanks for doing fiddly work in this component. I have one quick confusion about what's going on in the `modes-en` file, but nothing blocking!

packages/trip-form/src/__mocks__/modes-en.ts
grant-humphries: Ah, yes `find` is better, will change

grant-humphries: fixed!

PR #517, opened 2022-12-19
fix(core-utils): adjust baseline co2 calculation algorithm

miles-grant-ibigroup:
The current algorithm provides results that are too far from the "actual" ones based on actual leg length. This PR attempts to improve this by adding a multiplier to the "crow-flight" distance.

### DIFF
--- a/packages/core-utils/src/itinerary.ts
+++ b/packages/core-utils/src/itinerary.ts
@@ -521,8 +521,10 @@ export function calculateEmissions(
   const totalCarbon =
     itinerary?.legs?.reduce((total, leg) => {
       return (
-        (leg.distance * carbonIntensityWithDefaults[leg.mode.toLowerCase()] ||
-          0) + total
+        // We estimate that the actual driving distance is 1.34x the "crow-flying" distance
+        (leg.distance *
+          1.34 *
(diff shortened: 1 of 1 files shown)


packages/core-utils/src/itinerary.ts
```diff
@@ -521,8 +521,10 @@ export function calculateEmissions(
   const totalCarbon =
     itinerary?.legs?.reduce((total, leg) => {
       return (
-        (leg.distance * carbonIntensityWithDefaults[leg.mode.toLowerCase()] ||
-          0) + total
+        // We estimate that the actual driving distance is 1.34x the "crow-flying" distance
+        (leg.distance *
+          1.34 *
```
daniel-heppner-ibigroup, line 526: should this maybe be a const at the top of the file? Please at least add a comment explaining where it comes from

daniel-heppner-ibigroup (changes requested):
1.34???

packages/core-utils/src/itinerary.ts
miles-grant-ibigroup: I'm weary of making it a const since I pretty much just made it up. It feels kinda right based on how "off" the values were

daniel-heppner-ibigroup (approved):

philip-cline (approved):
1.34!

PR #518, opened 2022-12-21
OTP2 Tile Overlay Prerequisites 

miles-grant-ibigroup:
Clone of #472 without the new tile layer itself. This allows #472 to use correct non-alpha package versions.

### DIFF
--- a/packages/map-popup/package.json
+++ b/packages/map-popup/package.json
@@ -1,0 +1,33 @@
+{
+  "name": "@opentripplanner/map-popup",
+  "version": "0.0.1",
+  "description": "A component for displaying map popup contents",
+  "main": "lib/index.js",
+  "module": "esm/index.js",
+  "types": "lib/index.d.js",
+  "repository": "https://github.com/opentripplanner/otp-ui.git",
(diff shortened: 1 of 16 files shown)


packages/map-popup/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "module": "esm/index.js",
+  "types": "lib/index.d.js",
+  "repository": "https://github.com/opentripplanner/otp-ui.git",
+  "homepage": "https://github.com/opentripplanner/otp-ui#readme",
+  "author": "Evan Siroky",
+  "license": "MIT",
+  "private": false,
+  "dependencies": {
+    "@opentripplanner/base-map": "^3.0.6",
+    "@opentripplanner/core-utils": "^7.0.9",
```
binh-dam-ibigroup, line 15: Do you want to use core-utils 8.0.0?

binh-dam-ibigroup (approved):

packages/map-popup/package.json
miles-grant-ibigroup: Good call!

amy-corson-ibigroup (approved):

PR #519, opened 2022-12-27
prioritize config transit operator over gtfs agency

amy-corson-ibigroup:
In tandem with https://github.com/opentripplanner/otp-react-redux/pull/734, this prioritizes using the Transit Operators listed in the configurations file over the operators in the GTFS route.


demory:
@amy-corson-ibigroup I think I figured out why I was having problems testing this alongside otp-rr -- the dev branch is off a very old copy of the otp-ui code, and the file you edited (`itinerary-body/src/index.js`) no longer exists in otp-ui (the module has since been typescriptified).

When I tried to copy this into my otp-rr setup I was getting very odd errors like it not being able to find moment and other old dependencies.

Are you able to fast-forward the branch and migrate your edits to the TS code?

demory:
Looks like the operator name isn't making it into the alt text:
image

packages/itinerary-body/src/TransitLegBody/index.tsx
```diff
@@ -214,7 +222,7 @@ class TransitLegBody extends Component<Props, State> {
                               id: "otpUi.TransitLegBody.agencyLogo"
                             },
                             {
-                              agencyName
+                              transitOperatorName
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Sorry about that! Fixed in 06e599fe838792c08763a5f259ab284474af5599

demory:
Now seeing some broken examples for the Itinerary body in storybook, can you confirm if you're seeing as well:

image

demory (approved):

packages/itinerary-body/src/TransitLegBody/index.tsx
```diff
@@ -157,6 +157,13 @@ class TransitLegBody extends Component<Props, State> {
       leg
     );
 
+    // If the config contains an operator name, prefer that one over the
+    // one provided by OTP
+
+    const transitOperatorName = transitOperator?.name
+      ? transitOperator.name
+      : agencyName;
+
```
miles-grant-ibigroup, line 166:
```suggestion
    const transitOperatorName = transitOperator?.name || agencyName;
```

Does this work?

miles-grant-ibigroup (approved):
Wow was the prop we needed there but not being used! Well done on getting the fix to be this simple! Just one small style nitpick.

packages/itinerary-body/src/TransitLegBody/index.tsx
amy-corson-ibigroup: Yes I think so! Thank you!

PR #520, opened 2022-12-30
Various accessibility improvements

miles-grant-ibigroup:
This PR brings with it various small accessibility improvements. Highlights include:

* Location field syntax ordering now makes logical sense and flows the way a user would expect
* Location Icon is localized with sensible strings

Will be merged into https://github.com/opentripplanner/otp-react-redux/pull/742


packages/location-icon/src/index.tsx
```diff
@@ -37,29 +40,32 @@ export function LocationIcon({
         <S.FromIcon
           className={className}
           size={size}
-          title={title || "From Location Icon"}
+          title={title || intl.formatMessage({ id: "otpUi.LocationIcon.from" })}
```
binh-dam-ibigroup, line 43: I think a better approach would be to have this icon not have any alt text or title for these reasons, and that the alt text should be determined by the components that use these icons:
- "From"/"To" text is taken out of context, and other languages could use other words (e.g. origin, destination).
- The title text is not shown to screen readers on the LocationField component (although a "From"/"To" tooltip is shown on hovering the button - that tooltip could be more descriptive instead)
- The title prop generates unnecessary extra text for the icon components that are placed on the map, and you'd have.

packages/location-field/src/index.tsx
```diff
@@ -740,11 +740,10 @@ const LocationField = ({
           onToggle={onDropdownToggle}
           open={menuVisible}
           title={<LocationIconComponent locationType={locationType} />}
+          input={<>{textControl}{clearButton}</>}
```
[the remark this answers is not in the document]
binh-dam-ibigroup: I think the original markup was better because it separated the contents of the dropdown from the textbox portion of the component.

packages/location-icon/src/index.tsx
binh-dam-ibigroup: For the "generic" or "intermediate" location icon, it looks like the description text we provide by default does not match the actual use with intermediate locations.

binh-dam-ibigroup (changes requested):
I think the location icon package should not carry any text (see various reasons why).

packages/location-icon/src/index.tsx
miles-grant-ibigroup: I agree with you -- for the record, this is a departure from how it used to work, right?

packages/location-field/src/index.tsx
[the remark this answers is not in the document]
miles-grant-ibigroup: I agree this new method is annoying. However, it is the only way to have the dropdown contents render after the input field.

packages/location-icon/src/index.tsx
binh-dam-ibigroup: > for the record, this is a departure from how it used to work, right?

Correct.

packages/location-field/src/index.tsx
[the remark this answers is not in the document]
binh-dam-ibigroup: From the original markup, can you move the dropdown after the text and clear button, and use CSS to make show to the left?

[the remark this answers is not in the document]
miles-grant-ibigroup: That would be possible but I think would be much messier than this solution

packages/location-field/src/index.tsx
```diff
@@ -735,6 +735,7 @@ const LocationField = ({
       <S.InputGroup>
         {/* location field icon -- also serves as dropdown anchor */}
         <S.Dropdown
+          input={<>{textControl}{clearButton}</>}
```
binh-dam-ibigroup, line 738: Does that also apply to the mobile view?

packages/location-field/src/styled.tsx
```diff
@@ -74,6 +76,7 @@ export const Dropdown = ({
       <DropdownButton aria-label={dropdownButtonAriaLabel} onClick={onToggle}>
         {title}
       </DropdownButton>
+      {input}
       {open && (
         <MenuItemList uniqueId={listBoxIdentifier}>{children}</MenuItemList>
```
binh-dam-ibigroup, line 81: Internationalize the `arial-label` prop, for English change it to ~"Search Results"~ "Suggested locations".

binh-dam-ibigroup: While we are at this component, rename `uniqueId` to `id` and don't pass props in its styled-component definition.

binh-dam-ibigroup (changes requested):
Some more changes, and also for component `OptionSubtitle`, replace `margin-left` with `&::before { content: " "; }`, so that the subtitle (addresses) read out separate from the option name.

packages/location-icon/src/index.tsx
```diff
@@ -28,34 +22,15 @@ type Props = {
...
   switch (type) {
     case "from":
-      return (
-        <S.FromIcon
-          className={className}
-          size={size}
-          title={title || "From Location Icon"}
-        />
-      );
+      return <S.FromIcon className={className} size={size} />;
```
binh-dam-ibigroup, line 29: Yes, this is much cleaner.

packages/location-field/src/index.tsx
miles-grant-ibigroup: The mobile view has always had the dropdown in the semantically correct place!

packages/location-field/src/styled.tsx
miles-grant-ibigroup: We can't use `id` because it clashes with internal React props (id being an HTML reserved keyword).

binh-dam-ibigroup: We use an `id` prop with the map, overlays, i18n etc.

miles-grant-ibigroup: Oh good call... Totally slept through this one

packages/location-field/src/styled.tsx
```diff
@@ -74,6 +76,7 @@ export const Dropdown = ({
       <DropdownButton aria-label={dropdownButtonAriaLabel} onClick={onToggle}>
```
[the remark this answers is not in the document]
miles-grant-ibigroup: AXE disagrees with you. I'm tempted to disable the rule -- what do you think?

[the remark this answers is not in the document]
binh-dam-ibigroup: Okay. Referring to [this example](https://w3c.github.io/aria-practices/examples/combobox/combobox-autocomplete-list.html), the button does carry an `aria-label`, so reinstate the aria-label - sorry!... Also, the button has `aria-expanded` and `aria-controls` attributes, so we can probably follow that too.

packages/location-field/src/styled.tsx
```diff
@@ -26,12 +27,15 @@ export const DropdownContainer = styled.span`
...
 type MenuItemListProps = {
-  uniqueId: string;
+  id: string;
+  intl: IntlShape;
 };
 
 export const MenuItemList = styled.ul.attrs((props: MenuItemListProps) => ({
-  "aria-label": "Search Results",
-  id: props.uniqueId,
+  "aria-label": props.intl.formatMessage({
```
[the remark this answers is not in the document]
miles-grant-ibigroup: You can't do that with styled components, you have to do it like this. I'd prefer it the other way too.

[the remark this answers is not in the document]
binh-dam-ibigroup: I'd much rather keep props out of the styled component, since none of these props affect the style. Bonus, you don't have to pass `intl` around.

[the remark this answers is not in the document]
miles-grant-ibigroup: Me too, but it doesn't work. Applying an aria-label to a styled component doesn't apply it to the resulting html

binh-dam-ibigroup:
Regarding the search results, additional code changes need to happen:
- Convert the `<div>`s within the `<a>`s to `<span>`s.
- Group options by categories, probably using containing elements with `aria-role="group"`, if referring to the implicit role of [`<optgroup>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup) and the [ARIA group role reference](https://w3c.github.io/aria/#group)

packages/location-field/src/index.tsx
```diff
@@ -713,7 +713,12 @@ const LocationField = ({
             {clearButton}
           </S.InputGroup>
         </S.FormGroup>
-        <S.StaticMenuItemList uniqueId={listBoxId}>
+        <S.StaticMenuItemList
+          aria-label={intl.formatMessage({
+            id: "otpUi.LocationField.suggestedLocations"
```
binh-dam-ibigroup, line 718: Add a `defaultMessage` and `description`.

packages/location-field/src/styled.tsx
```diff
@@ -55,27 +50,49 @@ export const MenuItemList = styled.ul.attrs((props: MenuItemListProps) => ({
...
+
   return (
     <DropdownContainer>
-      <DropdownButton aria-label={dropdownButtonAriaLabel} onClick={onToggle}>
+      <DropdownButton
+        aria-controls={listBoxIdentifier}
+        aria-expanded={open}
+        aria-label={intl.formatMessage(
+          {
+            id: "otpUi.LocationField.suggestedLocationsLong"
```
binh-dam-ibigroup, line 77: Add a `defaultMessage` and a `description`.

packages/location-field/src/styled.tsx
```diff
@@ -55,27 +50,49 @@ export const MenuItemList = styled.ul.attrs((props: MenuItemListProps) => ({
...
+        onClick={onToggle}
+      >
         {title}
       </DropdownButton>
+      {input}
       {open && (
-        <MenuItemList uniqueId={listBoxIdentifier}>{children}</MenuItemList>
+        <MenuItemList
+          aria-label={intl.formatMessage({
+            id: "otpUi.LocationField.suggestedLocations"
```
binh-dam-ibigroup, line 90: Add a `defaultMessage` and a `description`.

binh-dam-ibigroup:
@miles-grant-ibigroup Please resolve conflicts on this PR.

packages/location-field/src/index.tsx
```diff
@@ -439,6 +439,7 @@ const LocationField = ({
           bgColor={layerColorMap.stations}
           centeredText
           header
+          role="group"
```
binh-dam-ibigroup, line 442: The element with the `group` role should be a container with the items of the group contained within. If not doing containing elements, an alternate approach is to say `role="heading"` for these items, akin to bootstrap's implementation.

packages/location-field/src/index.tsx
```diff
@@ -465,7 +467,13 @@ const LocationField = ({
       stopFeatures.map(feature => renderFeature(itemIndex++, feature)),
 
       transitFeaturesPresent && otherFeatures.length > 0 && (
-        <S.MenuItem bgColor="#333" centeredText header key="other-header">
+        <S.MenuItem
+          bgColor="#333"
+          centeredText
+          header
+          role="group"
```
binh-dam-ibigroup, line 474: Sort props. See other comment about the role to use.

packages/location-field/src/index.tsx
```diff
@@ -528,7 +536,7 @@ const LocationField = ({
   if (sessionSearches.length > 0) {
     // Add the menu sub-heading (not a selectable item)
     menuItems.push(
-      <S.MenuItem header centeredText key="ss-header">
+      <S.MenuItem header centeredText role="group" key="ss-header">
```
binh-dam-ibigroup, line 539: Sort props

packages/location-field/src/styled.tsx
```diff
@@ -55,27 +50,55 @@ export const MenuItemList = styled.ul.attrs((props: MenuItemListProps) => ({
...
+        aria-label={intl.formatMessage(
+          {
+            id: "otpUi.LocationField.suggestedLocationsLong",
+            defaultMessage: "Open the list of location suggestions",
+            description:
+              "Text to show as a a11y label for the button that opens the dropdown list of locations"
+          },
+          { locationType }
+        )}
+        tabIndex={-1}
```
binh-dam-ibigroup, line 84: Sort props

packages/location-field/src/styled.tsx
```diff
@@ -55,27 +50,55 @@ export const MenuItemList = styled.ul.attrs((props: MenuItemListProps) => ({
...
+
   return (
     <DropdownContainer>
-      <DropdownButton aria-label={dropdownButtonAriaLabel} onClick={onToggle}>
+      <DropdownButton
+        aria-controls={listBoxIdentifier}
+        aria-expanded={open}
+        aria-label={intl.formatMessage(
+          {
+            id: "otpUi.LocationField.suggestedLocationsLong",
```
binh-dam-ibigroup, line 77: Sort attributes (2 instances)

packages/location-field/i18n/en-US.yml
```diff
@@ -26,6 +26,8 @@ otpUi:
     # Note to translator: This is an implicit plural (as in "Other results").
     other: Other
     recentlySearched: Recently Searched
+    suggestedLocations: Suggested locations
+    suggestedLocationsLong: Open the list of {locationType} location suggestions
```
binh-dam-ibigroup, line 30: Let's not use `locationType` in this message, or we would need to add the different values to the languages files.
```suggestion
    suggestedLocationsLong: Open the list of suggested locations
```

binh-dam-ibigroup: Hmm, we might need a "Close list of suggested locations" too...

miles-grant-ibigroup: "toggle"?

binh-dam-ibigroup: Sure

miles-grant-ibigroup:
Thanks for the careful review!

packages/location-field/src/styled.tsx
```diff
@@ -74,26 +74,26 @@ export const Dropdown = ({
...
         aria-label={intl.formatMessage(
           {
-            id: "otpUi.LocationField.suggestedLocationsLong",
             defaultMessage: "Open the list of location suggestions",
             description:
-              "Text to show as a a11y label for the button that opens the dropdown list of locations"
+              "Text to show as a a11y label for the button that opens the dropdown list of locations",
+            id: "otpUi.LocationField.suggestedLocationsLong"
           },
           { locationType }
```
binh-dam-ibigroup, line 82: Passing parameters is no longer needed.

miles-grant-ibigroup: Oh thanks for catching that totally missed that in all the stress!

binh-dam-ibigroup: Remove the `locationType` prop from the `DropDown` component.

packages/location-field/src/index.tsx
```diff
@@ -488,7 +488,12 @@ const LocationField = ({
   if (nearbyStops.length > 0 && !suppressNearby) {
     // Add the menu sub-heading (not a selectable item)
     menuItems.push(
-      <S.MenuItem centeredText header key="ns-header" role="group">
+      <S.MenuItem
+        key="ns-header"
+        role="heading"
+        centeredText
```
binh-dam-ibigroup, line 494: Sort props.

packages/location-field/src/index.tsx
```diff
@@ -536,7 +541,12 @@ const LocationField = ({
   if (sessionSearches.length > 0) {
     // Add the menu sub-heading (not a selectable item)
     menuItems.push(
-      <S.MenuItem header centeredText role="group" key="ss-header">
+      <S.MenuItem
+        key="ss-header"
+        centeredText
```
binh-dam-ibigroup, line 546: Sort props

packages/location-field/src/index.tsx
```diff
@@ -576,7 +586,12 @@ const LocationField = ({
   if (userLocationsAndRecentPlaces.length > 0 && showUserSettings) {
     // Add the menu sub-heading (not a selectable item)
     menuItems.push(
-      <S.MenuItem header centeredText key="mp-header" role="group">
+      <S.MenuItem
+        key="mp-header"
+        role="heading"
+        centeredText
```
binh-dam-ibigroup, line 592: Sort props

binh-dam-ibigroup (approved):
Good to go for me in light of the latest changes, thanks for the quick response. See some remaining formatting remarks.

packages/location-field/src/index.tsx
```diff
@@ -735,16 +765,14 @@ const LocationField = ({
       <S.InputGroup>
         {/* location field icon -- also serves as dropdown anchor */}
         <S.Dropdown
+          input={<>{textControl}{clearButton}</>}
```
binh-dam-ibigroup, line 768: While this PR is open, what would be a good way to add support for the input's `required` attribute? This is in anticipation of form validation. An `isRequired` prop could work.

binh-dam-ibigroup: We would need to support some kind of invalid state too, probably via an `isInvalid` prop that's rendered as an `aria-invalid` attribute.

binh-dam-ibigroup (changes requested):
I'm gonna go ahead and request support for `aria-required` and `aria-invalid` attributes.

binh-dam-ibigroup (approved):
LGTM

packages/location-field/src/index.tsx
```diff
@@ -735,16 +769,14 @@ const LocationField = ({
       <S.InputGroup>
```
amy-corson-ibigroup, line 769: Would it be overkill to add a `role="group"` to this? Looking at the layout in the screenreader view, it could be nice to distinguish which inputs/containers/buttons are related 
image

packages/location-field/src/index.tsx
```diff
@@ -735,16 +769,14 @@ const LocationField = ({
...
         {/* location field icon -- also serves as dropdown anchor */}
         <S.Dropdown
+          input={<>{textControl}{clearButton}</>}
```
amy-corson-ibigroup, line 772: This layout is a bit wonky for me, see image below. Could just be local, but it looks like a `width: 100%` is getting added somewhere to the input, which is clashing with the 30px width on the `DropdownButtonIcon`

image

packages/location-field/src/index.tsx
```diff
@@ -54,7 +54,9 @@ const LocationField = ({
   hideExistingValue = false,
   initialSearchResults = null,
   inputPlaceholder = null,
+  isRequired = false,
   isStatic = false,
+  isValid = true,
```
amy-corson-ibigroup, line 59: Just so I'm clear, is this just setting up the ability to use `aria-invalid` for a future PR, not necessarily in this one?

amy-corson-ibigroup (approved):
LGTM provided that layout issue with the input widths isn't a problem! A few small suggestions and clarifications below

packages/location-field/src/index.tsx
binh-dam-ibigroup: Yes, that will be set by OTP-RR in an upcoming PR.

packages/location-field/src/index.tsx
miles-grant-ibigroup: I think it's a good idea!

packages/location-field/src/index.tsx
miles-grant-ibigroup: Yep there's some otp-ui changes required. There is a PR there that resolves this! Thanks for spotting

PR #521, opened 2023-01-04
Algorithmically determine whether route_long_name should be shown

daniel-heppner-ibigroup:
This PR adds functionality to determine whether route_long_name should be shown next to trip headsign based on the similarity to the trip_headsign. If they are closely overlapping, then the route_long_name is not needed.
For example, "Downtown Seattle / Aurora Village" to "Aurora Village". The rider does not need to know that the bus serves  Downtown Seattle in the other direction, they only need to know what will appear on the bus headsign. 

Meanwhile, in places like San Francisco, where bus routes have names and numbers, the name should not have much overlap with the headsign and therefore it should be shown.

### DIFF
--- a/packages/itinerary-body/package.json
+++ b/packages/itinerary-body/package.json
@@ -19,8 +19,9 @@
     "date-fns": "^2.28.0",
     "date-fns-tz": "^1.2.2",
     "flat": "^5.0.2",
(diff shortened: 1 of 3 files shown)


daniel-heppner-ibigroup:
Assigned to @miles-grant-ibigroup for further development.

PR #523, opened 2023-01-05
refactor(location-field): Remove props argument from MeniItemList sty…

binh-dam-ibigroup:
The changes herein should resolve the comment thread at https://github.com/opentripplanner/otp-ui/pull/520#discussion_r1062515956.


miles-grant-ibigroup (approved):
Not sure why this works all of a sudden but thanks for the changes!

PR #525, opened 2023-01-06
Fix itin styles

binh-dam-ibigroup:
This PR fixes the mis-alignment of items in the transit leg details (see screenshot).
It should maintain the relative position of the elevation chart and the walk/bike summary text.

image

### DIFF
--- a/packages/itinerary-body/src/styled.tsx
+++ b/packages/itinerary-body/src/styled.tsx
@@ -366,11 +366,7 @@ export const LineColumn = styled.div`
   padding-right: 5px;
 `;
 
-export const LegDetails = styled.span`
-  *:not(.fa) {
-    vertical-align: middle;
-  }
-`;
+export const LegDetails = styled.span``;
 
 export const PlaceRowWrapper = styled.div`
   /* needs to be a flexbox row */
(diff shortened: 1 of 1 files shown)


packages/itinerary-body/src/styled.tsx
```diff
@@ -366,11 +366,7 @@ export const LineColumn = styled.div`
   padding-right: 5px;
 `;
 
-export const LegDetails = styled.span`
-  *:not(.fa) {
-    vertical-align: middle;
-  }
-`;
+export const LegDetails = styled.span``;
```
amy-corson-ibigroup, line 369: I can't find the answer to this anywhere, but is it alright to leave this styled component blank?

amy-corson-ibigroup (approved):
Very quick question about best practices but looks great to me!

packages/itinerary-body/src/styled.tsx
binh-dam-ibigroup: It's fine to leave it. Actually, it is an exported symbol, and implementers could apply styles to that symbol by writing `styled(LegDetails)`.

amy-corson-ibigroup: Gotcha, that makes sense! Good to know for the future

daniel-heppner-ibigroup (approved):
Thanks for the fix and simplification!

PR #526, opened 2023-01-10
fix(map-popup): correct stop viewer link behavior

miles-grant-ibigroup:
The map-popup package brought a regression where the stop viewer link wasn't passing the correct stop id. The issue was using a stripped stop ID instead of the entire thing. This PR corrects this. A follow-up PR will be needed to make use of this update in the packages that depend on it.

### DIFF
--- a/packages/map-popup/src/index.tsx
+++ b/packages/map-popup/src/index.tsx
@@ -58,7 +58,7 @@ const StationHubDetails = ({ station }: { station: Station }) => {
   )
 }
 
-const StopDetails = ({ id, setViewedStop }: { id: string, setViewedStop: ({ stopId }: { stopId: string }) => void; }) => {
+const StopDetails = ({ id, setViewedStop }: { id: string, setViewedStop: () => void; }) => {
   return (
     <BaseMapStyled.PopupRow>
         <strong>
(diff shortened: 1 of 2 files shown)


packages/map-popup/src/index.tsx
```diff
@@ -114,7 +114,7 @@ export function MapPopup({ configCompanies, entity, getEntityName, setLocation,
       {entityIsStationHub && <StationHubDetails station={entity} />}
 
       {/* render stop viewer link if available */}
-      {setViewedStop && !bikesAvailablePresent && <StopDetails id={stopId} setViewedStop={setViewedStop} />}
+      {setViewedStop && !bikesAvailablePresent && <StopDetails id={stopId} setViewedStop={() => setViewedStop({ stopId: entity.id })} />}
```
binh-dam-ibigroup, line 117: Use `useCallback` for the event handler.

packages/map-popup/src/index.tsx
```diff
@@ -114,7 +114,7 @@ export function MapPopup({ configCompanies, entity, getEntityName, setLocation,
       {entityIsStationHub && <StationHubDetails station={entity} />}
```
binh-dam-ibigroup, line 115: Line 97 in the prop list: remove trailing comma.

binh-dam-ibigroup (approved):
Thanks for the fix!

daniel-heppner-ibigroup (approved):
Looks good, thanks for the fix!

PR #527, opened 2023-01-11
Hide duplicative long name

daniel-heppner-ibigroup:
This PR hides the duplicative `routeLongName` when it's too similar to the trip headsign.


daniel-heppner-ibigroup:
Oops, I didn't realize I had already done a fix commit.

packages/itinerary-body/src/defaults/route-long-name.tsx
```diff
@@ -40,7 +45,7 @@ export default function RouteLongName({
           }}
         />
       ) : (
-        routeLongName
+        headsign || routeLongName
```
binh-dam-ibigroup, line 48: There is a minor flaw here: if a headsign is shown but not a long name, this is rendering "67 Downtown" whereas previously it would read like "67 [...] *to* Downtown" ("to" is removed). Is this intended?

binh-dam-ibigroup: I think "to" should still be shown in such cases.

daniel-heppner-ibigroup: This was intended, but it is open to debate. I chose to hide the "to" so that the rider sees exactly what will be shown on the vehicle headsign. (assuming the GTFS is accurate) I will add this to the tech call, it's an easy change.

binh-dam-ibigroup (approved):
Whether to keep "to" is up for debate, but otherwise LGTM.

daniel-heppner-ibigroup:
Per tech call, it sounds like we are in agreement on not having the "to".

PR #528, opened 2023-01-11
Improve Itinerary and CO2 Types

daniel-heppner-ibigroup:
This PR adds a new type for mass units, and also a "looser" Itinerary type that makes all fields optional except legs, since in most places that is the only property we need. This will make Typescript happier when we are making "fake" itineraries in OTP-RR and in testing.

### DIFF
--- a/packages/trip-details/src/types.ts
+++ b/packages/trip-details/src/types.ts
@@ -1,6 +1,6 @@
 // Prettier does not recognize the import type syntax.
 // eslint-disable-next-line prettier/prettier
-import type { FareDetails, Money, Itinerary, Leg, Fare, } from "@opentripplanner/types";
+import type { FareDetails, Money, Itinerary, Leg, Fare, MassUnitOption, } from "@opentripplanner/types";
 import type { ReactElement } from "react";
 
 export interface CaloriesDetailsProps {
@@ -11,7 +11,7 @@ export interface CaloriesDetailsProps {
 
 export interface CO2ConfigType {
   carbonIntensity?: { [index: string]: number };
-  units?: string;
+  units?: MassUnitOption;
   enabled?: boolean;
 }
 
--- a/packages/trip-details/src/types.ts
+++ b/packages/trip-details/src/types.ts
@@ -1,6 +1,6 @@
 // Prettier does not recognize the import type syntax.
 // eslint-disable-next-line prettier/prettier
-import type { FareDetails, Money, Itinerary, Leg, Fare, MassUnitOption, } from "@opentripplanner/types";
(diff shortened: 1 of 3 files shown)


packages/types/src/index.ts
```diff
@@ -400,6 +400,13 @@ export type Itinerary = {
   walkTime: number;
 };
 
+/**
+ * In many places all we need from the Itinerary is the legs,
+ * this type makes all the other types optional except legs.
+ */
+export type ItineraryOnlyLegsRequired = Partial<Itinerary> &
```
miles-grant-ibigroup, line 407: super nitpick: can we put this on one line?

miles-grant-ibigroup (changes requested):
Not the hugest fan of the second Itinerary type just for this one use case -- some advanced typescript going on as well. Is there no way to make the itinerary type have optional fields?

daniel-heppner-ibigroup:
What do you mean by one use case? It's used like this everywhere in the itinerary core utils file. I wasn't sure about modifying the Itinerary type to make all the fields permanently optional, it would probably break a lot of other stuff. @miles-grant-ibigroup

miles-grant-ibigroup:
I thought we were making this change to make the tests work with typescript! In that case I'd either rename the new type to make it clear what it's for or try to unify the types.

daniel-heppner-ibigroup:
> I thought we were making this change to make the tests work with typescript! In that case I'd either rename the new type to make it clear what it's for or try to unify the types.

Let me explain what's going on a bit better here; I needed a more relaxed type for the calculateEmissions function, since when there is no driving itinerary in OTP, we generate a fake one that only contains the legs property. Typescript isn't happy about this object not containing any other Itinerary properties, but we don't actually need those. That's what I made this type for. Then I realized that actually, none of these functions require anything besides the legs, so I just applied the type to all of them. 

I can name it something else, but I thought it would be good just name it something super obvious.

miles-grant-ibigroup:
Thanks for that explanation! This all makes sense to me. I don't have a better alternative name right now, so will defer to the second reviewer

packages/trip-details/src/types.ts
```diff
@@ -1,6 +1,6 @@
 // Prettier does not recognize the import type syntax.
 // eslint-disable-next-line prettier/prettier
-import type { FareDetails, Money, Itinerary, Leg, Fare, } from "@opentripplanner/types";
+import type { FareDetails, Money, Itinerary, Leg, Fare, MassUnitOption, } from "@opentripplanner/types";
```
binh-dam-ibigroup, line 3: Sort imports, remove trailing comma

binh-dam-ibigroup (approved):
LGTM

PR #532, opened 2023-01-25
OTP2 FareProduct / Fare By Leg Support

daniel-heppner-ibigroup:
This PR adds support for the new fare by leg structure in OTP2. 
The fundamental thing to understand here is that each fare product used for a leg is stored under legProducts, along with the index of the leg it's associated with. An example itinerary was added to itinerary-body package for the story as well as to show an example of how it looks for review. 
This will be a lot more flexible in the future, and one way I took advantage of the new data is by adding a tooltip to the table cell explaining the transfer discount applied to a given fare. 

I tried to keep it so that in the future it will be easy to remove the OTP1 code. There is only one user of the OTP1 fare by leg, so once they upgrade, we can remove the code for that.

### DIFF
--- a/packages/core-utils/src/itinerary.ts
+++ b/packages/core-utils/src/itinerary.ts
@@ -555,3 +555,46 @@ export function getDisplayedStopId(placeOrStop: Place | Stop): string {
   }
   return stopCode || stopId?.split(":")[1] || stopId;
 }
+
+export function getLegsWithFares(itinerary: Itinerary): Leg[] {
+  return itinerary.legs.map((leg, i) => ({
+    ...leg,
+    legProducts: itinerary.fare.legProducts.filter(lp =>
+      lp.legIndicies.includes(i)
+    )
+  }));
+}
+
+export function getLegCost(
+  leg: Leg,
+  category: string,
+  container: string
+): { cost: Money | undefined; usesTransfer?: boolean } {
+  if (!leg.fareProducts) return { cost: undefined };
+  const relevantFareProducts = leg.fareProducts.filter(
+    fp => fp.category.name === category && fp.container.name === container
+  );
+  const totalCost = relevantFareProducts.find(fp => fp.name === "rideCost")
+    .amount;
+  const usesTransfer = !!relevantFareProducts.find(
+    fp => fp.name === "transfer"
+  );
+  return { cost: totalCost, usesTransfer };
+}
+
+export function getItineraryCost(
+  legs: Leg[],
+  category: string,
+  container: string
+): Money {
+  return legs
+    .filter(leg => !!leg.fareProducts)
+    .map(leg => getLegCost(leg, category, container).cost)
+    .reduce<Money>(
+      (prev, cur) => ({
+        cents: prev.cents + cur.cents,
+        currency: prev.currency ?? cur.currency
+      }),
+      { cents: 0, currency: null }
+    );
+}
--- a/packages/core-utils/src/itinerary.ts
+++ b/packages/core-utils/src/itinerary.ts
@@ -588,7 +588,7 @@ export function getLegCost(
   leg: Leg,
   category: string,
   container: string
-): { price: Money | undefined; transferAmount?: number } {
+): { price?: Money; transferAmount?: number } {
   if (!leg.fareProducts) return { price: undefined };
   const relevantFareProducts = leg.fareProducts.filter(
     fp => fp.category.name === category && fp.container.name === container
@@ -601,7 +601,7 @@ export function getLegCost(
 
   return {
     price: totalCost,
-    transferAmount: transferFareProduct?.amount?.cents || undefined
+    transferAmount: transferFareProduct?.amount?.cents
   };
 }
 
--- a/packages/core-utils/src/itinerary.ts
+++ b/packages/core-utils/src/itinerary.ts
@@ -559,9 +559,9 @@ export function getDisplayedStopId(placeOrStop: Place | Stop): string {
 export function getLegsWithFares(itinerary: Itinerary): Leg[] {
   return itinerary.legs.map((leg, i) => ({
     ...leg,
-    legProducts: itinerary.fare.legProducts.filter(lp =>
-      lp.legIndicies.includes(i)
-    )
+    fareProducts: itinerary.fare?.legProducts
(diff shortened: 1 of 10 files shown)


packages/core-utils/tsconfig.json
```diff
@@ -4,6 +4,7 @@
     "composite": true,
     "outDir": "./lib",
     "rootDir": "./src",
+    "lib": ["es2019", "dom"],
```
daniel-heppner-ibigroup, line 7: Is this allowed?

miles-grant-ibigroup (changes requested):
Well done getting the dual compatibility working like this! How are you planning to do the core-utils/trip-details package updates?

One small note about styling otherwise looks really good!

packages/trip-details/src/fare-table.tsx
```diff
@@ -159,14 +122,47 @@ const FareTypeTable = ({
...
-              const fare = leg.fares[col.key];
+              let fare;
+              if (!hasLegProducts) {
+                fare = leg.fares[col.key];
+              } else {
+                fare = getLegCost(leg, col.riderCategory, col.fareContainer);
+              }
               return (
-                <td key={col.key}>
+                <td
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: That's a good idea, I'll change that.

daniel-heppner-ibigroup:
For the package versions, can we merge this PR and then immediately do an update internal deps PR? This one seems to pass tests somehow, so I think it's mergable.

packages/core-utils/src/itinerary.ts
```diff
@@ -561,3 +562,69 @@ export function getDisplayedStopId(placeOrStop: Place | Stop): string {
...
+  );
+  const totalCost = relevantFareProducts.find(fp => fp.name === "rideCost")
+    ?.amount;
+  const transferFareProduct = relevantFareProducts.find(
+    fp => fp.name === "transfer"
+  );
+
+  return {
+    price: totalCost,
+    transferAmount: transferFareProduct?.amount?.cents || undefined
```
miles-grant-ibigroup, line 604: Is the undefined not automatic?

packages/core-utils/src/itinerary.ts
```diff
@@ -561,3 +562,69 @@ export function getDisplayedStopId(placeOrStop: Place | Stop): string {
...
+  category: string,
+  container: string
+): Money {
+  return legs
+    .filter(leg => !!leg.fareProducts)
+    .map(leg => getLegCost(leg, category, container).price)
+    .reduce<Money>(
+      (prev, cur) => ({
+        cents: prev.cents + cur?.cents || 0,
+        currency: prev.currency ?? cur?.currency
```
miles-grant-ibigroup, line 626: I don't think multi-currency trips are at the top of our priority list, but will this just take the last currency and show that?

packages/trip-details/i18n/en-US.yml
```diff
@@ -39,3 +39,4 @@ otpUi:
       electronic: "Electronic"
       cash: "Cash"
       special: "Special"
+    transferDiscountExplanation: Transfer discount of {transferAmount} is applied.
```
miles-grant-ibigroup, line 42: Can we get rid of this period?

packages/trip-details/src/fare-table.tsx
```diff
@@ -142,9 +96,18 @@ const FareTypeTable = ({
...
+            let fare;
+            if (!hasLegProducts) {
+              fare = fareTotals[col.key];
+            } else {
+              fare = getItineraryCost(
+                legs,
+                col.riderCategory,
+                col.fareContainer
+              );
+            }
```
miles-grant-ibigroup, line 108: Usually I prefer blocks to a ternary but I feel in this situation a `const` would be cleaner

packages/trip-details/src/fare-table.tsx
```diff
@@ -159,14 +122,49 @@ const FareTypeTable = ({
           <tr key={index}>
             <td className="no-zebra">{leg.routeShortName}</td>
             {colsToRender.map(col => {
-              const fare = leg.fares[col.key];
+              let fare;
+              if (!hasLegProducts) {
+                fare = leg.fares[col.key];
+              } else {
+                fare = getLegCost(leg, col.riderCategory, col.fareContainer);
+              }
```
miles-grant-ibigroup, line 130: Same pattern as earlier

packages/trip-details/src/fare-table.tsx
```diff
@@ -159,14 +122,49 @@ const FareTypeTable = ({
...
+                            style: "currency"
+                          }
+                        )
+                      }
+                    )
+                  }
+                >
+                  {(fare?.isTransfer || fare.transferAmount) && (
+                    <>
+                      <TransferIcon size={16} />{" "}
```
miles-grant-ibigroup, line 161: `paddingRight: '1ch'` instead of a space? Then can remove the fragment

miles-grant-ibigroup (approved):
Couldn't get the server part of this running locally? Think there might be some config I'm missing. Either way story looks great thanks for all these changes!

Just a few very small nitpicks but otherwise looks ready to go to me

packages/core-utils/src/itinerary.ts
daniel-heppner-ibigroup: When the totals were calculated on the server, that's also what happened. There's not really any other way to do it without currency conversions.

packages/core-utils/src/itinerary.ts
daniel-heppner-ibigroup: Yeah not sure why this is there

daniel-heppner-ibigroup: `undefined?.foobar` is undefined

packages/trip-details/src/fare-table.tsx
daniel-heppner-ibigroup: This also makes the typescript work better

packages/trip-details/src/fare-table.tsx
daniel-heppner-ibigroup: I think the space is preferred for LTR languages.

demory (approved):

PR #533, opened 2023-01-26
Location field cleanup

### LINKED ISSUE #354: Typo in `location-field`
The `location-field` package has a typo. https://github.com/opentripplanner/otp-ui/blob/f88cb003597f89388511f4e577a4f99356643edc/packages/location-field/i18n/en-US.yml#L12

When we move to a model of strings for otp-ui packages living in otp-ui instead of being duplicated in both otp-ui and otp-react-redux, we can fix this typo, and remove it from otp-react-redux

https://github.com/opentripplanner/otp-ui/blob/f88cb003597f89388511f4e577a4f99356643edc/packages/location-field/i18n/en-US.yml#L12

binh-dam-ibigroup:
Fix #354.
Requires/blocked by #531.

BREAKING CHANGES! This PR contains a number of breaking changes resulting from cleanup of the location-field package (see commits for more details):
- String id spelling change from #354
- `intl` no longer passed in the `onLocationSelected` event
- `Dropdown` component merged back into main rendering
- `MenuItem`, `FormGroup`, `InputGroupAddon` no longer exported for styling
- `DropdownContainer` removed
- `noResultsFound`, `placeNameWithDetails` removed from translatable content. `resultsFound` modified, and assistive technology text for number of stations/stops/other places has been added.

The PR also contains some accessibility improvements
- list ids are only shown when the suggestion lists are shown.
- Assistive technology instructions for geocoder results found, and for browsing and selecting results.

To visually test with OTP-RR, find and replace the following declaration as follows,
or use https://github.com/opentripplanner/otp-react-redux/pull/853.

```js
const StyledLocationField = styled(LocationField)`
  display: grid;
  grid-template-columns: 30px 1fr 30px;
  width: 100%;

  ${Input} {
    padding: 6px 12px;
  }

  ${MenuItemA} {
    text-decoration: none;
  }

  ${MenuItemA}:hover {
    color: #333;
  }
`

```


binh-dam-ibigroup:
> Only issue is this staying issue with otp-rr when integrating, even after applying the css in the PR description: Screenshot 2023-04-04 at 11 02 16 AM

The screenshot looks like the original state without applying CSS.

miles-grant-ibigroup:
I guess I'm ok merging this in for now, as it looks correct in the storybook. OTP-RR pr can be separate

binh-dam-ibigroup:
Assigning this PR back to me for handling OTP-RR and other fixes.

packages/location-field/src/utils.ts
```diff
@@ -64,3 +70,31 @@ export const getCombinedLabel = (properties: Properties): string => {
...
+export const getGeocoderErrorMessage = (
+  intl: IntlShape,
+  errorText?: string
+): string => {
+  const geocoderUnreachableText = intl.formatMessage({
+    description: "Geocoder unreachable status",
+    id: "otpUi.LocationField.geocoderUnreachable"
+  });
+
+  return addInParentheses(geocoderUnreachableText, errorText);
```
amy-corson-ibigroup, line 99: Do we think seeing the `errorText` is helpful to users? This error message feels a little technical: 
image

amy-corson-ibigroup (approved):
Looks great! Just a few questions:

packages/location-field/src/options.tsx
```diff
@@ -127,36 +158,30 @@ export function UserLocationIcon({
...
+      });
     default:
-      return <>{location.name}</>;
+      return location.name;
   }
 }
 
-export function StoredPlaceName({
-  location,
+export function getStoredPlaceName(
```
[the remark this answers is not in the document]
binh-dam-ibigroup: As in "Favorite (Ruby Thai, Orillia, Tukwila, WA)"?

packages/location-field/src/utils.ts
binh-dam-ibigroup: This was the existing behavior, but good question, @miles-grant-ibigroup what do you think?

packages/location-field/src/styled.ts
```diff
@@ -0,0 +1,180 @@
...
+  font-size: 12px;
+  line-height: 1.42857143;
+  padding: 0px 10px;
+  text-align: center;
+  white-space: nowrap;
+`;
+export const MenuGroupHeader = styled(MenuGroupMisc).attrs({
+  "aria-level": 1,
+  role: "heading"
+})``;
```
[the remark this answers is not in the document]
binh-dam-ibigroup: That's a good point. Maybe we could use a combination of (i) changing the heading to `<h2>` or create a prop for an implementer to request the desired heading, and (ii) remove `arial-level` altogether as a result?

packages/location-field/src/options.tsx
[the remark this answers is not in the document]
amy-corson-ibigroup: That's what I'm thinking, yes! Could be especially helpful if you're saving addresses for friends and family.

packages/location-field/src/styled.ts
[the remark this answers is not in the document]
amy-corson-ibigroup: That sounds great! Creating a prop to check the heading level from the implementer could actually be useful across a few packages, I think most of the headings we've added in the last few months have been with otp-rr as the assumed implementer.

packages/location-field/src/options.tsx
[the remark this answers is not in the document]
binh-dam-ibigroup: Implemented in OTP-RR in https://github.com/opentripplanner/otp-react-redux/pull/853/commits/18e30e90559e1bcfefad9e6e1119cc457cc5474b as I hadn't intended to change the structure and meaning of the location types in this PR.

packages/location-field/src/index.tsx
```diff
@@ -784,73 +868,71 @@ const LocationField = ({
...
+        tabIndex={-1}
+      >
+        <LocationIconComponent locationType={locationType} />
+      </S.DropdownButton>
+      {textControl}
+      {clearButton}
+      {/* Note: always render this status tag regardless of the open state,
+          so that assistive technologies correctly set up status monitoring. */}
+      <S.HiddenContent role="status">
+        {statusMessages.join(", ")}
```
miles-grant-ibigroup, line 904: Can we do this across all languages? I'm not too fussed since it's a very rarely seen feature.

packages/location-field/src/styled.tsx
```diff
@@ -1,317 +0,0 @@
...
-  min-width: 160px;
-  padding: 5px 0;
-  position: absolute;
-  text-align: left;
-  top: 100%;
-  /* this is an annoyingly high number, but is needed to be on top of some otp-rr components */
-  z-index: 1000000;
-`;
-
-export const Dropdown = ({
```
miles-grant-ibigroup, line 60: 🚨🚨🚨I'm not sure we're able to do this. I'm pretty sure I recall trimet or someone else requiring these components to be exported without styles!

packages/location-field/src/utils.ts
```diff
@@ -64,3 +70,31 @@ export const getCombinedLabel = (properties: Properties): string => {
...
+/**
+ * Helper method to append text in parentheses to some other text,
+ * if the added text is not null or blank.
+ */
+export const addInParentheses = (
+  mainText: string,
+  extraText?: string
+): string => {
+  return extraText && extraText !== ""
+    ? `${mainText} (${extraText})`
```
miles-grant-ibigroup, line 83: Is this i18n compliant? What about RTL languages?

packages/location-field/src/index.tsx
```diff
@@ -179,72 +201,143 @@ const LocationField = ({
...
+                  }
+                );
+                if (hasResults) {
+                  // If there are results, concatenate sentences about results found and
+                  // instructions for assistive technology users on how to access results.
+                  const instructions = intl.formatMessage({
+                    description: "Instructions on accessing geocoder results",
+                    id: "otpUi.LocationField.howToAccessResults"
+                  });
+                  message = `${resultsFoundText} ${instructions}`;
```
miles-grant-ibigroup, line 312: Another one -- is this i18n compatible?

packages/location-field/src/options.tsx
```diff
@@ -127,36 +158,30 @@ export function UserLocationIcon({
...
-        />
-      );
+      return intl.formatMessage({
+        defaultMessage: "Work",
+        description: "The work location",
+        id: "otpUi.LocationField.workLocation"
+      });
     default:
-      return <>{location.name}</>;
+      return location.name;
```
miles-grant-ibigroup, line 176: image
Not sure if this is related to this line, but I'm seeing this on the HERE story!

miles-grant-ibigroup (changes requested):
This is a big change to a complicated component, so I won't be too nitpickey. Just a few broader implementation questions.

packages/location-field/src/index.tsx
binh-dam-ibigroup: This one should work because it joins two independent sentences.

packages/location-field/src/utils.ts
binh-dam-ibigroup: It looks like the direction of parentheses should apply automatically depending on the direction of the language (see https://www.w3.org/International/articles/inline-bidi-markup/#mirrored), but we may need extra work if mixing LTR and RTL directions (e.g. RTL place names  and English addresses).

packages/location-field/src/styled.tsx
binh-dam-ibigroup: Can they just remain on the current version? Breaking changes are not automatically upgraded.

miles-grant-ibigroup: What happens if they want a new feature though

binh-dam-ibigroup: Per @grant-humphries on the other discussion thread, they are not using this component, so we should be good. Commit https://github.com/opentripplanner/otp-ui/pull/533/commits/c1877f10ab31f23e188785590efa92beb659256c has the breaking change for the `Dropdown` removal.

packages/location-field/src/utils.ts
binh-dam-ibigroup: Additional info for parentheses:
- Chinese has [extra spacing outside the parentheses](https://en.wikipedia.org/wiki/Chinese_punctuation#Marks_similar_to_European_punctuation)
- [Hebrew uses the regular parentheses](https://en.wikipedia.org/wiki/Hebrew_punctuation#Brackets/parentheses) (see comment above for mirroring)
- [Arabic folks seem to use regular parentheses](https://forum.wordreference.com/threads/arabic-parentheses.157788/) too for non-religious documents.
- Korean: there is typically no space before parentheses.

packages/location-field/src/options.tsx
binh-dam-ibigroup: This is a current behavior of the component. What was the desired effect?

packages/location-field/src/utils.ts
binh-dam-ibigroup: This item is addressed in https://github.com/opentripplanner/otp-ui/pull/533/commits/332726cbc87887218c61f296c7639ab2e09ff403. Thanks for pointing it out.

packages/location-field/src/options.tsx
binh-dam-ibigroup: It looks that the mock is missing one field. I don't remember if that was done on purpose or not. Please advise.

packages/location-field/src/index.tsx
binh-dam-ibigroup: Addressed in https://github.com/opentripplanner/otp-ui/pull/533/commits/6913f20a27f2942c71d290c89196c15fb9a5eaba, although I am not thrilled with what that does in Spanish.

packages/location-field/src/options.tsx
miles-grant-ibigroup: Oh sorry about that I think I got confused. Yes I think it's intentional to check if there's a crash

miles-grant-ibigroup (approved):
Tested again and works great thanks so much for the changes!

PR #535, opened 2023-01-31
Add OTP Geocoder

miles-grant-ibigroup:
OTP has a geocoder now! https://docs.opentripplanner.org/en/dev-2.x/sandbox/GeocoderAPI/

This PR lets you use it within otp-ui. Only the stops endpoint for now, as it would be a lot more work to get the rest of the layers to work (would have to parse the label strings) and they're not that high quality. ARCGIS provides better results without an API key.

### DIFF
--- a/packages/geocoder/src/apis/otp/index.ts
+++ b/packages/geocoder/src/apis/otp/index.ts
@@ -1,6 +1,6 @@
 
 // eslint-disable-next-line prettier/prettier
-import type { AutocompleteQuery  } from "../../geocoders/types"
+import type { AutocompleteQuery } from "../../geocoders/types"
 
 type FetchArgs = {
   url: string
@@ -14,10 +14,10 @@ type OTPGeocoderResponse = {
     description: string,
     id: string
   }[]
-}
+} | undefined
 
 
-function run({  query, url }: FetchArgs): Promise<OTPGeocoderResponse> {
+function run({ query, url }: FetchArgs): Promise<OTPGeocoderResponse> {
   // TODO: Support corners/osm nodes?
   return fetch(`${url}/geocode?corners=false&query=${query}`)
     .then(res => res.text())
@@ -29,9 +29,9 @@ function run({  query, url }: FetchArgs): Promise<OTPGeocoderResponse> {
  * OTP Geocoder
  *
  * @param  {Object} $0
- * @param  {string} $0.url                    The OTP instance, ending with /default/
- * @param  {string} $0.text                       query
- * @return {Promise}                              A Promise that'll get resolved with the autocomplete result
+ * @param  {string} $0.url  The OTP instance, ending with /default/
+ * @param  {string} $0.text query
+ * @return {Promise}        A Promise that'll get resolved with the autocomplete result
  */
 async function autocomplete({
   url,
@@ -47,5 +47,5 @@ function search(): Promise<unknown> { console.warn("Not implemented"); return nu
 function reverse(): Promise<unknown> { console.warn("Not implemented"); return null }
 
 
-export { autocomplete, reverse, search,  };
+export { autocomplete, reverse, search };
 export type { OTPGeocoderResponse }
--- a/packages/geocoder/src/apis/otp/index.ts
+++ b/packages/geocoder/src/apis/otp/index.ts
@@ -1,0 +1,47 @@
+
+// eslint-disable-next-line prettier/prettier
+import type { AutocompleteQuery  } from "../../geocoders/types"
+
+type FetchArgs = {
+  url: string
+  query: string
+}
+
+type OTPGeocoderResponse = { results:{
+  lat: number,
+  lng: number,
+  description: string,
+  id: string
+}[] }
+
+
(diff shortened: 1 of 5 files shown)


amy-corson-ibigroup (approved):
Looks good to me! Very cool to read about the OTP geocoder and see how you're fitting it together here.

Just want to double check, running this locally I'm not seeing stops data come up on location field search, is that expected for this PR?

packages/geocoder/src/apis/otp/index.ts
```diff
@@ -0,0 +1,51 @@
...
+    query: text,
+    url
+  })
+}
+
+function search(): Promise<unknown> { console.warn("Not implemented"); return null }
+function reverse(): Promise<unknown> { console.warn("Not implemented"); return null }
+
+
+export { autocomplete, reverse, search,  };
```
daniel-heppner-ibigroup, line 50: I honestly support trailing commas on the last item in a list like this, but I don't think it's compliant with our code formatting standard.

packages/geocoder/src/apis/otp/index.ts
```diff
@@ -0,0 +1,51 @@
...
+    .then(res => res.text())
+    .then(res => JSON.parse(`{"results": ${res}}`));
+}
+
+/**
+ * Search for an address using
+ * OTP Geocoder
+ *
+ * @param  {Object} $0
+ * @param  {string} $0.url                    The OTP instance, ending with /default/
```
daniel-heppner-ibigroup, line 32: so many spaces!

packages/geocoder/src/apis/otp/index.ts
```diff
@@ -0,0 +1,51 @@
...
+    id: string
+  }[]
+}
+
+
+function run({  query, url }: FetchArgs): Promise<OTPGeocoderResponse> {
+  // TODO: Support corners/osm nodes?
+  return fetch(`${url}/geocode?corners=false&query=${query}`)
+    .then(res => res.text())
+    .then(res => JSON.parse(`{"results": ${res}}`));
```
daniel-heppner-ibigroup, line 24: It feels a little weird to construct a JSON object as a string. Why does the results need to be wrapped in an object under "results" anyway? It seems like in the `geocoders/otp` file you could just remove the `.results` part of accessing the data and then also remove this string.

packages/geocoder/src/apis/otp/index.ts
```diff
@@ -0,0 +1,51 @@
...
+  results: {
+    lat: number,
+    lng: number,
+    description: string,
+    id: string
+  }[]
+}
+
+
+function run({  query, url }: FetchArgs): Promise<OTPGeocoderResponse> {
```
daniel-heppner-ibigroup, line 20: Maybe we should specify in the return type that it's possible to get undefined as a response, which will help enforce null coalescing operators when the function is used.

daniel-heppner-ibigroup (changes requested):
In addition to my comments, can you please add the OTP geocoder to the story to make testing easier?

miles-grant-ibigroup:
@amy-corson-ibigroup yeah this doesn't affect the location field at all at the moment

miles-grant-ibigroup:
@daniel-heppner-ibigroup which OTP server do you think we should point to in the story? want to add a field to set one?

packages/geocoder/src/apis/otp/index.ts
miles-grant-ibigroup: Not sure how this slipped in to begin with...

packages/geocoder/src/apis/otp/index.ts
miles-grant-ibigroup: I wish... Unfortunately the response returned by OTP is invalid JSON so we have to do some processing to it. I tried to figure out how to get a "raw" array working without pelias-stitch throwing a fit, but will have another go...

daniel-heppner-ibigroup: Oh, OTP is returning a straight up array? I see. That should be an easy fix in OTP, we should consider ticketing it in OTP upstream.

miles-grant-ibigroup: Others rely on this api being as is

daniel-heppner-ibigroup:
@miles-grant-ibigroup A field for the endpoint seems fine for me. Should we include a default or is that too much?

packages/geocoder/src/apis/otp/index.ts
daniel-heppner-ibigroup: I think we could just leave an issue for it then, at least

miles-grant-ibigroup:
@daniel-heppner-ibigroup I've added story support for the OTP geocoder,  let me know what you think!

daniel-heppner-ibigroup (approved):
The story is looks really good, loving all the fields with no padding. Thanks for the work!

daniel-heppner-ibigroup:
I just noticed that I got a yarn.lock change when I ran yarn on this. Please double check that yarn.lock is updated

miles-grant-ibigroup:
Good catch thanks!

PR #537, opened 2023-02-01
More lenient transit leg detection

miles-grant-ibigroup:
Sometimes we may want to use a non-gtfs mode type. However, this mode may still be a transit mode! This PR adds support for accepting this non-standard mode as a transit mode.


daniel-heppner-ibigroup (approved):
Makes sense to me to use that property!

amy-corson-ibigroup (approved):

Issue #116, opened 2020-04-27
Support distance display in metric units


binh-dam-ibigroup:
The following components are currently hardcoded to imperial units through call to `humanizeDistanceString()` and should support output to metric units:
- `AccessLegSummary` (package `itinerary-body`)
- `AccessLeg` (package `printable-itinerary`)
- `TransitStopOption` (package `location-field`)

EDIT: Other components may still have to be modified to support metric application-wide, and they will have to be covered under separate issues.

PR #538, opened 2023-02-04
fix: #116 Add outputMetricUnits to config types and use in itinerary-…

amenk:
…body

### DIFF
--- a/packages/itinerary-body/src/defaults/access-leg-description.tsx
+++ b/packages/itinerary-body/src/defaults/access-leg-description.tsx
@@ -99,7 +99,11 @@ export default function AccessLegDescription({
           id="otpUi.AccessLegBody.summaryAndDistance"
           values={{
             // TODO: Implement metric vs imperial (up until now it's just imperial).
-            distance: humanizeDistanceString(leg.distance, false, intl),
+            distance: humanizeDistanceString(
+              leg.distance,
+              config.outputMetricUnits,
+              intl
+            ),
             mode: modeContent,
             place: placeContent
           }}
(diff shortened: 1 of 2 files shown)


amenk:
TODO:

Add to 

packages/location-field/src/options.tsx:90

```
{humanizeDistanceString(stop.dist, config.outputMetricUnits,true)}
```

But we have to pass through the config somehow.

Also in the printable `printable-itinerary` still metric units are used, but I couldn't figure out where this is rendered.

amenk:
2. using react dev tools I confirmed that the config is correct in the print-component - I actually think this is related to my workflow (#773), because printable-itinerary depends on itinerary-body and probably is using the old version. So not a real issue.

amenk:
image

amenk:
Not sure if https://github.com/opentripplanner/otp-ui/pull/538/commits/c389bedaf9a3378795a825da2632c166af1ff9bf is the right approach (to pass the outputMetricUnits into the location field component) of if we should access some global configuration from there.

miles-grant-ibigroup:
Sorry for the long delay on this, but have finally had the time to take a look at this. I don't mind this approach, however is it not possible to use react-intl to handle this?

amenk:
Thanks, what do you mean exactly?

miles-grant-ibigroup:
Is https://formatjs.io/docs/react-intl/api#formatnumber not appropriate here?

amenk:
@miles-grant-ibigroup I don't know. As you can see from the PR, there is already a `humanizeDistanceString` function in the code base which can be used.

It's more about passing the setting, whether to use imperial or metric all the way down to the components.

miles-grant-ibigroup:
Thanks for clarifying, I've discussed it with others and I think your approach is the best way to implement this feature. Once the PR is ready the approach has my approval!

PR #539, opened 2023-02-06
fix(otp-tile-overlay): support enabling layers by default

miles-grant-ibigroup:
The overlays had support for displaying themselves by default. With the new otp2 tile layer we lost this functionality! This PR resolves this, allowing any OTP2 tile layer to be set to be initially visible


philip-cline (approved):
lgtm assuming tests pass

amy-corson-ibigroup (approved):

PR #542, opened 2023-02-14
fix(trip-details): fix transfer detection in fare by leg

daniel-heppner-ibigroup:
There was a bug that was making transfers appear too often in the fare table, which is fixed in this PR


miles-grant-ibigroup (approved):
Not sure if we need the `?` since we already check for the field presence in the first line, but I'm glad the fix is this easy!

daniel-heppner-ibigroup:
It's the `in` returns true even if the field is simply undefined. But having only the second check makes Typescript mad.

amy-corson-ibigroup (approved):

PR #546, opened 2023-02-16
Keyboard Navigation Support for `base-map` Layer Selector

miles-grant-ibigroup:
The layer selector was previously not accessible by keyboard. This PR rectifies this.

### DIFF
--- a/packages/base-map/src/index.tsx
+++ b/packages/base-map/src/index.tsx
@@ -164,7 +164,10 @@ const BaseMap = ({
                   ? "first-layer"
                   : toggleableLayers[0].id;
 
-              const lastId = toggleableLayers[toggleableLayers.length - 1].id;
+              const lastId =
+                toggleableLayers && toggleableLayers?.length > 0
+                  ? toggleableLayers[toggleableLayers.length - 1].id
+                  : baseLayer[baseLayer.length - 1];
               if (e.target.id === firstId || e.target.id === lastId) {
                 setFakeHover(false);
               }
--- a/packages/base-map/src/index.tsx
+++ b/packages/base-map/src/index.tsx
@@ -151,6 +151,8 @@ const BaseMap = ({
         <Styled.LayerSelector
           className="filter-group"
           id="filter-group"
+          tabIndex={0}
+          onFocus={() => setFakeMobileHover(true)}
           onTouchEnd={() => {
             setFakeMobileHover(true);
           }}
@@ -169,6 +171,7 @@ const BaseMap = ({
                         checked={activeBaseLayer === layer}
                         id={layer}
                         onChange={() => setActiveBaseLayer(layer)}
+                        name="base-layer"
                         type="radio"
                       />
                       {baseLayerNames?.[index] || layer}
--- a/packages/base-map/src/index.tsx
+++ b/packages/base-map/src/index.tsx
@@ -73,7 +73,7 @@ const BaseMap = ({
 
   // On mobile hover is unavailable, so we use this variable to use a two tap process
   // to simulate a hover
-  const [fakeMobileHover, setFakeMobileHover] = useState(false);
+  const [fakeMobileHover, setFakeHover] = useState(false);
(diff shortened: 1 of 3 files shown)


packages/base-map/src/index.tsx
```diff
@@ -151,6 +151,8 @@ const BaseMap = ({
         <Styled.LayerSelector
           className="filter-group"
           id="filter-group"
+          tabIndex={0}
```
binh-dam-ibigroup, line 154: Sort props (multiple instances)

binh-dam-ibigroup:
In the Optional Layers story, when the keyboard focus moves from the layer list to outside of it, the list remains displayed, and there is no way to dismiss it.

miles-grant-ibigroup:
Is that a problem?

binh-dam-ibigroup (changes requested):
I think what needs to happen with the layer selector to support broader accessibility is as follows: add a `<button>` (with labels, aria attributes etc) to open/close the list.  Or, try with a `<details>`/`<summary>` element to have the expanding features built in.

miles-grant-ibigroup:
I think we might want to have an offline design call about this -- I see this as a list that's always present, and only "hidden" via hover when a mouse/touch is present. I don't think it should ever be hidden behind a button.

binh-dam-ibigroup:
> Is that a problem?

Yes, if I'm zoomed in (or have a small screen) then I can't see much of the map anymore.

miles-grant-ibigroup:
Even at high zoom levels and small screen size, I'm still able to clearly see the map. Screenshot 2023-02-17 at 12 32 22 PM

The bigger issue in my eyes is the fake keyboard target at the top of the map. I've been trying to remove this, but can't find a way to do it without adding many extra elements... Will keep trying

miles-grant-ibigroup:
@binh-dam-ibigroup I've supported layers hiding via keyboard, but it's not pretty. Let me know what you think.

packages/base-map/src/index.tsx
```diff
@@ -151,9 +151,29 @@ const BaseMap = ({
         <Styled.LayerSelector
           className="filter-group"
           id="filter-group"
+          onFocus={() => setFakeHover(true)}
+          onBlur={e => {
```
binh-dam-ibigroup, line 155: There is still something going on with the blue handling. Using the story at http://localhost:5555/?path=/story/basemap--optional-layers&globals=locale:en-US, you cannot tab to toggle the second layer.

binh-dam-ibigroup (changes requested):
See https://github.com/opentripplanner/otp-ui/pull/546/files#r1115003171 which is blocking.

packages/base-map/src/index.tsx
binh-dam-ibigroup: On the story at http://localhost:5555/?path=/story/basemap--with-multiple-base-layers, the popup still cannot be dismissed.

binh-dam-ibigroup (changes requested):
See https://github.com/ibi-group/otp-ui/pull/12.

miles-grant-ibigroup:
Thanks again! Much cleaner

binh-dam-ibigroup (approved):

packages/base-map/src/index.tsx
```diff
@@ -151,9 +151,9 @@ const BaseMap = ({
         <Styled.LayerSelector
           className="filter-group"
           id="filter-group"
```
amy-corson-ibigroup, line 153: Any chance we could add a `tabIndex` here to take care of the focus order within the map? I think the `LayerSelector` should receive focus before the map attributions if possible!!

amy-corson-ibigroup (approved):
Looks good to me! Just a nitpick about the focus order:

packages/base-map/src/index.tsx
miles-grant-ibigroup: I think the difficulty here is how we get the focus into the selector to begin with. I'll make an issue from this for anyone in the future who'd like to improve this!

PR #547, opened 2023-02-23
fix(map-popup): correct typo in from/to picker

miles-grant-ibigroup:
The map popup was sending backwards coordinates. This PR fixes this typo

### DIFF
--- a/packages/map-popup/src/index.tsx
+++ b/packages/map-popup/src/index.tsx
@@ -24,8 +24,8 @@ const generateLocation = (entity: Entity, name: string) => {
   // @ts-expect-error some of these values may be null, but that's ok
   const { lon: entityLon, lat: entityLat, x, y } = entity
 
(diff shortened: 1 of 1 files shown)


philip-cline (approved):
I remember that longitude is "x" by remembering that coordinates in canada have + latitude and - longitude, and we're very "positive" up here in the North so Latitude must be north so it must be Y so longitude must be "x" hope that can help you too.

AdrianaCeric:
Not me fully believing longitude was "y" and then being confused why you switched the coordinates, then postponing my approval of this PR for like a week while not even consulting Google. I am so much more confident this should solve the "must be within -90 and 90" problem lol.

Also, as another Canadian I can confirm I have not learned this tidbit of knowledge Philip has presented to us here but given my aforementioned limited geographical knowledge I clearly need it.

AdrianaCeric (approved):

PR #548, opened 2023-02-23
fix(location-field): Add aria-activedescendant attribute

binh-dam-ibigroup:
This PR improves a11y to the `LocationField` component, so that screen readers can indicate the active item when the user uses the up/down arrows to browse the list of suggested locations. This is done by using the `aria-activedescendant` attribute to the input field.

### DIFF
--- a/packages/location-field/src/index.tsx
+++ b/packages/location-field/src/index.tsx
@@ -28,6 +28,8 @@ import {
 import * as S from "./styled";
 import { generateLabel, getCombinedLabel } from "./utils";
 
+const optionIdPrefix = "otpui-locf-option-";
+
 // FIXME have a better key generator for options
 let optionKey = 0;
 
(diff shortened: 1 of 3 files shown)


amy-corson-ibigroup (approved):
Looks good to me!

binh-dam-ibigroup:
cc @demory

packages/location-field/src/index.tsx
```diff
@@ -566,6 +569,7 @@ const LocationField = ({
         // Create and return the option menu item
         const option = (
           <TransitStopOption
+            id={`${optionIdPrefix}${itemIndex}`}
```
miles-grant-ibigroup, line 572: Is there any way to extract this string out to prevent it having to be written out multiple times? I'd also prefer to have the `-` in the template than in the prefix

miles-grant-ibigroup (approved):
So much prop drilling oh my but thanks for getting this working! Approved pending one potential cleanup

packages/location-field/src/index.tsx
binh-dam-ibigroup: I was on the fence with this, but thanks, fixed in https://github.com/opentripplanner/otp-ui/pull/548/commits/863460865b553d65d70bdbd65972d8e6ee95cf21.

PR #550, opened 2023-02-28
fix(trip-details): support routes without a short name

miles-grant-ibigroup:
Co-authored-by: daniel-heppner-ibigroup <Daniel.heppner@ibigroup.com>

Some routes don't have a short name. This PR allows the fare table to handle these routes.

### DIFF
--- a/packages/itinerary-body/src/__mocks__/itineraries/otp2-with-fareproducts.json
+++ b/packages/itinerary-body/src/__mocks__/itineraries/otp2-with-fareproducts.json
@@ -728,7 +728,6 @@
         "length": 368
       },
       "steps": [],
-      "routeShortName": "101",
       "routeLongName": "Mariner P&R - Aurora Village",
       "duration": 1560
     },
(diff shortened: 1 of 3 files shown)


packages/itinerary-body/src/__mocks__/itineraries/otp2-with-fareproducts.json
```diff
@@ -728,7 +728,6 @@
         "length": 368
       },
       "steps": [],
-      "routeShortName": "101",
```
AdrianaCeric, line 731: Was this removed for testing?

AdrianaCeric (approved):
👍

packages/itinerary-body/src/__mocks__/itineraries/otp2-with-fareproducts.json
miles-grant-ibigroup: Yes! The story now shows how we handle a route with no short name

philip-cline (approved):
Have we considered the case where leg.route includes an insanely long route_id?

miles-grant-ibigroup:
A very long route id and no long or short name sounds like a data issue to me

PR #551, opened 2023-02-28
Metro mode selector: Types

daniel-heppner-ibigroup:
This PR deprecates old types used for old mode selector code and adds new types to handle the new ModeButtons and other data such as the mode setting options.

The TransportMode type is based on the TransportMode type found in the GraphQL query.

### DIFF
--- a/packages/types/src/deprecated.ts
+++ b/packages/types/src/deprecated.ts
@@ -1,0 +1,29 @@
+export type ModeOption = {
+  id: string;
+  selected?: boolean;
+  showTitle?: boolean;
+  text: JSX.Element;
+  title?: string;
+};
+
+export type ModeSelectorOptions = {
+  primary: ModeOption;
+  secondary?: ModeOption[];
+  tertiary?: ModeOption[];
+};
+
+export type ConfiguredMode =
+  | string
+  | {
+      mode: string;
+      label: string;
+      company?: string;
+    };
+
+export type ConfiguredModes = {
+  transitModes: ConfiguredMode[];
+  accessModes: ConfiguredMode[];
+  exlcusiveModes: ConfiguredMode[];
+  bicycleModes: ConfiguredMode[];
+  micromobilityModes: ConfiguredMode[];
+};
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -4,6 +4,8 @@
  */
 
 import React, { FunctionComponent, ReactElement } from "react";
+import { StyledIcon } from "@styled-icons/styled-icon";
+import { ConfiguredModes } from "./deprecated";
 
 type ZeroOrOne = 0 | 1;
 
@@ -146,7 +148,8 @@ export type Config = {
     longDateFormat?: string;
   };
   homeTimezone: string;
-  modes: ConfiguredModes;
+  /** @deprecated */
+  modes?: ConfiguredModes;
   // TODO: add full typing
   map?: {
     overlays?: {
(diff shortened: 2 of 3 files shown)


packages/types/src/deprecated.ts
```diff
@@ -0,0 +1,29 @@
...
+      company?: string;
+    };
+
+export type ConfiguredModes = {
+  transitModes: ConfiguredMode[];
+  accessModes: ConfiguredMode[];
+  exlcusiveModes: ConfiguredMode[];
+  bicycleModes: ConfiguredMode[];
+  micromobilityModes: ConfiguredMode[];
+};
```
binh-dam-ibigroup, line 29: This configuration objects are still in use in OTP-RR outside of mode selection form, so should they be deprecated?
- `accessModes`: in "My Account" (for wheelchair access), CallTaker
- `transitModes`: in form actions, CallTaker, itinerary narrative, stop marker
- `exclusiveModes`: in Calltaker

packages/types/src/deprecated.ts
```diff
@@ -0,0 +1,29 @@
...
+  | {
+      mode: string;
+      label: string;
+      company?: string;
+    };
+
+export type ConfiguredModes = {
+  transitModes: ConfiguredMode[];
+  accessModes: ConfiguredMode[];
+  exlcusiveModes: ConfiguredMode[];
```
binh-dam-ibigroup, line 26: Oh I guess the typo was there originally!

packages/types/src/index.ts
```diff
@@ -710,6 +683,78 @@ export type GradationMap = Record<
...
+  SLIDER: "SLIDER"
+};
+
+export type DropdownOptions = {
+  default?: string;
+  label: string;
+  options: {
+    text: string;
+    value: string;
+    addTransportMode?: TransportMode;
```
binh-dam-ibigroup, line 698: Sort attributes.

packages/types/src/index.ts
```diff
@@ -582,52 +585,6 @@ export type Station = {
   y: number;
 };
 
-export type ModeOption = {
```
binh-dam-ibigroup, line 585: The breaking change was actually about the loss of these exports.

binh-dam-ibigroup (approved):
Not too hung up on some deprecated, non-exported types that seems to be used inly in niche areas in OTP-RR.

packages/types/src/deprecated.ts
daniel-heppner-ibigroup: Haha, did not even notice that in my copy pasta

packages/types/src/index.ts
daniel-heppner-ibigroup: I believe you can still import them but you have to specify the /deprecated file in your import.

binh-dam-ibigroup: ... which is a breaking change!

daniel-heppner-ibigroup: I'm confused about this comment, there *is* a breaking change commit here for this reason

AdrianaCeric: What's the consensus here? Does removing these exports cause a breaking change?

packages/types/src/index.ts
AdrianaCeric: On the topic of alphabetizing I have a fun VS Code plugin called "Alphabetize" which sorts lines really efficently

AdrianaCeric (approved):
Looks good, just need to followup on Binh's concern about deprecated types

packages/types/src/index.ts
binh-dam-ibigroup: Yes it is a breaking change and it is fine.

packages/types/src/deprecated.ts
AdrianaCeric: We should fix this

packages/types/src/index.ts
AdrianaCeric: Are you implying that we should bring back these exports?

binh-dam-ibigroup: @AdrianaCeric No, because there is already "BREAKING CHANGE" commit that announces this kind of stuff.

daniel-heppner-ibigroup:
The first merge of the mode selector!

PR #557, opened 2023-03-09
Trip form a11y

demory:
Adds accessibility support (screenreader / keyboard navigation) to the DateTimeSelector and SettingsSelectorPanel components and subcomponents.


packages/trip-form/src/SubmodeSelector/index.tsx
```diff
@@ -53,12 +53,18 @@ export default function SubmodeSelector({
...
   return (
-    <S.SubmodeSelector className={className} style={style}>
-      {label && <LabelType>{label}</LabelType>}
+    <S.SubmodeSelector
+      aria-label={label}
+      className={className}
+      role="group"
+      style={style}
+    >
+      {label && <LabelType aria-hidden="true">{label}</LabelType>}
```
binh-dam-ibigroup, line 62: This label is unused by any input element, so I would convert to a span:
```suggestion
      {label && <LabelType aria-hidden="true" as="span">{label}</LabelType>}
```

binh-dam-ibigroup (changes requested):
A few minor changes, and please include a commitizen `fix` commit, so that a new version number can be released.

binh-dam-ibigroup (approved):
Sorry it took a while to clear all a11y test errors. Should be good to go with our latest changes.

packages/trip-form/src/components.story.tsx
```diff
@@ -158,6 +164,17 @@ export const generalSettingsPanelWithOtp2 = makeStory(GeneralSettingsTemplate, {
   mode: "WALK,BUS,TRAM,SUBWAY",
   otp2: true
 });
+// This error may be raised by us rendering two copies of eagh story.
```
amy-corson-ibigroup, line 167:
```suggestion
// This error may be raised by us rendering two copies of each story.
```

packages/trip-form/src/components.story.tsx
```diff
@@ -120,6 +120,12 @@ export const sliderSelector = makeStory(Core.SliderSelector, {
   style: { display: "inline-block", width: "250px" },
   value: 3
 });
+// This error may be raised by us rendering two copies of eagh story.
```
amy-corson-ibigroup, line 123:
```suggestion
// This error may be raised by us rendering two copies of each story.
```

amy-corson-ibigroup (approved):
Two small typos, otherwise LGTM!

PR #559, opened 2023-03-14
Add additional needed parameters to GraphQL Query

daniel-heppner-ibigroup:
Adds support for arrive by and depart at times in the GraphQL Plan query


miles-grant-ibigroup (approved):
Better late than never!

daniel-heppner-ibigroup:
> Any chance we can add `network` on the from/to legs as well?

Good idea! Once you add that, feel free to merge this.

PR #563, opened 2023-03-17
Refactor Select Tag in DropdownSelector for improved iOS dropdown functionality

lightsage88:
This pull request addresses an issue where iOS does not allow for single taps on adjacent dropdowns. Currently, it requires one tap on the unopened dropdown for the opened dropdown to close, and another tap on the previously tapped one to open. To solve this problem, I have imported the React-Select library and used the Select element. React-Select is a mobile-friendly library of various HTML elements that allows for easy selection of options from a dropdown menu.

Changes Made:

Imported the React-Select library
Used the Select element from React-Select to replace the existing dropdowns
Ensured compatibility with mobile devices
Benefits:

Allows for easy selection of options from dropdown menus on iOS devices
Enhances user experience by eliminating the need for double taps on adjacent dropdowns
Provides a mobile-friendly solution that is compatible with a variety of devices
Overall, this pull request aims to improve the functionality and user experience of the dropdown menus on iOS devices.

### DIFF
--- a/package.json
+++ b/package.json
@@ -227,9 +227,5 @@
   },
   "msw": {
     "workerDirectory": "public"
-  },
-  "dependencies": {
-    "@types/react-select": "^5.0.1",
-    "react-select": "^5.7.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -227,5 +227,9 @@
   },
   "msw": {
     "workerDirectory": "public"
+  },
+  "dependencies": {
+    "@types/react-select": "^5.0.1",
+    "react-select": "^5.7.0"
   }
 }
--- a/packages/trip-form/src/DropdownSelector/index.tsx
+++ b/packages/trip-form/src/DropdownSelector/index.tsx
@@ -5,6 +5,11 @@ import * as S from "../styled";
 // eslint-disable-next-line prettier/prettier
 import type { QueryParamChangeEvent } from "../types";
 
+const handleTouchStart = (e: TouchEvent<HTMLSelectElement>) => {
+  e.preventDefault();
+  e.currentTarget.focus();
+};
+
 interface DropdownSelectorProps {
   /**
    * The CSS class name(s) to apply to this element.
@@ -72,11 +77,6 @@ export default function DropdownSelector({
 
   const id = `id-query-param-${name}`;
 
-  const handleTouchStart = (e: TouchEvent<HTMLSelectElement>) => {
-    e.preventDefault();
-    e.currentTarget.focus();
-  };
-
   return (
     <S.DropdownSelector className={className} style={style}>
       {/* The <div> elements below are here for the grid layout, see S.DropdownSelector. */}
@@ -85,7 +85,7 @@ export default function DropdownSelector({
       </div>
 
       <div>
-        <select id={id} onChange={handleChange} value={value} onTouchStart={handleTouchStart}>
+        <select id={id} onChange={handleChange} onTouchStart={handleTouchStart} value={value} >
           {options &&
             options.map((o, i) => (
               <option key={i} value={o.value}>
--- a/packages/trip-form/src/DropdownSelector/index.tsx
+++ b/packages/trip-form/src/DropdownSelector/index.tsx
@@ -1,8 +1,7 @@
 import CSS from "csstype";
-import React, { ChangeEvent, ReactElement, useCallback } from "react";
-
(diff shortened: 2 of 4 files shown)


packages/trip-form/src/DropdownSelector/index.tsx
```diff
@@ -67,6 +71,23 @@ export default function DropdownSelector({
   );
 
   const id = `id-query-param-${name}`;
+  const transformedOptions: TransformedOption[] = options.map(option => ({
+    label: option.text,
+    value: option.value
+  }));
+
+  const customStyles = {
```
miles-grant-ibigroup, line 79: There is no hover effect on the options, which makes it very difficult to see which option is being potentially selected. The new dropdowns are also larger than the old ones, which makes for some very awkward formatting in `TripOptions`.

package.json
```diff
@@ -227,5 +227,9 @@
   },
   "msw": {
     "workerDirectory": "public"
+  },
+  "dependencies": {
+    "@types/react-select": "^5.0.1",
+    "react-select": "^5.7.0"
```
miles-grant-ibigroup, line 233: This increases final bundle size by 5kb post-gzip. This is far too much in my opinion. If we manage to resolve all the other issues with this dropdown, it has to be lazy-loaded.

packages/trip-form/src/DropdownSelector/index.tsx
miles-grant-ibigroup: No focus styling also means that this is next to impossible for keyboard users to use.

miles-grant-ibigroup (changes requested):
Thanks for your first PR! Very cleanly implemented. However, I am definitely leaning against using this package. The bundle size increase is enormous and I forsee a lot of annoying tweaks before we're on par with the system dropdown.

I think the hacks needed to get this PR up to speed will be the same as doing iOS event hacks? Except that those hacks don't increase bundle size.

Finally, we use commitizen style commit messages to handle package releases! Check the master commit log to see what those look like. `fix` commits bumps the minor version number, `feat` commits bump the major version number. We usually do one fix/feat commit per PR. At least one is required for a new package version to be released.

lightsage88:
Ah, what a bummer. I'll just go back to the drawing board and investigate some deep cuts on making Apple things work the way developers want them to! I get the desire to curate a specific experience with your devices when they interact with code in the wild but they really don't make it easy, do they?

Once I crack this, I'll take a closer look into your commit style. So getting the dropdowns to behave the way I want would techinically be a 'fix'? Thus we'd up the fix number, yes? So by that reasoning, that would then make this (once it's accepted and approved) 14.1.14, right?

miles-grant-ibigroup:
Good luck! If things really don't work natively we can look into an approach using this -- it'll take a lot of work but it's work we can do if all other approaches are exhausted.

And yes that's exactly right! feat is generally for big new features. And yes a fix would generate 14.1.14

Please let me know if you have any other questions!

lightsage88:
I absolutely will. Thanks for explaining things so clearly and I'm looking forward to threading the needle on this for all of us to happy!

lightsage88:
I updated the snapshot because it was failing when I tested before my push. I'm not sure what to do with this. Apparently there's a Windows only error with snapshots? Can someone please help me with this?

miles-grant-ibigroup (approved):
Great fix thanks! Works well for me

binh-dam-ibigroup (approved):
Consider the minor tweaks indicated. I don't have an iOS device to check, but code changes seem fine.

packages/trip-form/src/DropdownSelector/index.tsx
```diff
@@ -68,6 +72,11 @@ export default function DropdownSelector({
 
   const id = `id-query-param-${name}`;
 
+  const handleTouchStart = (e: TouchEvent<HTMLSelectElement>) => {
```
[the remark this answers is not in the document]
lightsage88: Done!

packages/trip-form/src/DropdownSelector/index.tsx
```diff
@@ -76,7 +85,7 @@ export default function DropdownSelector({
       </div>
 
       <div>
-        <select id={id} onChange={handleChange} value={value}>
+        <select id={id} onChange={handleChange} value={value} onTouchStart={handleTouchStart}>
```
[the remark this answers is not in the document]
lightsage88: Done!

PR #569, opened 2023-04-06
Avoid Trip Details Crash on Missing Fare Info

miles-grant-ibigroup:
Previous crash could happen if currency was missing, `cents` was missing, or if the entire fare object was missing. This PR allows for all those cases to occur.

### DIFF
--- a/packages/trip-details/src/index.tsx
+++ b/packages/trip-details/src/index.tsx
@@ -92,7 +92,7 @@ const TransitFare = ({
   const currentFare = transitFares[fareKey];
 
   // TODO: Is this needed? Every implementation of TransitFare does a check for currentFare's existence, although not the cents field
-  if (!currentFare?.cents) {
+  if (typeof currentFare?.cents !== "number") {
     return (
       <FormattedMessage
         defaultMessage={defaultMessages["otpUi.TripDetails.transitFareUnknown"]}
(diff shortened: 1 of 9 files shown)


packages/trip-details/src/index.tsx
```diff
@@ -91,6 +91,17 @@ const TransitFare = ({
 }: TransitFareProps): ReactElement => {
   const currentFare = transitFares[fareKey];
 
+  // TODO: Is this needed? Every implementation of TransitFare does a check for currentFare's existence, although not the cents field
+  if (!currentFare?.cents) {
```
binh-dam-ibigroup, line 95: Does that include cases where `cents === 0`?

packages/trip-details/src/index.tsx
```diff
@@ -169,26 +180,29 @@ export function TripDetails({
...
-              }
-              return (
-                <TransitFare
-                  fareKey={fareKey}
-                  key={fareKey}
-                  fareKeyNameMap={fareKeyNameMap}
-                  transitFares={transitFares}
+          {fareDetailsLayout
+            ? // Show full ƒare details by leg
+              transitFares?.[defaultFare] && (
```
binh-dam-ibigroup, line 185: Why is this condition repeated? It is already applied to the containing element at Line 164.

packages/trip-details/src/index.tsx
```diff
@@ -169,26 +180,29 @@ export function TripDetails({
...
+                <FareLegTable
+                  layout={fareDetailsLayout}
+                  itinerary={itinerary}
```
binh-dam-ibigroup, line 188: Sort props.

binh-dam-ibigroup (commented):
Could you advise regarding `cents === 0` and repeating the check for `transitFares?.[defaultFare]`?

packages/trip-details/src/index.tsx
miles-grant-ibigroup: Oh very good catch thank you! Fixed in 26b50fef415e1dbf0339c25a520b56697861532b

packages/trip-details/src/index.tsx
miles-grant-ibigroup: Another great catch thanks! Repaired in 26b50fef415e1dbf0339c25a520b56697861532b

packages/trip-details/src/index.tsx
miles-grant-ibigroup: Sorted in 26b50fef415e1dbf0339c25a520b56697861532b

miles-grant-ibigroup:
Great catches thank you! All are addressed

binh-dam-ibigroup (approved):
Looks good, thanks for the changes!

PR #575, opened 2023-04-24
Remove context menu on map pinch

### LINKED ISSUE #574: Context popup should not appear while pinching map for zoom.
The context popup (see story at http://localhost:5555/?path=/story/basemap--on-context-menu-popup) appears while the map is being pinched, and that is not desired. On small mobile screens, the context popup obstructs the view while the user tries to pinch the map.

To test:
1. Grab a smartphone with touchscreen.
2. Open the URL above (replace `localhost` with IP corresponding to dev machine).
3. Zoom in/out by pinching the map.
4. Observe that the context popup is shown while zooming in/out.

binh-dam-ibigroup:
Fix #574.

### DIFF
--- a/packages/base-map/src/index.tsx
+++ b/packages/base-map/src/index.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from "react";
+import React, { useCallback, useEffect, useState } from "react";
 import { Map, MapProps } from "react-map-gl";
 import maplibregl, { Event } from "maplibre-gl";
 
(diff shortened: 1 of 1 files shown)


AdrianaCeric (approved):
I agree, this context menu felt super clunky when navigating with both fingers. I much prefer opening the context menu up with a long press and the timing feels natural. Nice work! 

(Tested on Samsung S8 and S20)

miles-grant-ibigroup (approved):
Tested and works great thanks!

PR #577, opened 2023-04-28
fix(itinerary-body): Add distances to access legs

### LINKED ISSUE #571: WALK + BIKE: add distance to turn instruction
In both OTP-UI and OTP-RR, the individual walk instructions lack a distance for how far to walk or bike.  A lack of distance in the walk instructions was pointed out as an issue in testing with sight-impaired folks.

Here's a visual of what OTP-UI is currently rendering:
Screenshot 2023-04-17 at 9 52 15 AM

Here's a picture of walk instructions with distance from an older (non OTP-UI) app:
Screenshot 2023-04-17 at 10 01 38 AM

NOTE: the distance information needs to make sense when read aloud by a screen reader (e.g., testing needed to come up with the proper wording, etc...).

miles-grant-ibigroup:
Close #571. This PR adds distances to each walking leg step. Styles are configurable in that the styled components can be overridden. If we want explicit prop-based configuration, that can be added.

Screenshot 2023-04-28 at 1 11 27 PM

### DIFF
--- a/packages/itinerary-body/src/styled.tsx
+++ b/packages/itinerary-body/src/styled.tsx
@@ -594,6 +594,12 @@ export const StepStreetName = styled.span`
   font-weight: 500;
 `;
 
+export const StepLength = styled.span`
+  font-weight: 200;
+  opacity: 0.8;
+  padding-left: 1ch;
+`;
+
 export const StopIdSpan = styled.span`
   font-weight: 200;
   font-size: 0.9em;
--- a/packages/itinerary-body/src/styled.tsx
+++ b/packages/itinerary-body/src/styled.tsx
@@ -596,7 +596,8 @@ export const StepStreetName = styled.span`
 
 export const StepLength = styled.span`
   font-weight: 200;
-  opacity: 0.8;
+  /* This is the lowest opacity which still meets AA text color contrast standards */
+  opacity: 0.8975;
   padding-left: 1ch;
 `;
 
--- a/packages/itinerary-body/src/defaults/access-leg-step.tsx
+++ b/packages/itinerary-body/src/defaults/access-leg-step.tsx
@@ -1,6 +1,7 @@
 import { Step } from "@opentripplanner/types";
 import React, { HTMLAttributes, ReactElement } from "react";
-import { FormattedMessage } from "react-intl";
+import { FormattedMessage, useIntl } from "react-intl";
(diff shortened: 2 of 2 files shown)


fpurcell (approved):
Hey Miles ... looks good.  (And bonus points that it renders also in the print itinerary).

packages/itinerary-body/src/styled.tsx
```diff
@@ -594,6 +594,12 @@ export const StepStreetName = styled.span`
   font-weight: 500;
 `;
 
+export const StepLength = styled.span`
+  font-weight: 200;
+  opacity: 0.8;
```
amy-corson-ibigroup, line 599: Color contrast is getting flagged here.  Maybe keep the opacity at 100%, and instead use `|` to provide some separation in the line? So it would read:
Left on Pioneer Sq N (path) | 22 feet | (Mapillary Btn)

amy-corson-ibigroup (commented):
Looks really nice! Approved pending this small color contrast issue.

packages/itinerary-body/src/styled.tsx
miles-grant-ibigroup: Ugh that's really annoying... I initially had the `|` but it looks really really messy. Can we get away with 0.9?

amy-corson-ibigroup: .9 does pass! Visually I'm not sure there's clear distinction between the opacities though. Maybe an `-` instead of `|`? or would that have the same issue?

miles-grant-ibigroup: I think that's even more confusing... Let me find the lowest opacity that still passes AA. Or are we targeting AAA?

amy-corson-ibigroup: Nope AA is fine! I think it'd be somewhere between `.8` and `.9`

miles-grant-ibigroup: `0.8975` 😵‍💫 (fbb2294ee85649a3b1d9ee3fec22da917411e9bb)

amy-corson-ibigroup (approved):
0.8975 it is! If I'm being aggressively nitpicky I'd say let's either just go with .9 or maybe add a comment saying why the number is _so_ specific, but ultimately I'll leave it up to you!

PR #578, opened 2023-05-03
Map layer button fixes

binh-dam-ibigroup:
This PR proposes the following fixes to the map layer selector:
- Same style and right margin as the zoom controls.
- Expanded layer selector contents appears above the zoom controls.
- Avoid most accidental activations of the first layer when opening the layer selector using touchscreens.

### DIFF
--- a/packages/base-map/src/index.tsx
+++ b/packages/base-map/src/index.tsx
@@ -71,6 +71,9 @@ const BaseMap = ({
     zoom: initZoom
   });
 
+  // Firefox and Safari on iOS: hover is not triggered when the user touches the layer selector
+  // (unlike Firefox or Chromium on Android), so we have to detect touch and trigger hover ourselves.
+  const [fakeMobileHover, setFakeHover] = useState(false);
   const [longPressTimer, setLongPressTimer] = useState(null);
 
   useEffect(() => {
@@ -133,6 +136,7 @@ const BaseMap = ({
         clearLongPressTimer();
       }}
       onTouchStart={e => {
+        setFakeHover(false);
         // Start detecting long presses on screens when there is only one touch point.
         // If the user is pinching the map or does other multi-touch actions, cancel long-press detection.
         const touchPointCount = e.points.length;
@@ -151,8 +155,18 @@ const BaseMap = ({
         (!!baseLayer &&
           typeof baseLayer === "object" &&
           baseLayer.length > 1)) && (
-        <Styled.LayerSelector className="filter-group" id="filter-group">
-          <ul className="maplibregl-ctrl-group layers-list">
+        <Styled.LayerSelector
+          className="filter-group"
+          id="filter-group"
+          onBlur={() => setFakeHover(false)}
+          onFocus={() => setFakeHover(true)}
+          onTouchEnd={() => setFakeHover(true)}
+        >
+          <ul
+            className={`maplibregl-ctrl-group layers-list ${
+              fakeMobileHover ? "fake-mobile-hover" : ""
+            }`}
+          >
             {!!baseLayer &&
               typeof baseLayer === "object" &&
               baseLayer.map((layer: string, index: number) => {
--- a/packages/base-map/src/index.tsx
+++ b/packages/base-map/src/index.tsx
@@ -71,9 +71,6 @@ const BaseMap = ({
     zoom: initZoom
   });
 
-  // On mobile hover is unavailable, so we use this variable to use a two tap process
-  // to simulate a hover
-  const [fakeMobileHover, setFakeHover] = useState(false);
   const [longPressTimer, setLongPressTimer] = useState(null);
 
   useEffect(() => {
@@ -136,7 +133,6 @@ const BaseMap = ({
         clearLongPressTimer();
       }}
       onTouchStart={e => {
-        setFakeHover(false);
         // Start detecting long presses on screens when there is only one touch point.
         // If the user is pinching the map or does other multi-touch actions, cancel long-press detection.
         const touchPointCount = e.points.length;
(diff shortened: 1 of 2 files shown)


packages/base-map/src/index.tsx
```diff
@@ -155,16 +151,8 @@ const BaseMap = ({
         (!!baseLayer &&
           typeof baseLayer === "object" &&
           baseLayer.length > 1)) && (
-        <Styled.LayerSelector
-          className="filter-group"
-          id="filter-group"
-          onBlur={() => setFakeHover(false)}
-          onFocus={() => setFakeHover(true)}
-          onTouchEnd={() => setFakeHover(true)}
```
miles-grant-ibigroup, line 163: I'm not longer able to open the layer selector when using a device without a mouse. Am I doing something wrong?

packages/base-map/src/styled.ts
```diff
@@ -73,15 +73,11 @@ export const LayerSelector = styled.aside`
   }
 
   .layers-list {
-    background: rgba(255, 255, 255, 0.95);
```
miles-grant-ibigroup, line 76: Why was this removed? I understand matching the zoom buttons but those are tighter and also have a box shadow

packages/base-map/src/styled.ts
```diff
@@ -100,14 +96,18 @@ export const LayerSelector = styled.aside`
       cursor: pointer;
     }
 
-    &:hover,
-    &.fake-mobile-hover {
+    /* :hover state is set on mouse over and on click on touch screens. */
```
miles-grant-ibigroup, line 99: This is not happening for me

miles-grant-ibigroup (changes requested):
I'd like a cleanup to be this easy, but I'm not sure it's fully working...

packages/base-map/src/styled.ts
binh-dam-ibigroup: Which device(s) are you using?

binh-dam-ibigroup: *device(s) and browser(s)

miles-grant-ibigroup: Had the same issue occur on Firefox for Android, Firefox for Mac, and Safari for Mac

packages/base-map/src/styled.ts
binh-dam-ibigroup: I am such a non-believer.

binh-dam-ibigroup: How about we keep the previous box shadow on hover?

packages/base-map/src/index.tsx
binh-dam-ibigroup: That was a mistake. Reinstated in https://github.com/opentripplanner/otp-ui/pull/578/commits/8c32bb4c8f950822c2ff724d31e0efaf3d25db14.

packages/base-map/src/styled.ts
binh-dam-ibigroup: Reverted the hover stuff in https://github.com/opentripplanner/otp-ui/pull/578/commits/8c32bb4c8f950822c2ff724d31e0efaf3d25db14.

packages/base-map/src/styled.ts
miles-grant-ibigroup: That's an acceptable compromise to me!

packages/base-map/src/styled.ts
```diff
@@ -103,11 +100,15 @@ export const LayerSelector = styled.aside`
     &:hover,
     &.fake-mobile-hover {
       label {
-        display: block;
```
miles-grant-ibigroup, line 106: I *think* this line is needed to override other styles that are applied -- even though it's defined above.

miles-grant-ibigroup (approved):
Looking alright to me, approved pending that box shadow on hover. I'll leave it to the second reviewer to really double check behavior.

packages/base-map/src/styled.ts
binh-dam-ibigroup: I added the on-hover box shadow in https://github.com/opentripplanner/otp-ui/pull/578/commits/7f0e274f8cc99694d04136d1e934cb5fa189c51b.

packages/base-map/src/styled.ts
binh-dam-ibigroup: `display:block` is still in effect on the client (see screenshot).

image

packages/base-map/src/styled.ts
```diff
@@ -74,14 +74,11 @@ export const LayerSelector = styled.aside`
...
-    box-shadow: 0px -1px 15px -3px rgba(0, 0, 0, 0.1);
-    display: flex;
-    flex-direction: column;
     list-style-type: none;
     padding: 1em;
     position: absolute;
     right: 0;
+    z-index: 1000;
 
     label {
```
amy-corson-ibigroup, line 83: While we're in here, I'm wondering if we can add a `margin-bottom: 0` here and a `margin-bottom: 5px` on `&.fake-mobile-hover` at line 102? There's a small margin on every list item that's skewing the button: 

image

packages/base-map/src/styled.ts
amy-corson-ibigroup: Can confirm the box shadow looks good on hover!

amy-corson-ibigroup (approved):
Looks really good! I've added a small suggestion regarding a small existing issue with the mobile button:

packages/base-map/src/styled.ts
binh-dam-ibigroup: I am aware of this glitch. That styling is set by OTP-RR, so I am looking to do a PR there to fix that.

PR #584, opened 2023-05-15
fix(route-viewer-overlay): fix crash on incomplete flex data

miles-grant-ibigroup:
There was a previous bug where the overlay would attempt to center the map on non-existent data. This PR resolves this issue.


daniel-heppner-ibigroup (approved):
nice! thanks for the simple fix.

amy-corson-ibigroup (approved):
Looks great!

PR #585, opened 2023-05-16
Support GeometryCollection in flex data

miles-grant-ibigroup:
Flex data was previously only rendered if the geojson type was "polygon". Some flex data however used "GeometryCollection" instead. This would cause these flex zones not to render. 

This PR repairs how we determine what geojson is flex data, and so allows this data to be rendered. There is also a new story to test out the new behavior.


philip-cline (approved):
Fix looks good, I'm not sure why Typescript won't let you use the util component. I don't think it's a big deal tho since it's in a test component.

PR #587, opened 2023-05-17
I18n check exceptions

binh-dam-ibigroup:
This is a feature PR on the `scripts` package, a minor release version will be created after merging.

This PR adds a mechanism to ignore certain id strings during i18n validation (`yarn check:i18n*`). Basically, in each affected i18n folder, add a `i18n-exceptions.json` file with a list of ignored message ids. Other types of i18n exceptions can be added to that file in the future.
```json
{
  "ignoredIds": [
    "message.id.1",
    "message.id.2"
  ]
}
```


daniel-heppner-ibigroup (approved):
Wow, we have tests for our tests! Things are getting deep.

Thank you for adding the ignored IDs feature. It seems like it doesn't support wildcards, so we couldn't have `OtpUiCustom.*`as a general place to throw all the "generated" IDs. But maybe you wanted it to be fully explicit, which also makes sense to me as a design consideration.

binh-dam-ibigroup:
> It seems like it doesn't support wildcards, so we couldn't have `OtpUiCustom.*`...

That's what I wanted to work on next.

miles-grant-ibigroup (approved):
Incredibly thorough and well developed! Thanks for this, this will be a great addition to the project.

PR #555, opened 2023-03-08
Metro mode selector trip form

daniel-heppner-ibigroup:
This PR includes the actual trip-form component for the new mode selector.

PR #588, opened 2023-05-24
fix: Refactor TNC Support to be OTP2 Compatible

miles-grant-ibigroup:
TNC Support was built around the OTP1 api. With OTP1 being phased out, this PR uses the GraphQL API to get the data in the new format, and updates all relevant components accordingly.

~~Blocked by #555~~ Unblocked


daniel-heppner-ibigroup:
Please merge master into this branch and run the graphql formatter.

packages/types/src/index.ts
daniel-heppner-ibigroup, line 1: I think you will need to cherry pick this update out to another PR so that you can update the types versions in the other packages.

packages/types/src/index.ts
```diff
@@ -319,18 +327,8 @@ export type Leg = {
   routeTextColor?: string;
   routeType?: number;
   serviceDate?: string;
-  startTime: number;
+  startTime: number | string;
```
daniel-heppner-ibigroup, line 330: Why is this possibly a string? In the API response it is just a long. 
image

packages/types/src/index.ts
```diff
@@ -343,6 +341,13 @@ export type Leg = {
   fareProducts?: Array<FareProduct>;
 };
 
+type Price = {
```
daniel-heppner-ibigroup, line 344: I am already adding this type in my fare related PR, and I'm using the name `Money` so that it matches what is used in the GraphQL API. That meant replacing the old `Money` in our Types package with the new one. Take a look: 
https://github.com/ibi-group/otp-ui/commit/d1025105774caa3e9e2ddfd25dc6367ac0e537d3#diff-e71f2026e90f1ed0c00f8711eb855f0a8810f33dbf1491f5da9c97cdd513e5b4L349

How do you want to reconcile this? If you want to implement the same change then we can resolve this in a merge conflict (unless Git figures it out?) in whichever PR gets merged later.

packages/types/src/index.ts
```diff
@@ -307,6 +307,14 @@ export type Leg = {
   mode: string;
   pathway: boolean;
   pickupBookingInfo?: FlexPickupBookingInfo;
+  rideHailingEstimate?: {
```
daniel-heppner-ibigroup, line 310: Can you add `productName?: string` to this type so that it matches what exists in the API? That way we don't need to do an update to types if we ever want that property. 
image

packages/types/src/index.ts
daniel-heppner-ibigroup: Another idea: since you probably need this file to be changed in a separate PR anyway, let's add all of our types changes in to that one PR. In fact, I already have a branch that I was preparing that you can add your changes to: https://github.com/ibi-group/otp-ui/tree/gql-transition/fare-product-types

packages/itinerary-body/src/util.ts
```diff
@@ -174,3 +174,8 @@ export function getPlaceName(
   // Default to place name
   return place.name;
 }
+
+// TODO: is this the best way to do this?
```
daniel-heppner-ibigroup, line 178: is there a `date-fns` function that could do this for us?

daniel-heppner-ibigroup (changes requested):
Good work on this change! Making progress towards dropping REST. We should coordinate on the work on the types package, though!

packages/types/src/index.ts
miles-grant-ibigroup: TNC legs can be a string I believe

packages/types/src/index.ts
miles-grant-ibigroup: Is this the same? I think TNC fares are using a separate data structure

packages/itinerary-body/src/util.ts
miles-grant-ibigroup: I looked and failed

packages/types/src/index.ts
miles-grant-ibigroup: Good catch thanks

packages/types/src/index.ts
miles-grant-ibigroup: It is. I have a fix I think you'll like

miles-grant-ibigroup:
Thanks, all changes are addressed. Rest can be dealt with in a separate PR.

daniel-heppner-ibigroup (approved):

amy-corson-ibigroup (approved):

PR #589, opened 2023-05-25
I18n message groups

binh-dam-ibigroup:
This PR allows declaring message groups using the `i18n-exceptions.json` file (see one of the files provided in the PR), so that such messages are treated accordingly during validation and are also included in CSV output produced by `collect-i18n-messages`.

### DIFF
--- a/packages/scripts/src/__tests__/__mocks__/i18n1/en-US.yml
+++ b/packages/scripts/src/__tests__/__mocks__/i18n1/en-US.yml
@@ -3,6 +3,9 @@ otpUi:
     from: From here
     planATrip: "Plan a trip:"
     to: To here
+  OtherComponent:
+    key1Message: Key 1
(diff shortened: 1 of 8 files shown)


daniel-heppner-ibigroup (changes requested):
This took me quite a while to figure out what the intended functionality is... although I really like the idea, it is not very easy to figure out. Can we add some documentation in the scripts folder that describes how the exceptions file works?

binh-dam-ibigroup:
> Fantastic documentation! Helped me understand better how it's supposed to work. Should we remove some of the big switch statements in the code in favor of using these groups?

@daniel-heppner-ibigroup Maybe one case such as the code that prints the different fare types? The other ones would have to be in other PRs.

miles-grant-ibigroup (approved):
This is finicky work! Thanks for doing it

PR #592, opened 2023-05-26
fix(trip-form): fix a missing export

daniel-heppner-ibigroup:
Found an export missing from index, this PR is to fix that.

### DIFF
--- a/packages/trip-form/src/index.ts
+++ b/packages/trip-form/src/index.ts
@@ -11,6 +11,7 @@ import SubmodeSelector from "./SubmodeSelector";
 import MetroModeSelector from "./MetroModeSelector";
 import {
   addSettingsToButton,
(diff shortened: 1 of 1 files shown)


philip-cline (approved):
easy win

miles-grant-ibigroup (approved):
Approving because I think I know the answer, but are we sure we are using this outside of `trip-form`?

daniel-heppner-ibigroup:
> Approving because I think I know the answer, but are we sure we are using this outside of `trip-form`?

@miles-grant-ibigroup  It's used outside of trip-form now because we are no longer handling state of the mode selector inside the OTP-UI component, it's now handled directly by OTP-RR, and therefore OTP-RR is handling the aggregation of modes from the enabled buttons.

PR #595, opened 2023-05-29
Location Field: Cancel pending requests on new request

miles-grant-ibigroup:
Attempt to resolve network-based race conditions. Adds a new abort controller. Aborts and resets on keystroke.

### DIFF
--- a/packages/location-field/src/index.tsx
+++ b/packages/location-field/src/index.tsx
@@ -5,7 +5,7 @@ import getGeocoder from "@opentripplanner/geocoder";
 // @ts-ignore Not Typescripted Yet
 import LocationIcon from "@opentripplanner/location-icon";
 import { Location } from "@opentripplanner/types";
-import React, { useEffect, useMemo, useState, useRef } from "react";
+import React, { useEffect, useState, useRef } from "react";
 import { FormattedList, FormattedMessage, useIntl } from "react-intl";
 import { Ban } from "@styled-icons/fa-solid/Ban";
 import { Bus } from "@styled-icons/fa-solid/Bus";
@@ -177,6 +177,8 @@ const LocationField = ({
 
   const listBoxId = `listbox-${optionKey}`;
 
+  let controller = new AbortController();
+
   const intl = useIntl();
 
   const [activeIndex, setActiveIndex] = useState(null);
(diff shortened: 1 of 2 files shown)


packages/location-field/src/index.tsx
```diff
@@ -201,132 +204,128 @@ const LocationField = ({
...
-                const hasResults = parts.length !== 0;
-                const results = hasResults
-                  ? intl.formatList(parts, { type: "conjunction" })
-                  : intl.formatMessage({
-                      description: "Indicates no results",
-                      id: "otpUi.LocationField.noResults"
-                    });
-                const resultsFoundText = intl.formatMessage(
+    getGeocoder(geocoderConfig)
+      // .autocomplete({ text })
```
daniel-heppner-ibigroup, line 225: remove comment

packages/location-field/src/index.tsx
```diff
@@ -185,6 +185,9 @@ const LocationField = ({
   const [isFetching, setFetching] = useState(false);
   const [stateMessage, setMessage] = useState(null);
   const [stateValue, setValue] = useState(getValueFromLocation());
+  const [abortControllers, setAbortController] = useState([
+    new AbortController()
```
daniel-heppner-ibigroup, line 189: Is there any need to intitalize this with an `AbortController`? I don't see this controller being used anywhere. You can set the type of the state with `useState<AbortController[]>`.

daniel-heppner-ibigroup (approved):
Interesting! I have never seen the Abort Controller before. This is a neat solution! I think you may have left some things behind though.

packages/location-field/src/index.tsx
miles-grant-ibigroup: Oh thanks for catching this yes this is residue

packages/location-field/src/index.tsx
```diff
@@ -201,132 +202,127 @@ const LocationField = ({
...
-        setFetching(true);
-        setMessage(
-          intl.formatMessage({
-            defaultMessage: "Fetching suggestions…",
-            description:
-              "Hint shown while geocoder suggestions are being fetched",
-            id: "otpUi.LocationField.fetchingSuggestions"
-          })
-        );
+  const geocodeAutocomplete = debounce(300, (text: string) => {
```
binh-dam-ibigroup, line 205: Would have to check in the future if we should consider a `useCallback`.

binh-dam-ibigroup (approved):
Approving for now. We'll have to revisit the big long `geocodeAutoComplete` method in the future.

packages/location-field/src/index.tsx
miles-grant-ibigroup: I've added a comment

PR #597, opened 2023-06-05
Suggest saved locations while typing

binh-dam-ibigroup:
This PR changes the `LocationField` component so that, when typing the name or the address of a provided user-saved location, that location is shown at the top of the suggestions list.


amy-corson-ibigroup (approved):
Tested in OTP-rr and I think it looks great! Thanks so much

miles-grant-ibigroup (approved):
Works great! Thanks for wrangling this tough pacakge

PR #604, opened 2023-06-19
GraphQL Trip Stop Sequence Support

miles-grant-ibigroup:
This PR adds stop sequence support when using the new GraphQL API.

### DIFF
--- a/packages/itinerary-body/src/TransitLegBody/index.tsx
+++ b/packages/itinerary-body/src/TransitLegBody/index.tsx
@@ -324,9 +324,15 @@ class TransitLegBody extends Component<Props, State> {
                   {showViewTripButton && (
                     <ViewTripButton
                       tripId={leg.tripId}
-                      fromIndex={leg.from.stopIndex}
+                      fromIndex={
+                        leg.from.stopIndex ||
+                        leg?.trip?.departureStoptime?.stopPosition
+                      }
                       setViewedTrip={setViewedTrip}
-                      toIndex={leg.to.stopIndex}
+                      toIndex={
+                        leg.to.stopIndex ||
+                        leg?.trip?.arrivalStoptime?.stopPosition
+                      }
                     />
                   )}
                 </S.TransitLegDetailsHeader>
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -333,6 +333,13 @@ export type Leg = {
   };
   to: Place;
   transitLeg: boolean;
+  trip?: {
+    arrivalStoptime?: TripStopTime;
+    departureStoptime?: TripStopTime;
+    gtfsId?: string;
+    id: string;
+    tripHeadsign?: string;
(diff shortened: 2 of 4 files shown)


packages/itinerary-body/src/TransitLegBody/index.tsx
```diff
@@ -324,9 +324,15 @@ class TransitLegBody extends Component<Props, State> {
...
                       tripId={leg.tripId}
-                      fromIndex={leg.from.stopIndex}
+                      fromIndex={
+                        leg.from.stopIndex ||
+                        leg?.trip?.departureStoptime?.stopPosition
+                      }
                       setViewedTrip={setViewedTrip}
-                      toIndex={leg.to.stopIndex}
+                      toIndex={
+                        leg.to.stopIndex ||
```
daniel-heppner-ibigroup, line 333: are we keeping this old `.to.stopIndex` on here for legacy support? I guess it doesn't hurt anyone. But I think the newer one should be prioritized.

packages/types/src/index.ts
```diff
@@ -333,6 +333,13 @@ export type Leg = {
   };
   to: Place;
   transitLeg: boolean;
+  trip?: {
```
daniel-heppner-ibigroup, line 336: Can we make Trip its own type? And I'd like to include the full type from the GraphQL schema, but we can do that later in our OTP2/GraphQL complete refactor.

packages/types/src/index.ts
```diff
@@ -333,6 +333,13 @@ export type Leg = {
...
+    arrivalStoptime?: TripStopTime;
+    departureStoptime?: TripStopTime;
+    gtfsId?: string;
+    id: string;
+    tripHeadsign?: string;
```
daniel-heppner-ibigroup, line 341: this field and `gtfsId` are both required on the schema. Should we make them required here too? Just because they're required on the schema doesn't mean we have to request them in our query, so I can see it both ways.

daniel-heppner-ibigroup (approved):
Looks good, just a few thoughts about types as usual

amy-corson-ibigroup (approved):

packages/itinerary-body/src/TransitLegBody/index.tsx
miles-grant-ibigroup: Oh yes this is in here for now since itinerary-body still supports OTP1

packages/types/src/index.ts
miles-grant-ibigroup: Yes I'd prefer to do this in a future PR

packages/types/src/index.ts
miles-grant-ibigroup: The data passed on this component is sometimes generated by a client instead of fed right in from OTP. Do you think it's still worth making this change?  I guess those projects would then throw TS errors, which is good.

PR #613, opened 2023-07-03
add fareproducts to query + update internal dependencies

daniel-heppner-ibigroup:
In the fare update PR, it seems like these changes somehow got overwritten so they didn't make it in. This will generate a fix release that updates the GraphQL query for the new schema. It also updates the schema by copying it from the latest version of OTP. 

Also, it runs prettier to format the GraphQL file, since it turns out Prettier supports that natively!

### DIFF
--- a/package.json
+++ b/package.json
@@ -259,6 +259,7 @@
       "git add"
     ],
     "*.graphql": [
+      "prettier --write",
       "yarn lint:graphql --fix",
       "git add"
     ],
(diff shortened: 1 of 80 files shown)


daniel-heppner-ibigroup:
The dependencies were a big problem for some reason, so I updated internal dependencies.

binh-dam-ibigroup:
@daniel-heppner-ibigroup Is the intent to only release a new package for core-utils? Note that commit https://github.com/opentripplanner/otp-ui/pull/613/commits/15d741e84448d63fae1fb045185daabe15f10d16 will not trigger a release for the other ones you've updated.

miles-grant-ibigroup (approved):
I hope this fixes things

PR #614, opened 2023-07-03
Add fare product types to types package

daniel-heppner-ibigroup:
This PR adds the new fare product types to the types package. It breaks compatibility with older versions of OTP and REST API.

### DIFF
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -746,6 +746,11 @@ export type FareProduct = {
   };
 };
 
+export type FareProductSelector = {
+  riderCategoryId: string;
+  mediumId: string;
+};
+
 /**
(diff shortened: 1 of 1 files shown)


miles-grant-ibigroup (approved):
I'm not sure if a refactor commit can be a breaking change, but other than that this works great for me! Thanks for the changes.

daniel-heppner-ibigroup:
don't know how to fix this CI issue... it's failing because the types break stuff in other packages. but if I merge them in the same PR we can't update the version number and so they still fail.

miles-grant-ibigroup:
I think the solution is to merge it all, get the ci to pass, merge into alpha, do an alpha release, use those versions here, and then it releases. then a subsequent pr updates the version numbers away from alpha

PR #622, opened 2023-07-19
Handle null fareproduct IDs in core-utils

daniel-heppner-ibigroup:
This PR allows core-utils to handle null ID values in fare products which can be returned by OTP on instances without specific fare modules. (GTFS fares)

### DIFF
--- a/packages/core-utils/src/itinerary.ts
+++ b/packages/core-utils/src/itinerary.ts
@@ -569,12 +569,15 @@ export function getLegCost(
 ): { price?: Money; transferAmount?: Money | undefined } {
   if (!leg.fareProducts) return { price: undefined };
   const relevantFareProducts = leg.fareProducts.filter(({ product }) => {
+    // riderCategory and medium can be specifically defined as null to handle
+    // generic GTFS based fares from OTP when there is no fare model
     return (
       (product.riderCategory === null ? null : product.riderCategory.id) ===
         riderCategoryId &&
       (product.medium === null ? null : product.medium.id) === mediumId
     );
   });
+  // Custom fare models return "rideCost", generic GTFS fares return "regular"
   const totalCost = relevantFareProducts.find(
     fp => fp.product.name === "rideCost" || fp.product.name === "regular"
   )?.product?.price;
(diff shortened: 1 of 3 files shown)


miles-grant-ibigroup (approved):
Really clean!

packages/core-utils/src/itinerary.ts
```diff
@@ -564,18 +564,19 @@ export function getDisplayedStopId(placeOrStop: Place | Stop): string {
...
-  riderCategoryId: string
+  mediumId: string | null,
+  riderCategoryId: string | null
 ): { price?: Money; transferAmount?: Money | undefined } {
   if (!leg.fareProducts) return { price: undefined };
   const relevantFareProducts = leg.fareProducts.filter(({ product }) => {
     return (
-      product.riderCategory.id === riderCategoryId &&
-      product.medium.id === mediumId
+      (product.riderCategory === null ? null : product.riderCategory.id) ===
```
binh-dam-ibigroup, line 573: Can this simply be written as `product.riderCategory?.id`? Same for `product.medium`.

packages/core-utils/src/itinerary.ts
```diff
@@ -564,18 +564,19 @@ export function getDisplayedStopId(placeOrStop: Place | Stop): string {
...
+        riderCategoryId &&
+      (product.medium === null ? null : product.medium.id) === mediumId
     );
   });
   const totalCost = relevantFareProducts.find(
-    fp => fp.product.name === "rideCost"
+    fp => fp.product.name === "rideCost" || fp.product.name === "regular"
```
binh-dam-ibigroup, line 579: Add a comment for the "regular" case.

binh-dam-ibigroup (approved):
LGTM with some cleanup.

packages/core-utils/src/itinerary.ts
miles-grant-ibigroup: We do this to handle null differently from undefined

binh-dam-ibigroup: Thanks for the clarification. Add what you said in a comment above that code, please.

PR #626, opened 2023-07-24
fix(location-field): Unpin geocoder version.

binh-dam-ibigroup:
Unpin the geocoder version to prevent duplicates in OTP-RR.


daniel-heppner-ibigroup (approved):
Nice, thanks!

miles-grant-ibigroup (approved):
Thanks!

PR #627, opened 2023-07-24
Handle fareproduct use IDs correctly

daniel-heppner-ibigroup:
This fixes fares being counted multiple times when a fare module doesn't provide per-leg fares.

### DIFF
--- a/packages/core-utils/src/itinerary.ts
+++ b/packages/core-utils/src/itinerary.ts
@@ -620,7 +620,7 @@ export function getItineraryCost(
       }
       return prev;
     }, [])
-    .map(prodUse => prodUse.price);
+    .map(productUse => productUse.price);
 
   if (legCosts.length === 0) return undefined;
   // Calculate the total
--- a/packages/core-utils/src/itinerary.ts
+++ b/packages/core-utils/src/itinerary.ts
@@ -566,7 +566,7 @@ export function getLegCost(
   leg: Leg,
   mediumId: string | null,
   riderCategoryId: string | null
-): { price?: Money; transferAmount?: Money | undefined } {
+): { price?: Money; transferAmount?: Money | undefined; useId?: string } {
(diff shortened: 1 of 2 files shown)


leonardehrenfried:
I support this change as it looks correct to me.

packages/core-utils/src/itinerary.ts
```diff
@@ -566,7 +566,7 @@ export function getLegCost(
   leg: Leg,
   mediumId: string | null,
   riderCategoryId: string | null
-): { price?: Money; transferAmount?: Money | undefined } {
+): { price?: Money; transferAmount?: Money | undefined; useId?: string } {
```
binh-dam-ibigroup, line 569: Maybe rename `useId` to `productId`?

binh-dam-ibigroup (approved):
LGTM, I would rename `useId`  as indicated but the other reviewer can weigh in.

packages/core-utils/src/itinerary.ts
leonardehrenfried: The thing is, it's _not_ the product id. It is an id that refers to the "instance" of the product, so for example a day pass that you use several times, hence "useId".

The product id globally identifies the product that you're buying and probably has an id like `atl:daypass` but the "useId" is an identifier local to the itinerary allows you to deduplicate the day pass so you know you don't have to buy three passes but it's applicable to all of the legs.

binh-dam-ibigroup: Okay. How about `ticketId`?

daniel-heppner-ibigroup: How about `productUseId` to more closely match what it's called in the backend?

binh-dam-ibigroup: Sure

miles-grant-ibigroup (approved):
Well done figuring out the fix! No notes from me, I think all variable names are fine.

PR #630, opened 2023-07-26
fix(trip-details): update core utils

daniel-heppner-ibigroup:
This implements the changes from #629 in trip details.

### DIFF
--- a/packages/trip-details/package.json
+++ b/packages/trip-details/package.json
@@ -11,7 +11,7 @@
   "license": "MIT",
   "private": false,
   "dependencies": {
-    "@opentripplanner/core-utils": "^9.0.2",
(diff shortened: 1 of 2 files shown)


binh-dam-ibigroup (commented):
Is there an example for using the new `productUseId` field?

daniel-heppner-ibigroup:
> Is there an example for using the new `productUseId` field?

It's not used directly by trip-details, but by the getItineraryCost function in core-utils. I don't think we need the trip-details story to showcase every use case of its dependencies; there is a test in core-utils which validates this functionality already.

binh-dam-ibigroup (approved):

amy-corson-ibigroup (approved):
LGTM!

PR #632, opened 2023-07-31
Add OTP2 advanced leg route attributes

binh-dam-ibigroup:
This PR adds missing OTP2 leg route attributes to the `Leg` type (see the graphql query template for details).

Should create a fix release for the types package, and no release to the itinerary-body package (that will be handled separately).

### DIFF
--- a/packages/itinerary-body/src/defaults/line-column-content.tsx
+++ b/packages/itinerary-body/src/defaults/line-column-content.tsx
@@ -65,7 +65,7 @@ export default function LineColumnContent({
     }
   );
 
-  const routeNumber = typeof route === "string" ? route : route?.shortName;
+  const routeShortName = typeof route === "string" ? route : route?.shortName;
 
   return (
     <S.LegLine>
@@ -75,7 +75,7 @@ export default function LineColumnContent({
         {!interline && !isDestination && transitLeg && (
           <RouteBadge
             abbreviation={toRouteAbbreviation(
-              parseInt(routeNumber, 10) || routeNumber
+              parseInt(routeShortName, 10) || routeShortName
             )}
             color={routeColor}
             name={routeLongName || ""}
--- a/packages/itinerary-body/src/defaults/line-column-content.tsx
+++ b/packages/itinerary-body/src/defaults/line-column-content.tsx
@@ -65,6 +65,8 @@ export default function LineColumnContent({
     }
   );
 
+  const routeNumber = typeof route === "string" ? route : route?.shortName;
+
   return (
     <S.LegLine>
       {!isDestination && <S.InnerLine mode={mode} routeColor={routeColor} />}
@@ -72,7 +74,9 @@ export default function LineColumnContent({
         {/* TODO: This is a placeholder for a routebadge when we create the transit leg */}
         {!interline && !isDestination && transitLeg && (
           <RouteBadge
-            abbreviation={toRouteAbbreviation(parseInt(route, 10) || route)}
+            abbreviation={toRouteAbbreviation(
+              parseInt(routeNumber, 10) || routeNumber
+            )}
             color={routeColor}
             name={routeLongName || ""}
           />
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -202,6 +202,8 @@ export type Alert = {
   alertDescriptionText?: string;
   alertUrl?: string;
   effectiveStartDate?: number;
+  /** Returned by OTP2 graphql queries, but not by OTP1 */
+  id?: string;
 };
 
 /**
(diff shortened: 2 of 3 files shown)


packages/types/src/index.ts
```diff
@@ -202,6 +202,8 @@ export type Alert = {
   alertDescriptionText?: string;
   alertUrl?: string;
   effectiveStartDate?: number;
+  /** Returned by OTP2 graphql queries, but not by OTP1 */
+  id?: string;
```
daniel-heppner-ibigroup, line 206: I have been of the mind that we are removing OTP1 support entirely. Why don't we just make this field mandatory?

packages/types/src/index.ts
```diff
@@ -316,7 +334,7 @@ export type Leg = {
   rentedBike: boolean;
   rentedCar: boolean;
   rentedVehicle: boolean;
-  route?: string;
+  route?: string | LegRoute;
```
daniel-heppner-ibigroup, line 337: Is this for OTP1 support? Same as above, how do we feel about removing the string from this for 100% type compatibility with OTP2 GraphQL?

daniel-heppner-ibigroup (changes requested):
I have some thoughts on the OTP2ification of our types... I think we should discuss strategy a bit on this so that we're on the same page about OTP1 deprecation/removal and OTP2 support.

packages/types/src/index.ts
```diff
@@ -269,6 +271,22 @@ type FlexPickupBookingInfo = {
...
+  id: string;
+  longName?: string;
+  shortName: string;
+  textColor?: string;
+  // TS TODO: route type enum
+  type?: number;
+}
+
+/** Transit route attributes from itinerary legs */
+export type LegRoute = BasicRouteInfo & {
```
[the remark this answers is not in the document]
binh-dam-ibigroup: This one includes an additional `alerts` field.

binh-dam-ibigroup:
> I have some thoughts on the OTP2ification of our types... I think we should discuss strategy a bit on this so that we're on the same page about OTP1 deprecation/removal and OTP2 support.

@daniel-heppner-ibigroup I would be interested in dropping OTP1 support, but outside of a bug fix series.

daniel-heppner-ibigroup (approved):
Sounds good to me! The breaking changes can come later.

packages/types/src/index.ts
```diff
@@ -470,22 +488,15 @@ export type Agency = {
...
   fareUrl?: string;
 };
-export type Route = {
+export type Route = BasicRouteInfo & {
   agency: Agency;
   agencyId?: string | number;
   agencyName?: string | number;
-  shortName: string;
-  longName?: string;
+  bikesAllowed?: ZeroOrOne;
```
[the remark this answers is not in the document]
miles-grant-ibigroup: I think there should at least be a comment about this!

packages/itinerary-body/src/defaults/line-column-content.tsx
```diff
@@ -65,14 +65,18 @@ export default function LineColumnContent({
     }
   );
 
+  const routeNumber = typeof route === "string" ? route : route?.shortName;
```
miles-grant-ibigroup, line 68: I'm not a huge fan of calling it `routeNumber` since sometimes it will be a word or similar. How about `routeShortName`?

packages/itinerary-body/src/defaults/line-column-content.tsx
```diff
@@ -65,14 +65,18 @@ export default function LineColumnContent({
...
   return (
     <S.LegLine>
       {!isDestination && <S.InnerLine mode={mode} routeColor={routeColor} />}
       <S.LineBadgeContainer>
         {/* TODO: This is a placeholder for a routebadge when we create the transit leg */}
         {!interline && !isDestination && transitLeg && (
           <RouteBadge
-            abbreviation={toRouteAbbreviation(parseInt(route, 10) || route)}
+            abbreviation={toRouteAbbreviation(
+              parseInt(routeNumber, 10) || routeNumber
```
miles-grant-ibigroup, line 78: I think this will crash on a short name like "Circulator" or similar

miles-grant-ibigroup (changes requested):
A few potential issues!

packages/itinerary-body/src/defaults/line-column-content.tsx
binh-dam-ibigroup: This already handles route short name such as "MAX Blue Line" correctly, see http://www.opentripplanner.org/otp-ui/?path=/story/itinerarybody-otp-ui--walk-transit-walk-itinerary&globals=locale:en-US.

packages/itinerary-body/src/defaults/line-column-content.tsx
binh-dam-ibigroup: Renamed in https://github.com/opentripplanner/otp-ui/pull/632/commits/a5c09f781f22711bdfef8409f61fa5d53d1703f9.

packages/types/src/index.ts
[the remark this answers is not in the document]
binh-dam-ibigroup: I added a TODO in https://github.com/opentripplanner/otp-ui/pull/632/commits/2f4605764b574f07298c7162f2eb8d5a42913f1a.

miles-grant-ibigroup (approved):
All stories working even if I'm a bit nervous. Thanks for all the changes!

PR #634, opened 2023-08-01
Core add leg route names

binh-dam-ibigroup:
Requires #632.
This PR centralizes logic to extract a route short/long name from a Leg object.

### DIFF
--- a/packages/core-utils/src/itinerary.ts
+++ b/packages/core-utils/src/itinerary.ts
@@ -692,3 +692,11 @@ export const getLegRouteShortName = (
   }
   return route?.shortName;
 };
+
+/** Extract the route ling name for a leg returned from OTP1 or OTP2. */
+export const getLegRouteLongName = (
+  leg: Pick<Leg, "route" | "routeLongName">
+): string => {
+  const { route, routeLongName } = leg;
+  return typeof route === "string" ? routeLongName : route?.longName;
+};
--- a/packages/core-utils/src/itinerary.ts
+++ b/packages/core-utils/src/itinerary.ts
@@ -688,15 +688,18 @@ export const getLegRouteShortName = (
 ): string => {
   const { route, routeShortName } = leg;
   if (typeof route === "string") {
+    // typeof route === "string" denotes OTP1 responses.
     return typeof routeShortName === "string" ? routeShortName : route;
   }
+  // This applies to newer OTP2 responses.
   return route?.shortName;
 };
 
-/** Extract the route ling name for a leg returned from OTP1 or OTP2. */
+/** Extract the route long name for a leg returned from OTP1 or OTP2. */
 export const getLegRouteLongName = (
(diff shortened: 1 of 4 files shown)


packages/core-utils/src/itinerary.ts
```diff
@@ -681,3 +681,22 @@ export const convertGraphQLResponseToLegacy = (leg: any): any => ({
   tripHeadsign: leg.trip?.tripHeadsign,
   tripId: leg.trip?.gtfsId
 });
+
+/** Extracts the route number for a leg returned from OTP1 or OTP2. */
+export const getLegRouteShortName = (
+  leg: Pick<Leg, "route" | "routeShortName">
+): string => {
+  const { route, routeShortName } = leg;
+  if (typeof route === "string") {
```
daniel-heppner-ibigroup, line 690: Let's add a comment to this that explains this block only applies for OTP1.

packages/core-utils/src/itinerary.ts
```diff
@@ -681,3 +681,22 @@ export const convertGraphQLResponseToLegacy = (leg: any): any => ({
...
+    return typeof routeShortName === "string" ? routeShortName : route;
+  }
+  return route?.shortName;
+};
+
+/** Extract the route ling name for a leg returned from OTP1 or OTP2. */
```
daniel-heppner-ibigroup, line 696: Typo: `ling name`

daniel-heppner-ibigroup (approved):
LGTM. What do you think about adding a function that gets a route label? I think we might already have on in OTP-RR, but might be good to have here instead. 
There was a recent PR where I contributed a fix where a function like this would have been useful: 
https://github.com/opentripplanner/otp-react-redux/pull/980/commits/a05c6bd2293c6ab453a0f84c2cf3a1c096d483ea

binh-dam-ibigroup:
> LGTM. What do you think about adding a function that gets a route label?

I found a use of that logic in the fare table, so I will add it. How should that be named? `getLegRouteShortOrLongName`? `getLegRouteName`?

packages/core-utils/src/itinerary.ts
binh-dam-ibigroup: Thanks, fixed in https://github.com/opentripplanner/otp-ui/pull/634/commits/1b3839f259632ce90dad68b5149df3258e54fe4e.

binh-dam-ibigroup:
@daniel-heppner-ibigroup Let me know what you think of https://github.com/opentripplanner/otp-ui/pull/634/commits/41e4aab7b5d217d9dd58935e63e64d1edde0ed1b with the new function (and revised logic).

daniel-heppner-ibigroup (approved):
Perfect use of that logic. Thank you for the addition, this will be useful in OTP-RR and OTP-UI.

miles-grant-ibigroup (approved):
This is a great addition thanks so much

PR #638, opened 2023-08-07
fix(itinerary-body): Repair trip viewer button props

miles-grant-ibigroup:
The existing props were sending garbage data. Instead of sending stop sequence which isn't accurate anymore with OTP2, we instead send gtfs ids which are much more usable


miles-grant-ibigroup:
No longer the approach.

PR #640, opened 2023-08-08
Itinerary Body: migrate to OTP2 trip props

miles-grant-ibigroup:
The OTP1 trip props are no longer in use, and are fundamentally incompatible with the OTP2 props. This PR resolves this incongruity.


binh-dam-ibigroup (changes requested):
I smell a breaking change so it will require a breaking change commit. To avoid a breaking change, fall back on `fromIndex` and `toIndex` if `fromStopId` and `toStopId` are not provided.

packages/itinerary-body/src/TransitLegBody/view-trip-button.tsx
```diff
@@ -11,8 +11,8 @@ type Props = TripSection & {
 
 class ViewTripButton extends Component<Props> {
   onClick = (): void => {
-    const { fromIndex, setViewedTrip, toIndex, tripId } = this.props;
-    setViewedTrip({ fromIndex, toIndex, tripId });
+    const { fromStopId, setViewedTrip, toStopId, tripId } = this.props;
+    setViewedTrip({ fromStopId, toStopId, tripId });
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Yes

miles-grant-ibigroup:
I think we have to make a breaking change. OTP1 is no longer supported, so it makes no sense to support it here. I will modify the commit message

binh-dam-ibigroup (commented):
Could you update the mocks for the "Individual Leg Fare Components" to have `fromStopId` and `toStopId`?

binh-dam-ibigroup (changes requested):
See https://github.com/opentripplanner/otp-ui/pull/640#pullrequestreview-1595578597

miles-grant-ibigroup:
Updating the mocks would be a 2-release process...

packages/itinerary-body/src/TransitLegBody/view-trip-button.tsx
```diff
@@ -11,8 +11,19 @@ type Props = TripSection & {
 
 class ViewTripButton extends Component<Props> {
   onClick = (): void => {
-    const { fromIndex, setViewedTrip, toIndex, tripId } = this.props;
-    setViewedTrip({ fromIndex, toIndex, tripId });
+    const {
+      fromIndex,
+      toIndex,
```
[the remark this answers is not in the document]
binh-dam-ibigroup: This wasn't addressed.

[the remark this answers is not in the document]
miles-grant-ibigroup: See 2836d0e84d502018f665127a1a20f78ed79af8ba

binh-dam-ibigroup:
> Updating the mocks would be a 2-release process...

Why? The mocks are in the itinerary-body package.

miles-grant-ibigroup:
> > Updating the mocks would be a 2-release process...
> 
> Why? The mocks are in the itinerary-body package.

Wait sorry, I got myself confused. Mocks are updated

miles-grant-ibigroup:
I'm not sure what's going on... The new mocks aren't showing up in the GitHub diffs either, but they are showing up in my local git commit history

packages/itinerary-body/src/TransitLegBody/index.tsx
```diff
@@ -321,16 +321,12 @@ class TransitLegBody extends Component<Props, State> {
 
                   {showViewTripButton && (
                     <ViewTripButton
-                      tripId={leg.tripId}
-                      fromIndex={
-                        leg.from.stopIndex ||
-                        leg?.trip?.departureStoptime?.stopPosition
-                      }
+                      fromIndex={leg.from?.stopIndex}
+                      fromStopId={leg?.from?.stopId}
```
binh-dam-ibigroup, line 325: I think I know what is going on. The "from" and "to" stop ids come in OTP1 and OTP2 flavors. The code change in this file only cover the OTP1 response. The change you made to the mocks cover OTP2 response, which is not covered by this code, so please handle that. You need to add `leg.trip?.fromStopId` then fall back on `leg.from?.stopId`.

miles-grant-ibigroup: The problem is the actual re-writing of the response is handled by OTP-RR. I can add a `stopId` to the mock, but this wouldn't represent actual OTP2 responses. It relies on otp-rr rewriting the responses

binh-dam-ibigroup: In that case revert the mocks I made you change.

miles-grant-ibigroup: Done!

binh-dam-ibigroup (approved):
LGTM, sorry for the confusion!

packages/itinerary-body/src/TransitLegBody/index.tsx
```diff
@@ -321,16 +321,12 @@ class TransitLegBody extends Component<Props, State> {
...
                       setViewedTrip={setViewedTrip}
-                      toIndex={
-                        leg.to.stopIndex ||
-                        leg?.trip?.arrivalStoptime?.stopPosition
-                      }
+                      toIndex={leg.to?.stopIndex}
+                      toStopId={leg?.to?.stopId}
```
daniel-heppner-ibigroup, line 328: unneeded question mark

packages/itinerary-body/src/TransitLegBody/index.tsx
daniel-heppner-ibigroup, line 325: unneded question mark

daniel-heppner-ibigroup (approved):
I think the types could use some work, but this looks good... Thanks!

packages/itinerary-body/src/TransitLegBody/index.tsx
miles-grant-ibigroup: What happens if there is no `to`?

packages/itinerary-body/src/TransitLegBody/index.tsx
miles-grant-ibigroup: What happens if there is no `from`

PR #641, opened 2023-08-09
Add walk speed support to query generation

daniel-heppner-ibigroup:
This PR allows us to specify walk speed via a mode setting.

It also improves the mode setting value casting function to handle dropdowns via an elegant solution that infers the value type based on the type of the specified default value. This will help make things work smoothly in custom configurations so the end user doesn't have to worry about the type of the value.


packages/trip-form/src/SettingsSelectorPanel.story.tsx
```diff
@@ -92,8 +92,8 @@ const SettingsPanelTemplate = (args: StoryArgs, context: StoryContext) => {
         queryParamMessages={
           args.useCustomMessages
             ? {
-                walkSpeed: {
-                  label: "Custom walk speed label",
+                maxWalkDistance: {
+                  label: "Custom max walk distance label",
```
amy-corson-ibigroup, line 96: Didn't `maxWalkDistance` get removed from OTP?

amy-corson-ibigroup (approved):
Looks good! One question about the maxWalkDistance that's in the stories, but otherwise thanks so much for doing this and for removing so much old code!!

Also this works, but I think it does need a follow up PR in OTP-RR - I can't run locally without adding a conditional to `getInitialState`. Wondering if the second reviewer runs into the same issue!

packages/trip-form/src/SettingsSelectorPanel.story.tsx
daniel-heppner-ibigroup: Yes, it absolutely did, but this is just demoing the ability of this unrelated component to display custom labels. You're right that it doesn't really make sense, but updating this other component's story isn't really within the scope of this PR.

binh-dam-ibigroup (commented):
I would split this PR into multiple, starting with types. That change could be added to #646 for expediency.
The changes to core-utils are breaking, I'd recommend combining those changes with #645 which is also a breaking change PR.

daniel-heppner-ibigroup:
@binh-dam-ibigroup I don't think there was any changes to types. But for the breaking change, I can cherry pick the core-utils changes out into your PR.

binh-dam-ibigroup:
> @binh-dam-ibigroup I don't think there was any changes to types. But for the breaking change, I can cherry pick the core-utils changes out into your PR.

@daniel-heppner-ibigroup 

> @binh-dam-ibigroup I don't think there was any changes to types. But for the breaking change, I can cherry pick the core-utils changes out into your PR.

@daniel-heppner-ibigroup That's right, I see that you reverted https://github.com/opentripplanner/otp-ui/pull/641/commits/c42516469572853ebb38c77e378bf67bcba9e471. Could you go ahead and cherry-pick those changes?

daniel-heppner-ibigroup:
done, @binh-dam-ibigroup. let me know if that works. I am hoping the CI passes; the changes could cause problems in the other packages.

binh-dam-ibigroup:
> done, @binh-dam-ibigroup. let me know if that works. I am hoping the CI passes; the changes could cause problems in the other packages.

I updated the snapshots in #645.

PR #650, opened 2023-08-29
Itinerary Body Support OTP2 Place Names

miles-grant-ibigroup:
I think this is a regression... This method was only reading OTP1 type place names. This PR adds OTP2 support. A future PR will be needed to get this fix into the map components. 

This fixes rental bikes showing up as "Default vehicle type" and instead uses the company name.


daniel-heppner-ibigroup (approved):
Thanks for the fix!

AdrianaCeric (approved):
Looks good! Is there a reason why shared vehicles show up in pre-prod and not QA? 
image

miles-grant-ibigroup:
Yes @AdrianaCeric those configs are different

PR #652, opened 2023-08-30
fix(core-utils): Set agencyId when converting graphql resp to legacy.

binh-dam-ibigroup:
This PR populates the `agencyId` of an itinerary from graphql responses, for use in determining agency logos in itinerary-body.

### DIFF
--- a/packages/core-utils/src/itinerary.ts
+++ b/packages/core-utils/src/itinerary.ts
@@ -655,6 +655,7 @@ const pickupDropoffTypeToOtp1 = otp2Type => {
 export const convertGraphQLResponseToLegacy = (leg: any): any => ({
   ...leg,
   agencyBrandingUrl: leg.agency?.url,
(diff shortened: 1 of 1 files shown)


miles-grant-ibigroup (approved):
Working well for me, but @daniel-heppner-ibigroup should be second reviewer as this is his code

daniel-heppner-ibigroup (approved):
This is good, but I actually just put this function in to make the stories work. It's not actually used by OTP-RR, but this is a good reminder to add that. I'll open a PR to go along with this one.

binh-dam-ibigroup:
> 

Oh I totally missed that, but good point, could that big `processLeg` function in OTP-RR reuse this code?

daniel-heppner-ibigroup:
https://github.com/opentripplanner/otp-react-redux/pull/991 makes use of this!

PR #654, opened 2023-09-12
Allow false value in mode setting

daniel-heppner-ibigroup:
This PR allows a `falseValue` to be set in a checkbox mode setting, which will allow us to override the OTP default value for unchecked options.


miles-grant-ibigroup (approved):
Clean and it works thanks so much!

AdrianaCeric (approved):
Good job!

PR #655, opened 2023-09-13
Nearby view support

daniel-heppner-ibigroup:
This PR adjusts all the places where the stop viewer is called to pass in the whole stop object rather than just the stop ID string. This allows support for the nearby view in OTP-RR, which relies on the lat/lon of the stop rather than its ID.


miles-grant-ibigroup (approved):
Could use some improvement in the future but it's good enough for now!

binh-dam-ibigroup:
@miles-grant-ibigroup @daniel-heppner-ibigroup Is this one of these PRs that need to be split, especially for the types package, so that a version number is available for other packages to use as reference?

packages/itinerary-body/src/otp-react-redux/view-stop-button.tsx
```diff
@@ -1,18 +1,19 @@
...
+import { Stop } from "@opentripplanner/types";
 import { FormattedMessage } from "react-intl";
 
 import * as S from "../styled";
 import { defaultMessages } from "../util";
 
 interface Props {
-  onStopClick: ({ stopId: string }) => void;
-  stopId: string;
+  onStopClick: (stop: Stop) => void;
```
binh-dam-ibigroup, line 9: This is a breaking change. Either:
1. Add breaking change commits to all packages that use this new method signature, or,
2. If maintaining compatibility, change the signature of `onStopClick` to `(stop: Stop | { stopId: string })` (You could create an alias `type StopEventHanlder = (stop: Stop | { stopId: string }) => void` in the types package to avoid repetition.)

daniel-heppner-ibigroup: That's a great idea! I'll add it. Thank you for the suggestion

daniel-heppner-ibigroup:
Merge conflicts resolved, ready for another look @binh-dam-ibigroup

binh-dam-ibigroup (changes requested):
Could you address the `tsc` error during `yarn build`?

daniel-heppner-ibigroup:
That's embarrassing. I assigned it back to you before I finished my commits.

packages/endpoints-overlay/package.json
```diff
@@ -19,17 +19,17 @@
...
+    "@opentripplanner/core-utils": "^11.1.2",
     "flat": "^5.0.2",
     "@styled-icons/fa-solid": "^10.34.0"
   },
   "devDependencies": {
     "@types/flat": "^5.0.2"
   },
   "peerDependencies": {
-    "@opentripplanner/types": "^6.1.0",
+    "@opentripplanner/types": "^6.1.1",
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Update all references to the `types` package to 6.2.0.

packages/itinerary-body/src/otp-react-redux/view-stop-button.tsx
```diff
@@ -1,18 +1,19 @@
 import React, { Component, ReactElement } from "react";
+import { Stop, StopEventHandler } from "@opentripplanner/types";
 import { FormattedMessage } from "react-intl";
 
 import * as S from "../styled";
 import { defaultMessages } from "../util";
 
 interface Props {
-  onStopClick: ({ stopId: string }) => void;
-  stopId: string;
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: Made it an either or type, let me know if that's good

daniel-heppner-ibigroup:
I can't update the Map Overlay package first because it breaks the types in other packages, since the callback can now return a Stop. I think that means a breaking change is unavoidable in Map Overlay and they have to be released with alpha packages.

Sidenote, this is why upgrading to a more modern package manager such as yarn 3 or pnpm would be advantageous, since you can put `workspace:` as the package version in your monorepo packages.

packages/itinerary-body/src/otp-react-redux/view-stop-button.tsx
```diff
@@ -5,15 +5,23 @@ import { FormattedMessage } from "react-intl";
 import * as S from "../styled";
 import { defaultMessages } from "../util";
 
-interface Props {
+interface PropsWithStopId {
```
binh-dam-ibigroup, line 8: I think you should be able to combine these two types, with the check for `"gtfsId" in stop` be sufficient. 
```
interface Props {
  onStopClick: ...;
  stopId?: string;
  stop?: Stop;
}
```

binh-dam-ibigroup (approved):
I think it is good to go with one tiny type tweak.

binh-dam-ibigroup:
> I think that means a breaking change is unavoidable in Map Overlay and they have to be released with alpha packages.

That's fine, but then upgrade to the non-alpha versions soon after they are released.

packages/itinerary-body/package.json
```diff
@@ -10,9 +10,9 @@
   "license": "MIT",
   "private": false,
   "dependencies": {
-    "@opentripplanner/core-utils": "^11.0.2",
+    "@opentripplanner/core-utils": "^11.1.3",
     "@opentripplanner/humanize-distance": "^1.2.0",
-    "@opentripplanner/icons": "^2.0.5",
+    "@opentripplanner/icons": "^2.0.6",
```
miles-grant-ibigroup, line 15: Can we please do a merge-in of master and then another update-internal-dependencies!

miles-grant-ibigroup (approved):
Off to the races 🫡

PR #656, opened 2023-09-14
fix(printable-itinerary): Use latest itinerary-body, fix route name render

binh-dam-ibigroup:
This PR makes the printable itinerary print correct route numbers and headsigns with OTP2 responses.


daniel-heppner-ibigroup (approved):
Nice!

amy-corson-ibigroup (approved):

PR #657, opened 2023-09-14
Cleanup yarn.lock, Remove IE11

miles-grant-ibigroup:
This PR removes IE11 support. OTP-UI has not worked with IE11 for a while now, but for some reason we still had some compatibility code in there. 

This PR also cleans some other dependencies and runs yarn deduplicate, which should help with bundle sizes

### DIFF
--- a/packages/core-utils/src/ui.ts
+++ b/packages/core-utils/src/ui.ts
@@ -1,19 +1,9 @@
-import bowser from "bowser";
-
 export function isMobile(): boolean {
   // TODO: consider using 3rd-party library?
   return /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(
     navigator.userAgent
   );
 }
-
-/**
- * Returns true if the user is using a [redacted] browser
- */
-export function isIE(): boolean {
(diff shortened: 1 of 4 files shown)


AdrianaCeric (approved):
Internet Explorer 11 is so vintage. I think there is just a duplicate mock to sort out

packages/core-utils/src/ui.ts
```diff
@@ -1,19 +1,9 @@
...
   // TODO: consider using 3rd-party library?
   return /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(
     navigator.userAgent
   );
 }
-
-/**
- * Returns true if the user is using a [redacted] browser
- */
-export function isIE(): boolean {
```
binh-dam-ibigroup, line 13: Although this is a breaking change, we can market that IE is no longer supported as of core-utils 11.0.0.

binh-dam-ibigroup (changes requested):
Why are tests failing?

miles-grant-ibigroup:
Looks like deduplicate is causing a package to update that is breaking story shot. Will revert

binh-dam-ibigroup (approved):

PR #658, opened 2023-09-20
Configure driving directions

amy-corson-ibigroup:
Removes step-by-step driving directions from the itinerary body if `hideDrivingDirections` is enabled in the configuration.

### DIFF
--- a/packages/itinerary-body/package.json
+++ b/packages/itinerary-body/package.json
@@ -24,7 +24,7 @@
     "string-similarity": "^4.0.4"
   },
   "devDependencies": {
-    "@opentripplanner/types": "^6.1.0",
+    "@opentripplanner/types": "7.0.0-alpha.11",
     "@types/flat": "^5.0.2"
   },
   "peerDependencies": {
--- a/packages/itinerary-body/package.json
+++ b/packages/itinerary-body/package.json
@@ -24,7 +24,7 @@
     "string-similarity": "^4.0.4"
   },
   "devDependencies": {
-    "@opentripplanner/types": "7.0.0-alpha.11",
+    "@opentripplanner/types": "^6.1.0",
     "@types/flat": "^5.0.2"
   },
   "peerDependencies": {
--- a/packages/itinerary-body/src/AccessLegBody/index.tsx
+++ b/packages/itinerary-body/src/AccessLegBody/index.tsx
@@ -77,6 +77,9 @@ class AccessLegBody extends Component<Props, State> {
     } = this.props;
     const { expanded } = this.state;
 
+    const hideDrivingDirections =
+      config?.itinerary?.hideDrivingDirections && leg.mode === "CAR";
+
     if (leg.mode === "CAR" && leg.rideHailingEstimate) {
       return (
         <TNCLeg
@@ -111,38 +114,47 @@ class AccessLegBody extends Component<Props, State> {
           <S.LegDetails>
             <S.StepsHeaderAndMapLink>
               <S.StepsHeader
-                aria-expanded={expanded}
-                onClick={this.onStepsHeaderClick}
+                aria-expanded={!hideDrivingDirections ? expanded : undefined}
+                as={hideDrivingDirections && "span"}
+                autoCursor={hideDrivingDirections}
+                onClick={!hideDrivingDirections && this.onStepsHeaderClick}
               >
                 <Duration seconds={leg.duration} />
-                {leg.steps && <S.CaretToggle expanded={expanded} />}
-                <S.InvisibleAdditionalDetails>
-                  <FormattedMessage
-                    defaultMessage={
-                      defaultMessages["otpUi.TransitLegBody.expandDetails"]
-                    }
-                    description="Screen reader text added to expand steps"
-                    id="otpUi.TransitLegBody.expandDetails"
-                  />
-                </S.InvisibleAdditionalDetails>
+                {leg.steps && !hideDrivingDirections && (
+                  <S.CaretToggle expanded={expanded} />
+                )}
+                {!hideDrivingDirections && (
+                  <S.InvisibleAdditionalDetails>
+                    <FormattedMessage
+                      defaultMessage={
+                        defaultMessages["otpUi.TransitLegBody.expandDetails"]
+                      }
+                      description="Screen reader text added to expand steps"
+                      id="otpUi.TransitLegBody.expandDetails"
+                    />
+                  </S.InvisibleAdditionalDetails>
+                )}
               </S.StepsHeader>
               <MapillaryButton
                 clickCallback={mapillaryCallback}
                 coords={leg.from}
+                hideDrivingDirections={hideDrivingDirections}
                 mapillaryKey={mapillaryKey}
               />
             </S.StepsHeaderAndMapLink>
-            <AnimateHeight
-              duration={500}
-              height={expanded ? "auto" : 0}
-              style={{ gridColumn: "1 / span 2" }}
-            >
-              <AccessLegSteps
-                mapillaryCallback={mapillaryCallback}
-                mapillaryKey={mapillaryKey}
-                steps={leg.steps}
-              />
-            </AnimateHeight>
+            {!hideDrivingDirections && (
+              <AnimateHeight
+                duration={500}
+                height={expanded ? "auto" : 0}
+                style={{ gridColumn: "1 / span 2" }}
+              >
+                <AccessLegSteps
+                  mapillaryCallback={mapillaryCallback}
+                  mapillaryKey={mapillaryKey}
+                  steps={leg.steps}
+                />
+              </AnimateHeight>
+            )}
             <LegDiagramPreview
               diagramVisible={diagramVisible}
               leg={leg}
--- a/packages/itinerary-body/src/AccessLegBody/index.tsx
+++ b/packages/itinerary-body/src/AccessLegBody/index.tsx
@@ -17,7 +17,11 @@ import TNCLeg from "./tnc-leg";
 import { defaultMessages } from "../util";
 
 interface Props {
-  config: Config;
+  config: Config & {
+    itinerary?: {
+      hideDrivingDirections?: boolean;
+    };
+  };
   /**
    * Should be either null or a legType. Indicates that a particular leg diagram
    * has been selected and is active.
@@ -93,6 +97,13 @@ class AccessLegBody extends Component<Props, State> {
       );
     }
 
+    const Mapillary = () => (
+      <MapillaryButton
+        clickCallback={mapillaryCallback}
+        coords={leg.from}
+        mapillaryKey={mapillaryKey}
+      />
+    );
     return (
       <>
         {/* Place subheading: rented vehicle (e.g., scooter, bike, car)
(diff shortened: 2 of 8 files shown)


packages/itinerary-body/package.json
```diff
@@ -24,7 +24,7 @@
     "string-similarity": "^4.0.4"
   },
   "devDependencies": {
-    "@opentripplanner/types": "^6.1.0",
+    "@opentripplanner/types": "7.0.0-alpha.11",
```
binh-dam-ibigroup, line 27: What is this change for?

packages/itinerary-body/src/AccessLegBody/index.tsx
```diff
@@ -111,38 +114,47 @@ class AccessLegBody extends Component<Props, State> {
           <S.LegDetails>
             <S.StepsHeaderAndMapLink>
               <S.StepsHeader
-                aria-expanded={expanded}
-                onClick={this.onStepsHeaderClick}
+                aria-expanded={!hideDrivingDirections ? expanded : undefined}
+                as={hideDrivingDirections && "span"}
+                autoCursor={hideDrivingDirections}
+                onClick={!hideDrivingDirections && this.onStepsHeaderClick}
```
binh-dam-ibigroup, line 120: Please use a `?` ternary to separate the button mode vs. the span mode, so that we don't end up with all of these conditional props around.

packages/itinerary-body/src/AccessLegBody/index.tsx
```diff
@@ -77,6 +77,9 @@ class AccessLegBody extends Component<Props, State> {
     } = this.props;
     const { expanded } = this.state;
 
+    const hideDrivingDirections =
+      config?.itinerary?.hideDrivingDirections && leg.mode === "CAR";
```
binh-dam-ibigroup, line 81: Quite frankly, I hate passing the config like that, but because that's already done elsewhere in this package I won't hold this against this PR.

packages/itinerary-body/src/AccessLegBody/mapillary-button.tsx
```diff
@@ -27,7 +27,8 @@ const generateBoundingBoxFromCoordinate = ({
   return [west, south, east, north];
 };
 
-const Container = styled.a`
+const Container = styled.a<{ hideDrivingDirections: boolean }>`
+  margin-left: ${props => (props.hideDrivingDirections ? ".4em" : "0")};
```
binh-dam-ibigroup, line 31: To avoid passing a prop, add an after margin to the steps header span instead of setting a margin here.

binh-dam-ibigroup (changes requested):
I think the changes would be much easier to read if separating the component and render logic when hiding driving directions vs. not (see comments).

packages/itinerary-body/package.json
amy-corson-ibigroup: Adding `itinerary` to the Config type. It wasn't being recognized locally.

packages/itinerary-body/src/AccessLegBody/index.tsx
amy-corson-ibigroup: Much cleaner, thank you! Fixed in [53a5c5f](https://github.com/opentripplanner/otp-ui/pull/658/commits/53a5c5f9b01fdd1b78e0a02d13743805c600454b)

packages/itinerary-body/src/AccessLegBody/mapillary-button.tsx
amy-corson-ibigroup: Good call, fixed in [53a5c5f](https://github.com/opentripplanner/otp-ui/pull/658/commits/53a5c5f9b01fdd1b78e0a02d13743805c600454b)

packages/itinerary-body/package.json
binh-dam-ibigroup: Why isn't there a non-beta package that includes the definition?

packages/itinerary-body/src/AccessLegBody/index.tsx
```diff
@@ -109,40 +112,57 @@ class AccessLegBody extends Component<Props, State> {
...
+              </S.StepsHeaderAndMapLink>
+            ) : (
+              <S.StepsHeaderAndMapLink>
+                <S.StepsHeader
+                  aria-expanded={expanded}
+                  onClick={this.onStepsHeaderClick}
+                >
+                  <Duration seconds={leg.duration} />
+                  {leg.steps && <S.CaretToggle expanded={expanded} />}
+                  {!hideDrivingDirections && (
```
binh-dam-ibigroup, line 134: This condition is now redundant, you can remove it.

packages/itinerary-body/src/AccessLegBody/index.tsx
```diff
@@ -109,40 +112,57 @@ class AccessLegBody extends Component<Props, State> {
...
+                </S.StepsHeader>
+                <MapillaryButton
+                  clickCallback={mapillaryCallback}
+                  coords={leg.from}
+                  mapillaryKey={mapillaryKey}
+                />
+              </S.StepsHeaderAndMapLink>
+            )}
+            {!hideDrivingDirections && (
+              <AnimateHeight
```
binh-dam-ibigroup, line 154: Move `<AnimateHeight>` to under the `:` portion of the previous statement.

packages/itinerary-body/src/AccessLegBody/index.tsx
```diff
@@ -109,40 +112,57 @@ class AccessLegBody extends Component<Props, State> {
...
+            {hideDrivingDirections ? (
+              <S.StepsHeaderAndMapLink>
+                <S.StepsHeader as="span" span>
+                  <Duration seconds={leg.duration} />
+                </S.StepsHeader>
+                <MapillaryButton
+                  clickCallback={mapillaryCallback}
+                  coords={leg.from}
+                  mapillaryKey={mapillaryKey}
+                />
```
binh-dam-ibigroup, line 124: Extract a variable for the Mapillary button (it appears twice in this render)

packages/itinerary-body/src/styled.tsx
```diff
@@ -588,10 +588,14 @@ export const StepsHeaderAndMapLink = styled.span`
   }
 `;
 
-export const StepsHeader = styled(TransparentButton)`
+export const StepsHeader = styled(TransparentButton)<{
+  span?: boolean;
```
binh-dam-ibigroup, line 592: I would create separate `StepHeaderSpan` and `StepHeaderButton` components and apply the common and specific styles accordingly.

packages/itinerary-body/package.json
binh-dam-ibigroup: Instead of using a beta package, extend the Config type in the props type at `packages/itinerary-body/src/AccessLegBody/index.tsx` line 19 by writing something like:
```js
interface Props {
  config: Config & {
     itinerary?: {
       hideDrivingDirections?: boolean
     }
  }
  ...
}
```

binh-dam-ibigroup (changes requested):
Could you take care of some more cleanup, please.

packages/itinerary-body/src/AccessLegBody/index.tsx
amy-corson-ibigroup: Whoops, good catch thank you! Fixed in [a2374ac](https://github.com/opentripplanner/otp-ui/pull/658/commits/a2374ac0508f42fa387e4659b5fd306722c96f40)

packages/itinerary-body/src/AccessLegBody/index.tsx
```diff
@@ -109,40 +123,48 @@ class AccessLegBody extends Component<Props, State> {
...
-                <Duration seconds={leg.duration} />
-                {leg.steps && <S.CaretToggle expanded={expanded} />}
-                <S.InvisibleAdditionalDetails>
-                  <FormattedMessage
-                    defaultMessage={
-                      defaultMessages["otpUi.TransitLegBody.expandDetails"]
-                    }
-                    description="Screen reader text added to expand steps"
-                    id="otpUi.TransitLegBody.expandDetails"
+            {hideDrivingDirections ? (
```
binh-dam-ibigroup, line 126: Forgot to mention that, could you add a story where the driving directions are hidden?

packages/itinerary-body/src/AccessLegBody/index.tsx
```diff
@@ -90,6 +97,13 @@ class AccessLegBody extends Component<Props, State> {
       );
     }
 
+    const Mapillary = () => (
```
binh-dam-ibigroup, line 100: Just make this a variable.
```js
const mapillary = (
  <MapillaryButton ... />
)
```

binh-dam-ibigroup: Then write:
```js
<SomeContainer>
  {mapillary}
</SomeContainer>
```

packages/itinerary-body/src/styled.tsx
```diff
@@ -588,13 +588,22 @@ export const StepsHeaderAndMapLink = styled.span`
...
   font-size: 13px;
   font-style: normal;
   padding: 0;
 `;
 
+export const StepsHeaderButton = styled(TransparentButton)`
+  ${stepsHeaderStyling}
+`;
+
+export const StepsHeaderSpan = styled.span`
```
binh-dam-ibigroup, line 602: Yes this is much cleaner.

binh-dam-ibigroup (changes requested):
Sorry, I should have requested the additional story before. See also the cleanup comment.

packages/itinerary-body/src/AccessLegBody/index.tsx
amy-corson-ibigroup: Added in [16d5df4](https://github.com/opentripplanner/otp-ui/pull/658/commits/16d5df44cab0d972e67efa40eb36cae46829e831)!

packages/itinerary-body/src/AccessLegBody/index.tsx
amy-corson-ibigroup: Added in [c0931b3](https://github.com/opentripplanner/otp-ui/pull/658/commits/c0931b308342f6fd92ac7dfc8192b44a20dda5c9)!

binh-dam-ibigroup (approved):
Looks good, thanks for the changes!

binh-dam-ibigroup:
@amy-corson-ibigroup Please remember to update the snapshots.

packages/itinerary-body/src/AccessLegBody/index.tsx
AdrianaCeric: Hi Binh, I'm also doing a configurable change within otp-ui. What is the preferred method to pass the config?

packages/itinerary-body/src/AccessLegBody/index.tsx
```diff
@@ -17,7 +17,11 @@ import TNCLeg from "./tnc-leg";
 import { defaultMessages } from "../util";
 
 interface Props {
-  config: Config;
+  config: Config & {
+    itinerary?: {
+      hideDrivingDirections?: boolean;
+    };
+  };
```
daniel-heppner-ibigroup, line 24: I think you should add this to the `config` type in OTP-UI rather than adding it here so it's centralized!

daniel-heppner-ibigroup: @binh-dam-ibigroup what do you think? You're adding typescript to the config so maybe this point is mute

packages/itinerary-body/src/AccessLegBody/index.tsx
daniel-heppner-ibigroup: I'm guessing it's preferred to just pass the individual options in rather than use the whole config object.

daniel-heppner-ibigroup (approved):
LGTM, thank you for adding the stories too!

packages/itinerary-body/src/AccessLegBody/index.tsx
amy-corson-ibigroup: Spoke with Binh about this offline, and I think we'll deal with config typing in subsequent PRs if that's alright!

binh-dam-ibigroup: At some point, I would actually prefer to remove passing a config object and pass individual props instead. (That would be a breaking change)

PR #659, opened 2023-09-20
Typeahead Results on Voiceover

amy-corson-ibigroup:
Adjust markup for location field so that typeahead results from geocoder are read to AT (specifically voiceover)

Note for testing that Voiceover is only officially supported on safari!!


miles-grant-ibigroup (approved):
All is working! You know a lot more about what's going on here than I do

PR #660, opened 2023-09-26
Update Geocoder to match latest OTP API

miles-grant-ibigroup:
The OTP api has changed the response of the geocoder. This PR accounts for that. Breaking change. Older OTP versions are not supported


daniel-heppner-ibigroup (approved):
Interesting, thanks for the fix

amy-corson-ibigroup (approved):

PR #661, opened 2023-09-27
Align fare table content

AdrianaCeric:
A quick change to the alignment of the contents inside the table to allow for equal spacing between prices and the edge of the table cells. 

To test: 
- Storybook --> TripDetails --> Leg Fare Products Itinerary
- View on any configuration in otp-rr

Before
image
After
image

### DIFF
--- a/packages/trip-details/src/fare-table.tsx
+++ b/packages/trip-details/src/fare-table.tsx
@@ -50,7 +50,7 @@ const Table = styled.table`
   padding: 0;
 
   td {
-    text-align: right;
+    text-align: center;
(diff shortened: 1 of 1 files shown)


daniel-heppner-ibigroup:
I am not sure if this is something that should be made universal. I think I'd rather see this be configurable, since the original design choice was made intentionally to help people differentiate between the total and the per leg fares. 

Do you think this is something we could override via stylesheet in the OTP-RR deployment where it's desired? Maybe we should just add some class names to the items we want to be able to override.

PR #662, opened 2023-09-28
fix(core-utils): replace route id with gtfs id

amy-corson-ibigroup:
Replaces the randomly generated route ID with `gtfsId`, in order to apply things like `routeModeOverrides` to itineraries.

I could not think of a reason the randomly generated id shouldn't be replaced, and itineraries look fine in testing but please let me know if the id should be kept and a new `gtfsId` field should be added.

### DIFF
--- a/packages/core-utils/src/planQuery.graphql
+++ b/packages/core-utils/src/planQuery.graphql
@@ -157,7 +157,7 @@ query Plan(
             id
           }
           color
-          id
+          id: gtfsId
           longName
           shortName
           textColor


daniel-heppner-ibigroup (changes requested):
Can we keep `id` and just add `gtfsId` separately? That is the convention we've been using elsewhere. If you want to rewrite it for now it can happen in the `convertGraphQLResponseToLegacy` function in `itinerary.ts`.

amy-corson-ibigroup:
Yes! Closing in favor of #667

PR #663, opened 2023-09-28
Building Blocks - Color Palettes

amy-corson-ibigroup:
Start Building Blocks repo, and initialize color variables


packages/building-blocks/src/stories/ColorPalette.tsx
```diff
@@ -0,0 +1,52 @@
...
+  &:first-of-type {
+    border-radius: 8px 8px 0 0;
+  }
+  &:last-of-type {
+    border-radius: 0 0 8px 8px;
+  }
+`;
+
+const ColorPalette = ({ color }: Props): JSX.Element => {
+  const colorArray: Array<Hue> = Object.entries(color);
```
miles-grant-ibigroup, line 35: Total nitpick, but this might be cleaner

```suggestion
  const [weight, hex]: Array<Hue> = Object.entries(color);
```

miles-grant-ibigroup (approved):
This is really exciting! Great work, I love the styling on the color viewer, I love how clean this is. Excited to use it!

packages/building-blocks/src/colors/blue.js
```diff
@@ -0,0 +1,16 @@
+const blue = {
+  50: "#E5ECF3",
+  100: "#CCDAE7",
+  200: "#B2C7DB",
+  300: "#99B5CF",
+  400: "#7FA2C2",
+  500: "#6690B7",
+  600: "#4D7EAB",
+  // 700 and above text is WCAG AA accessible on white backgrounds.
```
daniel-heppner-ibigroup, line 9: I wonder if there's a way to integrate this comment into the storybook somehow. The stories themselves should provide all the documentation needed for someone to use these colors, including code samples!

daniel-heppner-ibigroup (changes requested):
I love this, it's a great start and probably the start of something very good for us. I am a little worried about how this is going to work when we try to import them. I think the library needs an `index.ts` file that is used in the import in a consumer. 

Also, can we make the color files use the `.ts` extension? I think the code can just stay the same, but with a `ts` extension the typescript compiler will generate typescript bindings for it.

packages/building-blocks/src/colors/blue.js
daniel-heppner-ibigroup: This is something that can come later, I'd be happy to get this merged in for now!

packages/building-blocks/src/stories/ColorPalette.tsx
```diff
@@ -0,0 +1,52 @@
+import React from "react";
+
+import styled from "styled-components";
+import { getMostReadableTextColor } from "@opentripplanner/core-utils/lib/route";
+
+interface Hue {
```
daniel-heppner-ibigroup, line 6: This type can probably be exported and then used as the type in the different color files.

daniel-heppner-ibigroup (commented):
one more comment

packages/building-blocks/src/stories/ColorPalette.tsx
amy-corson-ibigroup: Is there a way to do this and still `map` this array on 39?

daniel-heppner-ibigroup (approved):
Great work! Let's start using them!

PR #664, opened 2023-09-29
fix(mode-selector): Fix vert alignment of transit modes

binh-dam-ibigroup:
This PR fixes vertical alignment of transit mode icons and text when a mode name that spans over two lines is next to a mode with a 1-line name.
This PR also fixes the expand button size for keyboard access.

| Before | After |
|---|---|
| before screenshot | after screenshot |

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -6564,11 +6564,6 @@ bottleneck@^2.18.1:
   resolved "https://registry.yarnpkg.com/bottleneck/-/bottleneck-2.19.5.tgz#5df0b90f59fd47656ebe63c78a98419205cadd91"
   integrity sha512-VHiNCbI1lKdl44tGrhNfU3lup0Tj/ZBMJB5/2ZbNXRCPuRCO7ed2mgcK4r17y+KB2EfuYuRaVlwNbAeaWGSpbw==
 
-bowser@^2.7.0:
-  version "2.11.0"
-  resolved "https://registry.yarnpkg.com/bowser/-/bowser-2.11.0.tgz#5ca3c35757a7aa5771500c70a73a9f91ef420a8f"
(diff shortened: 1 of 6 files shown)


yarn.lock
```diff
@@ -6564,11 +6564,6 @@ bottleneck@^2.18.1:
   resolved "https://registry.yarnpkg.com/bottleneck/-/bottleneck-2.19.5.tgz#5df0b90f59fd47656ebe63c78a98419205cadd91"
   integrity sha512-VHiNCbI1lKdl44tGrhNfU3lup0Tj/ZBMJB5/2ZbNXRCPuRCO7ed2mgcK4r17y+KB2EfuYuRaVlwNbAeaWGSpbw==
 
-bowser@^2.7.0:
```
amy-corson-ibigroup, line 6567: What is this?

amy-corson-ibigroup (approved):
Looks good! Thank you for addressing the focus button as well, looks much cleaner!

yarn.lock
AdrianaCeric: Yeah I get this too. Would it make more sense to update this in a separate PR?

AdrianaCeric (changes requested):
The alignment is a big improvement, thanks!
One note: When an option doesn't have an icon, the text is farther from the checkbox than the options with an icon
image

binh-dam-ibigroup:
> One note: When an option doesn't have an icon, the text is farther from the checkbox than the options with an icon image

Good observation. I have fixed that in https://github.com/opentripplanner/otp-ui/pull/664/commits/ceaf0725aa88b35e76a44e0803601b7a76da9775 while at it.

AdrianaCeric (approved):
Looks great thanks for the fixes!

PR #668, opened 2023-10-04
Make Wheelchair Accessibility Badges accessible to AT

amy-corson-ibigroup:
Provides invisible labeling for wheelchair accessibility badges so that users with AT can identify whether a leg or itinerary is wheelchair accessible.

### DIFF
--- a/packages/itinerary-body/i18n/en-US.yml
+++ b/packages/itinerary-body/i18n/en-US.yml
@@ -81,6 +81,16 @@ otpUi:
       car: by car
       escooter: by e-scooter
       walk: by walking
+    tripAccessibility:
+      tripAccessibilityUnclear: >
+        The wheelchair accessibility for this {leg, select, true {leg of your
+        itinerary} other {itinerary} } is unknown.
+      tripIsAccessible: >
+        This {leg, select, true {leg of your itinerary} other {itinerary} } is
+        likely wheelchair accessible.
+      tripIsInaccessible: >
+        This {leg, select, true {leg of your itinerary} other {itinerary} } is
+        wheelchair inaccessible.
     viewOnMap: View on map
   TransitLegBody:
     AlertsBody:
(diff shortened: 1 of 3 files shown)


packages/itinerary-body/src/ItineraryBody/accessibility-rating.tsx
```diff
@@ -50,17 +55,28 @@ interface Props {
 const AccessibilityRating = ({
   gradationMap,
   grayscale = false,
+  intl,
```
binh-dam-ibigroup, line 58: Instead of passing `intl` as a prop, use the hook `useIntl` to access the `intl` context from within the component without the need for a breaking change.

binh-dam-ibigroup (changes requested):
We don't need a breaking change, see my comment regarding accessing `intl`. Could do either revert the commit that carries the breaking change, or create the PR anew without the breaking change commit (a fix commit should be enough).

packages/itinerary-body/i18n/en-US.yml
```diff
@@ -81,6 +81,19 @@ otpUi:
       car: by car
       escooter: by e-scooter
       walk: by walking
+    tripAccessibility:
+      tripAccessibilityUnclear: >
```
binh-dam-ibigroup, line 85: Maybe we can use some default labels, however the gradation map in the configuration already comes with language-specific text. How about we fall back on the text in this file if no text is provided in the configuration?

amy-corson-ibigroup: That's fine! Could we update the configuration with some of this language? I'm worried for the majority of our use cases, phrasing like "mostly accessible" isn't going to be clear to AT what that's referring to.

binh-dam-ibigroup: Yes, we should update the configurations.

PR #670, opened 2023-10-11
Get real agency id in graphql plan request

miles-grant-ibigroup:
The agency id we get at the moment is generated. This PR uses the real one instead


philip-cline (approved):

daniel-heppner-ibigroup (approved):
This is bad

PR #672, opened 2023-10-17
Allow multiple modes selected from one checkbox

AdrianaCeric:
Allows multiple modes to be selected in one checkbox for mode selector


daniel-heppner-ibigroup (changes requested):
This is good, although we don't urgently need this for anything, it would be nice to have it merged in. I realized that the types package change needs to be put into its own PR so that we can get a new stable version of it. 

You can cherry pick that commit out to its own branch and open a PR for it.

miles-grant-ibigroup:
@daniel-heppner-ibigroup are you ok to handle this

PR #673, opened 2023-10-18
fix(ItineraryBody): fix crash on no latestBookingTime

daniel-heppner-ibigroup:
This fixes a crash happening sometimes when there is no latestBookingTime.


binh-dam-ibigroup (approved):

miles-grant-ibigroup (approved):
OK for now...

PR #671, opened 2023-10-17
Metro Mode Selector Spacing Adjustments

miles-grant-ibigroup:
Also includes master branch repairs

PR #674, opened 2023-10-19
fix(core-utils): fix duplicate gtfsId

amy-corson-ibigroup:
planQuery having two gtfsIds is throwing an error


miles-grant-ibigroup (changes requested):
Duplicate of #671

PR #676, opened 2023-10-31
fix(trip-details): show fares of $0

daniel-heppner-ibigroup:
This is a fix that enables the fare table to function correctly when fares are $0. Previously these were assumed to be equal to undefined, and they were hidden. But we now have the ability to pass back actual undefined fares in the API, so it's time to correct this in the UI.

### DIFF
--- a/packages/building-blocks/package.json
+++ b/packages/building-blocks/package.json
@@ -12,7 +12,7 @@
     "mapbox-gl": "npm:empty-npm-package@1.0.0",
     "maplibre-gl": "^2.1.9",
     "react-map-gl": "^7.0.15",
-    "@opentripplanner/core-utils": "9.0.3"
+    "@opentripplanner/core-utils": "11.1.2"
   },
   "peerDependencies": {
     "@opentripplanner/types": "^6.1.0",
--- a/packages/trip-details/src/fare-table.tsx
+++ b/packages/trip-details/src/fare-table.tsx
@@ -89,7 +89,7 @@ const FareTypeTable = ({
       ...col,
       total: getItineraryCost(legs, col.mediumId, col.riderCategoryId)
     }))
-    .filter(col => col.total);
+    .filter(col => col.total !== undefined);
 
   const headerString = useGetHeaderString(headerKey);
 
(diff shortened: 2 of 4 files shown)


packages/trip-details/src/fare-table.tsx
```diff
@@ -89,7 +89,7 @@ const FareTypeTable = ({
       ...col,
       total: getItineraryCost(legs, col.mediumId, col.riderCategoryId)
     }))
-    .filter(col => col.total);
+    .filter(col => col.total !== undefined);
```
binh-dam-ibigroup, line 92: Should there be a story to illustrate this new case?

binh-dam-ibigroup (changes requested):
Please remove that reference to the old core-utils version.

packages/trip-details/src/fare-table.tsx
daniel-heppner-ibigroup: I thought about it, but I think it's too small of a corner case and adding another story for it would require another itinerary, so I don't think it's worth it.

yarn.lock
```diff
@@ -3241,6 +3241,25 @@
   resolved "https://registry.yarnpkg.com/@open-draft/until/-/until-1.0.3.tgz#db9cc719191a62e7d9200f6e7bab21c5b848adca"
   integrity sha512-Aq58f5HiWdyDlFffbbSjAlv596h/cOnt2DO1w3DOC7OJ5EHs0hd/nycJfiu9RJbT6Yk6F1knnRRXNSpxoIVZ9Q==
 
+"@opentripplanner/core-utils@9.0.3":
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: Looks like it came from building blocks! I fixed the dep

packages/building-blocks/package.json
```diff
@@ -12,7 +12,7 @@
     "mapbox-gl": "npm:empty-npm-package@1.0.0",
     "maplibre-gl": "^2.1.9",
     "react-map-gl": "^7.0.15",
-    "@opentripplanner/core-utils": "9.0.3"
+    "@opentripplanner/core-utils": "11.1.2"
```
binh-dam-ibigroup, line 15: For a next time, try to not include changes to unrelated packages in PRs.

binh-dam-ibigroup (approved):

packages/building-blocks/package.json
daniel-heppner-ibigroup: This is just here because it was causing the old core-utils version to show up in yarn.lock!

miles-grant-ibigroup (approved):
Does this handle `null` fares?

PR #681, opened 2023-11-08
Nearby view support (types)

daniel-heppner-ibigroup:
This PR adds the necessary properties to the types for the nearby view.

### DIFF
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -238,9 +238,22 @@ export type Place = {
   name: string;
   networks?: string[];
   rentalVehicle?: { network: string };
+  stop?: Stop;
+  /**
+   * @deprecated Only for OTP1 support, removal is immenent
+   */
   stopCode?: string;
+  /**
+   * @deprecated Only for OTP1 support, removal is immenent
+   */
   stopId?: string;
+  /**
+   * @deprecated Only for OTP1 support, removal is immenent
+   */
   stopIndex?: number;
+  /**
+   * @deprecated Only for OTP1 support, removal is immenent
+   */
   stopSequence?: number;
   vertexType: string;
   zoneId?: string;
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -473,6 +473,8 @@ export type StopLayerStop = LayerEntity & {
   name: string;
 };
 
+export type StopEventHanlder = (stop: Stop | { stopId: string }) => void;
+
 /**
  * This models data about a stop and it's associated routes that is obtained
  * from a transit index API.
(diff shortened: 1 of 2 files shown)


daniel-heppner-ibigroup:
I'm thinking maybe instead of `gtfsId` we should just use `id` and rename the property in the query. We seem to have a mix though in the app, and should probably standardize. Thoughts?

packages/types/src/index.ts
```diff
@@ -473,10 +486,11 @@ export type Stop = {
   dist?: number;
   geometries?: { geoJson?: GeoJSON.Polygon };
   id: string;
-  lat: number;
-  lon: number;
+  lat?: number;
+  lon?: number;
   name: string;
   routes?: Route[];
+  gtfsId: string;
```
binh-dam-ibigroup, line 493: Sort props

binh-dam-ibigroup (approved):

packages/types/src/index.ts
```diff
@@ -473,10 +486,11 @@ export type Stop = {
   dist?: number;
   geometries?: { geoJson?: GeoJSON.Polygon };
   id: string;
```
binh-dam-ibigroup, line 488: Consider adding a type alias per https://github.com/opentripplanner/otp-ui/pull/655/files#r1387985351 if we are to avoid breaking compatibility in a whole bunch of packages in #655.

daniel-heppner-ibigroup: Very good suggestion, thank you

packages/types/src/index.ts
```diff
@@ -460,6 +473,8 @@ export type StopLayerStop = LayerEntity & {
   name: string;
 };
 
+export type StopEventHanlder = (stop: Stop | { stopId: string }) => void;
```
miles-grant-ibigroup, line 476: Typo Handler

packages/types/src/index.ts
```diff
@@ -472,9 +487,10 @@ export type Stop = {
   color?: string;
   dist?: number;
   geometries?: { geoJson?: GeoJSON.Polygon };
+  gtfsId: string;
   id: string;
-  lat: number;
-  lon: number;
+  lat?: number;
+  lon?: number;
```
miles-grant-ibigroup, line 493: Why are these marked optional?

miles-grant-ibigroup (approved):
Small fixes, nothing serious

packages/types/src/index.ts
daniel-heppner-ibigroup: I added them to make something work I think, but we should probably remove them and just expect these fields to always be there.

packages/types/src/index.ts
daniel-heppner-ibigroup: thnx

PR #682, opened 2023-11-09
Wheelchair accessible labeling

amy-corson-ibigroup:
Provides invisible labeling for wheelchair accessibility badges so that users with AT can identify whether a leg or itinerary is wheelchair accessible.

See discussion on #675, closed due to mishandled merge

### DIFF
--- a/packages/itinerary-body/src/ItineraryBody/accessibility-rating.tsx
+++ b/packages/itinerary-body/src/ItineraryBody/accessibility-rating.tsx
@@ -2,6 +2,8 @@ import { GradationMap } from "@opentripplanner/types";
 import React, { ReactElement } from "react";
 import styled from "styled-components";
 import { Wheelchair } from "@styled-icons/foundation/Wheelchair";
+import { useIntl } from "react-intl";
+import { InvisibleAdditionalDetails } from "../styled";
 
 interface WrapperProps {
   border: boolean;
@@ -39,6 +41,7 @@ const TextWrapper = styled.span`
 interface Props {
   gradationMap?: GradationMap;
   grayscale?: boolean;
+  isLeg?: boolean;
   large?: boolean;
   score: number;
 }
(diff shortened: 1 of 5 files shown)


packages/itinerary-body/src/ItineraryBody/accessibility-rating.tsx
```diff
@@ -75,15 +95,28 @@ const AccessibilityRating = ({
   // External configuration may report "0.0" as 0, so include fallback
   const mapped = mapping[mappedKey] || mapping[0.0];
 
+  const accessibilityPreface = intl.formatMessage({
```
binh-dam-ibigroup, line 98: This can be up for discussion later, but instinctively, adding this text makes the whole trip summary feel too long.

binh-dam-ibigroup (approved):

packages/itinerary-body/src/ItineraryBody/accessibility-rating.tsx
amy-corson-ibigroup: I'm open to alternatives! There has to be some way to indicate whether the accessibility summary is referring to the leg or the trip, I think this was the best idea we had without rewriting the gradation map or making a breaking change.

amy-corson-ibigroup: But I agree with you, it is a little awkwardly phrased.

AdrianaCeric: Hm I think a way to make it more clear is "Wheelchair accessibility of this trip from `____` to `____` is `inaccessible/unknown`".

AdrianaCeric (approved):
I think it looks good on screen reader!

packages/itinerary-body/src/ItineraryBody/accessibility-rating.tsx
amy-corson-ibigroup: I think the issue with that is that on any given list of trip results, they will all have the same origin and destination so it may feel a little redundant!

PR #691, opened 2023-12-12
Add Trolleybus Support

miles-grant-ibigroup:
We forgot to add trolleybus support. This PR adds trolleybus support across the entire project. Adds unit tests to make sure it worked.


AdrianaCeric (approved):
Trolleybuses are important

packages/core-utils/src/route.ts
```diff
@@ -184,6 +184,7 @@ function getSortValues(
 const modeComparatorValue = {
   SUBWAY: 1,
   TRAM: 2,
+  TROLLEYBUS: 9,
```
amy-corson-ibigroup, line 187: Wait shouldn't this be 11? In the GTFS spec trolleybus is 11, doesn't OTP follow the spec

packages/core-utils/src/route.ts
```diff
@@ -207,7 +208,8 @@ const routeTypeComparatorValue = {
   // TODO: 11 and 12 are not a part of OTP as of 2019-02-14, but for now just
   // associate them with bus/rail.
   11: modeComparatorValue.BUS, // - Trolleybus.
-  12: modeComparatorValue.RAIL // - Monorail.
+  12: modeComparatorValue.RAIL, // - Monorail.
+  13: modeComparatorValue.TROLLEYBUS
```
amy-corson-ibigroup, line 212: Lines 210 and 212 both seem to be accounting for Trolleybuses here?

amy-corson-ibigroup (commented):
Quick question!!

packages/core-utils/src/route.ts
miles-grant-ibigroup: I don't think this has anything to do with OTP mode numbers. This has to do with determining route sort order

packages/core-utils/src/route.ts
miles-grant-ibigroup: The comment is wrong, but only 212 accounts for trolleybus

amy-corson-ibigroup: Could we remove the wrong comment

amy-corson-ibigroup (approved):

packages/core-utils/src/route.ts
miles-grant-ibigroup: I'm worried about what information we might lose... These comments are very old!

PR #692, opened 2023-12-15
Fix Crash on Missing Route Data (itin-body)

miles-grant-ibigroup:
we treated a value that is optional as non optional. this is now corrected


amy-corson-ibigroup (approved):
Beautiful

daniel-heppner-ibigroup (approved):
symptom of bad types

PR #693, opened 2023-12-15
Support Null Route Object (Printable Itinerary)

miles-grant-ibigroup:
Temporary patch to support null route objects until we figure out how we want to handle it inside of core utils

### DIFF
--- a/packages/printable-itinerary/src/transit-leg.tsx
+++ b/packages/printable-itinerary/src/transit-leg.tsx
@@ -27,6 +27,12 @@ export default function TransitLeg({
   const stopIdFrom = getDisplayedStopId(leg.from);
   const stopIdTo = getDisplayedStopId(leg.to);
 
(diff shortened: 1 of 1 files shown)


amy-corson-ibigroup (approved):

daniel-heppner-ibigroup (approved):
Is there a bug report or something where we can track that work needs to be done elsewhere? I don't want the tech debt to pile up

miles-grant-ibigroup:
I don't think there's any work to be done elsewhere @daniel-heppner-ibigroup

daniel-heppner-ibigroup:
```
  // TODO: core-utils needs some larger-scale type fixes
  // null is an object, so we need to redefine it as undefined to prevent
  // it from being read as an object. The long term solution is to avoid
  // using type checks for this purpose
```

this @miles-grant-ibigroup

PR #695, opened 2023-12-18
Add tabIndex to Viewer Link

amy-corson-ibigroup:
Changing the stop and trip viewer from buttons to links caused a slight regression where the links were not keyboard-focusable. This adds a `tabIndex` to those links so that they appear back in the focus order. It also renames some of the involved styled components to reflect the fact that they are links and not buttons.


daniel-heppner-ibigroup (approved):
LGTM, thanks

packages/itinerary-body/src/styled.tsx
```diff
@@ -85,7 +85,7 @@ export const AnchorButton = styled.a`
   }
 `;
 
-export const LinkButton = styled.a`
+export const StyledLink = styled.a`
```
binh-dam-ibigroup, line 88: 🚨🚨 Breaking changes...

binh-dam-ibigroup: I see what is going on. `onClick` on an anchor `<a>` element is incorrect. To use `onClick`, you have to use a button. The correct use with anchors is `href`.

binh-dam-ibigroup: This is what I would do as a `fix` to avoid a breaking change:
- Revert the name changes
- Change to `export const LinkButton = styled(TransparentButton)`

binh-dam-ibigroup (changes requested):
The correct a11y fix IMO is to change the "links" to be buttons as indicated. Then, the artificial `tabIndex` is not needed.

amy-corson-ibigroup:
We changed these to links earlier this year due to some accessibility feedback. Because they behave as links (take a user to a different URL) it causes confusion to screenreaders when they're marked up as buttons. 

What qualifies this as a breaking change?

If the lack of href is causing an issue, maybe we can do something [like this](https://stackoverflow.com/questions/1070760/javascript-href-vs-onclick-for-callback-function-on-hyperlink)? `<S.ViewerLink href="#" onClick={this.onClick} tabIndex={0}>`

binh-dam-ibigroup:
> What qualifies this as a breaking change?

You renamed an exported symbol.

binh-dam-ibigroup:
How does `<S.ViewerLink href="#" ... >` behave?

amy-corson-ibigroup:
@binh-dam-ibigroup Gotcha, I didn't realize those were getting exported in the `index.ts` file, thanks for catching that! Removed the name changes to avoid the breaking change.

I think the `href="#"` should only reset to the top of the page, but that gets overriden with the `onClick`. Not sure it actually does anything, but if we're running into issues with a missing `href`, that should help!

packages/itinerary-body/src/TransitLegBody/view-trip-button.tsx
```diff
@@ -28,7 +28,7 @@ class ViewTripButton extends Component<Props> {
 
   render(): ReactElement {
     return (
-      <S.ViewerButton onClick={this.onClick} type="button">
+      <S.ViewerButton href="#" onClick={this.onClick} tabIndex={0}>
```
binh-dam-ibigroup, line 31: `href="#"` does not work, it will clear out the query params in the url bar and that will mess up OTP-RR state, so i guess if we must use `<a>`, we are stuck with not setting `href` and keep `tabIndex={0}`.

binh-dam-ibigroup (changes requested):
Remove `href="#"`, otherwise it will mess up OTP-RR state.

binh-dam-ibigroup:
@amy-corson-ibigroup In the same vein as our team discussion using anchors the other day, one way to make the code correct (via a breaking change) is to make the ItineraryBody component accept some kind of URL template or URL getter function, so that the the stop viewer and trip viewer links actually work as links with `href` and with the links including the search params.

packages/itinerary-body/src/TransitLegBody/view-trip-button.tsx
binh-dam-ibigroup: @amy-corson-ibigroup Do this ^^

binh-dam-ibigroup (changes requested):
Sorry, I just noticed this playing with the keyboard. You need to implement an event handler for `onKeyDown` and respond to the `ENTER` key. Otherwise, the stop viewer and trip viewer links are inoperable when using the keyboard (in Storybook, you will see that no action is triggered when pressing `ENTER` while the link is focused). This, or inform the client that buttons are a much better solution.

amy-corson-ibigroup:
Okay, I have reverted back to `button` and used the `role="link"` hopefully that should solve it!

PR #696, opened 2023-12-28
Support for routing based on mobility profile

binh-dam-ibigroup:
This PR adds support for a new, unofficial OTP query parameter `mobilityProfile`.


binh-dam-ibigroup:
Superseded by #781.

PR #697, opened 2023-12-28
trip-details: improve handing of undefined fares

daniel-heppner-ibigroup:
This PR improves handling of undefined fares in the fare table. Sometimes some legs of a journey don't have a fare for every fare type. For example, a trip planner for the Puget Sound might have a response that includes some legs without ORCA fares for the rural agencies that don't accept it. This used to default to $0, but now it will show a "-" indicating no value. Also, the total headers will show - if any legs in the column don't have data.

### DIFF
--- a/packages/trip-details/src/fare-table.tsx
+++ b/packages/trip-details/src/fare-table.tsx
@@ -1,4 +1,4 @@
-import { Leg } from "@opentripplanner/types";
+import { FareProductSelector, Leg } from "@opentripplanner/types";
 import React from "react";
 import styled from "styled-components";
 import { Transfer } from "@styled-icons/boxicons-regular/Transfer";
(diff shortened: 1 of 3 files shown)


packages/trip-details/src/fare-table.tsx
```diff
@@ -152,16 +162,14 @@ const FareTypeTable = ({
...
-                    fare?.price?.currency?.code,
-                    fare?.price?.amount || 0
+                  {transferAmount?.amount > 0 && (
+                    <>
+                      <TransferIcon size={16} />{" "}
+                    </>
                   )}
+                  {price
+                    ? renderFare(price?.currency?.code, price?.amount)
+                    : "-"}
```
binh-dam-ibigroup, line 172: So "-" means that it is not possible for those riders to use the indicated payment methods? Should we announce it that way for screen readers?

binh-dam-ibigroup (approved):
LGTM with some non-blocking questions.

packages/trip-details/src/styled.ts
```diff
@@ -1,5 +1,14 @@
...
 
+export const InvisibleA11yLabel = styled.span`
+  display: inline-block;
+  clip: rect(0, 0, 0, 0);
+  height: 0;
+  overflow: hidden;
+  position: absolute;
+  width: 0;
+`;
+
```
amy-corson-ibigroup, line 11: We do already have this in itinerary body. Should there be one that gets exported from a more central location?

amy-corson-ibigroup (approved):
Looks good! Just quick q about the new `InvisibleAdditionalDetails` component

packages/trip-details/src/styled.ts
daniel-heppner-ibigroup: Yeah you're right. I just really didn't want to go through the pain of updating several packages.

daniel-heppner-ibigroup:
@binh-dam-ibigroup can you help with i18n?

packages/trip-details/src/fare-table.tsx
```diff
@@ -113,7 +119,16 @@ const FareTypeTable = ({
...
                 {boldText(useGetHeaderString(col.columnHeaderKey))}
                 <br />
-                {renderFare(fare?.currency?.code, fare?.amount || 0)}
+                {fare?.amount !== undefined ? (
+                  renderFare(fare?.currency?.code, fare?.amount)
+                ) : (
+                  <>
+                    -
+                    <InvisibleA11yLabel>
+                      <FormattedMessage id="otpUi.TripDetails.legMissingFareInfo" />
```
binh-dam-ibigroup, line 128: I think you swapped the leg and total fare by mistake...

binh-dam-ibigroup (changes requested):
I think you swapped the a11y rendering of unknown leg and total fare text.

packages/trip-details/src/fare-table.tsx
daniel-heppner-ibigroup: Oops... good catch

daniel-heppner-ibigroup:
Really good catch on that one. We probably wouldn't have noticed that for a long time. @binh-dam-ibigroup

binh-dam-ibigroup (approved):
Thanks for the changes!

PR #698, opened 2023-12-29
fix(core-utils): adds position information to stop query

daniel-heppner-ibigroup:
The from/to fields in the plan query needed lat lon information. 

Currently in use via alpha branch in nearby-view PR.


binh-dam-ibigroup (approved):

miles-grant-ibigroup (approved):
Wonderful

PR #700, opened 2024-01-10
fix(geocoder/otp): correctly accept `stop.code` as optional

miles-grant-ibigroup:
The OTP geocoder response was not correctly implemented. `stop.code` is optional. This means we were sometimes rendering `(undefined)`. This PR corrects this mistake.


daniel-heppner-ibigroup (approved):
Good catch, thanks for the fix

amy-corson-ibigroup (approved):

PR #702, opened 2024-01-15
feat(geocoder/photon): allow to use a custom Photon geocoding service

otrichet:
This change add the ability to use a custom Photon instance instead of the default one at https://photon.komoot.io/.

When the baseUrl is not provided in the configuration, the existing behavior of use the _komoot_ instance is kept.

### DIFF
--- a/packages/geocoder/src/geocoders/photon.ts
+++ b/packages/geocoder/src/geocoders/photon.ts
@@ -40,7 +40,7 @@ export default class PhotonGeocoder extends Geocoder {
       focusPoint,
       options,
       size,
-      url: baseUrl ? `${baseUrl}/autocomplete` : undefined,
+      url: baseUrl ? `${baseUrl}/api` : undefined,
       ...query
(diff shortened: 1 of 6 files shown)


miles-grant-ibigroup (approved):
Sorry for the delay on reviewing this. Thanks so much for the changes!

packages/geocoder/src/geocoders/photon.ts
```diff
@@ -40,7 +40,7 @@ export default class PhotonGeocoder extends Geocoder {
       focusPoint,
       options,
       size,
-      url: baseUrl ? `${baseUrl}/autocomplete` : undefined,
+      url: baseUrl ? `${baseUrl}/api` : undefined,
```
binh-dam-ibigroup, line 43: Did these paths become incorrect at one point?

binh-dam-ibigroup (approved):
Changes look fine. Merci !

PR #703, opened 2024-01-17
map-popup: hide popup when clicking action button

miles-grant-ibigroup:
This was sort of annoying -- you'd click "view stop" but then the popup would just linger! This PR resolves this issue.

### DIFF
--- a/packages/otp2-tile-overlay/src/index.tsx
+++ b/packages/otp2-tile-overlay/src/index.tsx
@@ -150,7 +150,7 @@ const OTP2TileLayerWithPopup = ({
             configCompanies={configCompanies}
             entity={{ ...clickedEntity, id: clickedEntity?.id || clickedEntity?.gtfsId }}
             setLocation={setLocation ? (location) => { setClickedEntity(null); setLocation(location) } : null}
-            setViewedStop={setViewedStop}
+            setViewedStop={setViewedStop ? (stop) => { setClickedEntity(null);setViewedStop(stop) } : null}
           />
 
         </Popup>
--- a/packages/park-and-ride-overlay/src/index.tsx
+++ b/packages/park-and-ride-overlay/src/index.tsx
@@ -48,7 +48,14 @@ const ParkAndRideOverlay = (props: Props): JSX.Element => {
                       lon: location.x,
                       name
                     }}
-                    setLocation={setLocation}
+                    setLocation={
+                      setLocation
+                        ? l => {
+                            setLocation(null);
+                            setLocation(l);
+                          }
+                        : null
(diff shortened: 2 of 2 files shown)


packages/otp2-tile-overlay/src/index.tsx
```diff
@@ -150,7 +150,7 @@ const OTP2TileLayerWithPopup = ({
             configCompanies={configCompanies}
             entity={{ ...clickedEntity, id: clickedEntity?.id || clickedEntity?.gtfsId }}
             setLocation={setLocation ? (location) => { setClickedEntity(null); setLocation(location) } : null}
-            setViewedStop={setViewedStop}
+            setViewedStop={setViewedStop ? (stop) => { setClickedEntity(null);setViewedStop(stop) } : null}
```
daniel-heppner-ibigroup, line 153: hate to do this to ya, but can you put this in its own callback function with a `useCallback`?

packages/park-and-ride-overlay/src/index.tsx
```diff
@@ -48,7 +48,14 @@ const ParkAndRideOverlay = (props: Props): JSX.Element => {
...
                     }}
-                    setLocation={setLocation}
+                    setLocation={
+                      setLocation
+                        ? l => {
+                            setLocation(null);
+                            setLocation(l);
+                          }
+                        : null
+                    }
```
daniel-heppner-ibigroup, line 58: same as above, move this into a callback please.

daniel-heppner-ibigroup (changes requested):
I left a few comments on the callbacks, but my request for changes is because I was thinking about this and I think we could make this reactive rather and eliminate the `useEffect`. 

I'm thinking that we could have a computed value for hidden layers that is based on the state + a prop called `computedHiddenLayers`. If `showEverything` is true, then `computedHiddenLayers` is just an empty array, otherwise it's set equal to `hiddenLayers` from the state. This would allow the layer state to revert back when you leave nearby view.

packages/otp2-tile-overlay/src/index.tsx
miles-grant-ibigroup: When is this reused though? The stop is always changing, so the function would constantly be getting regenerated anyway!

daniel-heppner-ibigroup: You're right, good point

daniel-heppner-ibigroup (approved):
Looks great, thank you for the change. Excited to try out showEverything!

amy-corson-ibigroup (approved):
Looks good! Thank you!

PR #704, opened 2024-01-17
itinerary body: never display `“default vehicle type” `

miles-grant-ibigroup:
We had some override techniques in place, but nothing will be as reliable as this for removing that silly string!

### DIFF
--- a/packages/core-utils/src/itinerary.ts
+++ b/packages/core-utils/src/itinerary.ts
@@ -410,10 +410,10 @@ export function getCompanyForNetwork(
  * @return {string}  A label for use in presentation on a website.
  */
 export function getCompaniesLabelFromNetworks(
-  networks: string[],
+  networks: string[] | string,
   companies: Company[] = []
 ): string {
-  return networks
+  return (Array.isArray(networks) ? networks : [networks])
     .map(network => getCompanyForNetwork(network, companies))
     .filter(co => !!co)
     .map(co => co.label)
--- a/packages/itinerary-body/src/util.ts
+++ b/packages/itinerary-body/src/util.ts
@@ -157,10 +157,11 @@ export function getPlaceName(
     companies
   );
   if (
+    intl &&
     // Don't ever show this useless OTP default string
-    place.name === "Default vehicle type" ||
-    (place.name.match(/-/g) || []).length > 3 ||
-    company?.overridePlaceNames
+    (place.name === "Default vehicle type" ||
+      (place.name.match(/-/g) || []).length > 3 ||
+      company?.overridePlaceNames)
   ) {
     return intl.formatMessage(
       {
--- a/packages/itinerary-body/src/util.ts
+++ b/packages/itinerary-body/src/util.ts
@@ -162,19 +162,17 @@ export function getPlaceName(
     (place.name.match(/-/g) || []).length > 3 ||
     company?.overridePlaceNames
   ) {
-    if (intl) {
-      return intl.formatMessage(
-        {
-          defaultMessage: defaultMessages["otpUi.AccessLegBody.vehicleTitle"],
-          description: "Formats rental vehicle company and type",
-          id: "otpUi.AccessLegBody.vehicleTitle"
-        },
-        {
-          company: company?.label,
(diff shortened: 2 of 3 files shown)


packages/core-utils/src/itinerary.ts
```diff
@@ -410,10 +410,10 @@ export function getCompanyForNetwork(
  * @return {string}  A label for use in presentation on a website.
  */
 export function getCompaniesLabelFromNetworks(
-  networks: string[],
+  networks: string[] | string,
```
daniel-heppner-ibigroup, line 413: Is there a reason we can't make a single item array where this method is called rather than polluting the argument types?

packages/map-popup/src/util.ts
```diff
@@ -21,7 +21,11 @@ export function makeDefaultGetEntityName(
     let stationName: string | null = entity.name || entity.id;
     // If the station name or id is a giant UUID (with more than 3 "-" characters)
     // best not to show that at all. The company name will still be shown.
-    if ((stationName.match(/-/g) || []).length > 3) {
+    // Also ignore "Default Vehicle Type"
+    if (
+      (stationName.match(/-/g) || []).length > 3 ||
+      stationName === "Default vehicle type"
```
daniel-heppner-ibigroup, line 27: I remember when this was written and thought it was a terrible hack, so it seems like a fitting spot to add a hardcoded string.

packages/itinerary-body/src/util.ts
```diff
@@ -157,18 +157,20 @@ export function getPlaceName(
...
+    if (intl) {
       return intl.formatMessage(
         {
           defaultMessage: defaultMessages["otpUi.AccessLegBody.vehicleTitle"],
           description: "Formats rental vehicle company and type",
           id: "otpUi.AccessLegBody.vehicleTitle"
         },
         {
-          company: company.label,
+          company: company?.label,
```
daniel-heppner-ibigroup, line 173: What happens if company is undefined? Do we still want to display this label?

packages/itinerary-body/src/util.ts
```diff
@@ -157,18 +157,20 @@ export function getPlaceName(
     companies
   );
   if (
+    // Don't ever show this useless OTP default string
+    place.name === "Default vehicle type" ||
     (place.name.match(/-/g) || []).length > 3 ||
     company?.overridePlaceNames
   ) {
-    if (company && intl) {
+    if (intl) {
```
daniel-heppner-ibigroup, line 165: Can this if block be merged with the above one?

daniel-heppner-ibigroup (approved):
A few little comments but looks good overall

packages/core-utils/src/itinerary.ts
miles-grant-ibigroup: Yes it's because I don't understand why or when we get a string instead of an array

packages/map-popup/src/util.ts
miles-grant-ibigroup: The hacks feel safer around each other

packages/itinerary-body/src/util.ts
miles-grant-ibigroup: I think it falls back gracefully

packages/itinerary-body/src/util.ts
miles-grant-ibigroup: Sure can! Good catch thanks

packages/itinerary-body/src/util.ts
```diff
@@ -157,22 +157,22 @@ export function getPlaceName(
...
-          description: "Formats rental vehicle company and type",
-          id: "otpUi.AccessLegBody.vehicleTitle"
-        },
-        {
-          company: company.label,
-          vehicleType: getVehicleType(place.vertexType, intl)
-        }
-      );
-    }
+    return intl.formatMessage(
```
binh-dam-ibigroup, line 165: Is `intl` guaranteed to be defined?

miles-grant-ibigroup: Good catch thanks!

binh-dam-ibigroup (approved):

PR #707, opened 2024-01-25
fix(icons): Pass leg prop from classicicon to modeicon

amy-corson-ibigroup:
In some cases, `ModeIcon` requires the `leg` prop, but `ClassicIcon` does not pass it. This fixes that!

### DIFF
--- a/packages/icons/src/leg-icon.js
+++ b/packages/icons/src/leg-icon.js
@@ -20,7 +20,13 @@ const LegIcon = ({ getCompanyIcon, leg, ModeIcon, ...props }) => {
   // Do this for P&R, K&R and TNC trips without company icon
   if (iconStr && iconStr.startsWith("CAR")) iconStr = "CAR";
 
-  return <ModeIcon mode={iconStr} leg={leg} {...props} />;
+  const shorterLeg = {
+    longName: leg.longName,
+    shortName: leg.shortName,
+    routeId: leg.routeId
+  };
+
+  return <ModeIcon mode={iconStr} leg={shorterLeg} {...props} />;
 };
 
 LegIcon.propTypes = {
--- a/packages/icons/src/leg-icon.js
+++ b/packages/icons/src/leg-icon.js
@@ -22,11 +22,11 @@ const LegIcon = ({ getCompanyIcon, leg, ModeIcon, ...props }) => {
 
   const shorterLeg = {
     longName: leg.longName,
-    shortName: leg.shortName,
-    routeId: leg.routeId
+    routeId: leg.routeId,
+    shortName: leg.shortName
   };
 
-  return <ModeIcon mode={iconStr} leg={shorterLeg} {...props} />;
+  return <ModeIcon leg={shorterLeg} mode={iconStr} {...props} />;
 };
 
 LegIcon.propTypes = {
--- a/packages/icons/src/leg-icon.js
+++ b/packages/icons/src/leg-icon.js
@@ -20,7 +20,7 @@ const LegIcon = ({ getCompanyIcon, leg, ModeIcon, ...props }) => {
   // Do this for P&R, K&R and TNC trips without company icon
   if (iconStr && iconStr.startsWith("CAR")) iconStr = "CAR";
 
-  return <ModeIcon mode={iconStr} {...props} />;
+  return <ModeIcon mode={iconStr} leg={leg} {...props} />;
 };
 
 LegIcon.propTypes = {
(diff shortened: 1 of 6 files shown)


packages/icons/src/leg-icon.js
```diff
@@ -20,7 +20,7 @@ const LegIcon = ({ getCompanyIcon, leg, ModeIcon, ...props }) => {
   // Do this for P&R, K&R and TNC trips without company icon
   if (iconStr && iconStr.startsWith("CAR")) iconStr = "CAR";
 
-  return <ModeIcon mode={iconStr} {...props} />;
+  return <ModeIcon mode={iconStr} leg={leg} {...props} />;
```
miles-grant-ibigroup, line 23: `leg` is on props so doesn't this get passed anyway?

miles-grant-ibigroup (changes requested):
I'm a bit confused

packages/icons/src/leg-icon.js
amy-corson-ibigroup: I would have thought so too but it's not getting passed. I think because `getCompanyIcon`, `leg`, and `ModeIcon` all get destructured and so `props` winds up being what's left over? I could be wrong about why

miles-grant-ibigroup: Oooh that's a good point. Honestly this is probably the right way to do it anyway...

miles-grant-ibigroup (approved):
Thanks for the explanation!

miles-grant-ibigroup:
Actually wait. Looking at the snapshots this looks like it's a LOT of pollution. Can we try to only pick out the parts of leg we actually want?

miles-grant-ibigroup (approved):
Terrific!

binh-dam-ibigroup (commented):
Could you write a story to illustrate the change?

packages/icons/src/stories/custom-mode-icon-renderer.js
```diff
@@ -0,0 +1,41 @@
...
+import { StyledTable } from "../icon-renderer";
+import ClassicModeIcon from "../classic-mode-icon";
+import CustomModeIcon from "./custom-mode-icon";
+
+const walkTransitWalkItinerary = require("@opentripplanner/itinerary-body/src/__mocks__/itineraries/walk-transit-walk.json");
+
+const exampleLeg = walkTransitWalkItinerary.legs[1];
+
+const CustomModeIconRenderer = () => {
+  return (
```
binh-dam-ibigroup, line 11: You can replace `{ return (` with just `(`.

packages/icons/src/stories/custom-mode-icon-renderer.js
```diff
@@ -0,0 +1,41 @@
...
+        <tr>
+          <th>{exampleLeg.mode}</th>
+          <th>Result</th>
+        </tr>
+      </thead>
+      <tbody>
+        <tr>
+          <td>Classic Mode</td>
+          <td>
+            <div>
```
binh-dam-ibigroup, line 23: Do we need those divs?

binh-dam-ibigroup (approved):
Yep, thanks for the changes. Cleanup a few things before merging.

packages/icons/src/stories/custom-mode-icon.js
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+
+import { ClassicGondola } from "../classic";
+import ClassicModeIcon from "../classic-mode-icon";
+
+function CustomModeIcon({ mode, leg, ...props }) {
+  if (!mode) return null;
+  if (leg?.routeId === "TriMet:100") return <ClassicGondola {...props} />;
+  return <ClassicModeIcon mode={mode} leg={leg} props={props} />;
```
binh-dam-ibigroup, line 9: Sort props.

packages/icons/src/stories/custom-mode-icon.js
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+
+import { ClassicGondola } from "../classic";
+import ClassicModeIcon from "../classic-mode-icon";
+
+function CustomModeIcon({ mode, leg, ...props }) {
```
binh-dam-ibigroup, line 6: Sort props.

packages/icons/src/stories/custom-mode-icon-renderer.js
```diff
@@ -0,0 +1,41 @@
...
+              <ClassicModeIcon mode={exampleLeg.mode} leg={exampleLeg} />
```
binh-dam-ibigroup, line 24: Sort props

packages/icons/src/stories/custom-mode-icon-renderer.js
amy-corson-ibigroup: There's `div` styling in the `StyledTable` that keeps the icons the right size!

PR #708, opened 2024-01-29
Transit vehicle overlay show longname

daniel-heppner-ibigroup:
Makes the transit vehicle overlay use the long name if the route short name isn't available. This saves us from using a hack in OTP-RR to make the tooltip show up for agencies that don't publish a short name.


miles-grant-ibigroup (approved):
Oy vey...

PR #710, opened 2024-02-08
fix: support new OTP geocoder `agencies` field

miles-grant-ibigroup:
Support for https://github.com/opentripplanner/OpenTripPlanner/pull/5634

No localization, but I think this is fine since there are no English words and all the GTFS will be non-localized anyway

### DIFF
--- a/packages/geocoder/src/geocoders/otp.ts
+++ b/packages/geocoder/src/geocoders/otp.ts
@@ -5,6 +5,24 @@ import type { AutocompleteQuery, MultiGeocoderResponse } from "./types";
 import Geocoder from "./abstract-geocoder";
 import { OTPGeocoderResponse } from "../apis/otp";
 
+
+const generateLabel = stop => {
+  let brackets = ""
+  if (stop?.agencies?.[0]?.name) {
+    brackets += stop.agencies[0].name
+  } else if (stop?.feedPublisher?.name) {
+    brackets += stop.feedPublisher.name
+  }
+  if (stop?.code) {
+    if (brackets !== "") brackets += " "
+    brackets += stop.code
+  }
+
+  if (brackets === "") return stop.name
+
+  return `${stop.name} (${brackets})`
+}
+
 /**
  * Allows fetching results from OTP instance with the geocoder endpoint enabled
  */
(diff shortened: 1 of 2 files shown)


amy-corson-ibigroup (approved):
Makes sense to me!!

miles-grant-ibigroup:
Wait assigning this back to me. I need to add a check that only adds a space to the string if there is nothing present already

packages/geocoder/src/geocoders/otp.ts
```diff
@@ -21,7 +21,23 @@ export default class OTPGeocoder extends Geocoder {
 
 
   rewriteAutocompleteResponse(response: OTPGeocoderResponse): MultiGeocoderResponse {
-    const generateLabel = stop => stop.code ? `${stop.name} (${stop.code})` : stop.name
+
+    const generateLabel = stop => {
```
daniel-heppner-ibigroup, line 25: Does this function need to be inside `rewriteAutocompleteResponse`? It doesn't seem like you're using the closure, so it could probably be outside. Maybe just put it at the top of the file?

packages/geocoder/src/geocoders/otp.ts
```diff
@@ -21,7 +21,23 @@ export default class OTPGeocoder extends Geocoder {
...
+      let brackets = ""
+      if (stop?.agencies?.[0]?.name) {
+        brackets += stop.agencies[0].name
+      } else if (stop?.feedPublisher?.name) {
+        brackets += stop.feedPublisher.name
+      } 
+      if (stop?.code) {
+        if (brackets !== "") brackets += " "
+        brackets += stop.code
+      }
```
daniel-heppner-ibigroup, line 35: This is probably fine, but it might be a bit cleaner depending on your opinion to make brackets an array, remove the logic around putting a space into it, and using array.join(" ") to convert it to a string.

daniel-heppner-ibigroup (approved):
A few thoughts for cleanup but nothing blocking.

packages/geocoder/src/geocoders/otp.ts
miles-grant-ibigroup: If we had more then 2 things in the brackets this would be good

PR #711, opened 2024-02-08
MAP: Ignore Stops with No Routes

miles-grant-ibigroup:
Finally wrangled the filters enough to get this working! What nonsense


daniel-heppner-ibigroup (approved):
What a crazy syntax. There's a perfectly good programming language right there! Where are some examples of this rule getting applied?

miles-grant-ibigroup:
> What a crazy syntax. There's a perfectly good programming language right there! Where are some examples of this rule getting applied?

In the OTP2 tile layer, any stops layer will now hide any stop that has no departures at it

PR #712, opened 2024-02-08
feat(core-utils): remove old query params handled by new mode selector

daniel-heppner-ibigroup:
This PR removes a bunch of old query parameters that are no longer needed here, either because they're now directly managed by the mode selector (mode settings) or because they're no longer supported by OTP (like max walk distance).


binh-dam-ibigroup (changes requested):
I think you should add a breaking change commit (or redo the PR with that commit being breaking change).

daniel-heppner-ibigroup:
Lost track of this PR for a bit, but I force pushed the breaking change commit. If you had this branch cloned, you'll need to delete it and pull it again. Or maybe force pull.

packages/core-utils/src/query-params.js
binh-dam-ibigroup, line 1: The comment on lines 1-2 doesn't apply anymore and can be removed.

binh-dam-ibigroup, line 1: We can probably remove the profile-only params that have `routingTypes: ["PROFILE"]` (these are holdouts from before my time) such as `startTime`, `endTime`.

binh-dam-ibigroup: In the same vein, I think we can get rid of `routingTypes`, `profileRewrite` alltogether. Thoughts?

binh-dam-ibigroup, line 1: I have never seen `showIntermediateStops`, `otherThanPreferredRoutesPenalty` in action, and have not seen `ignoreRealtimeUpdates`, `companies`,  in action in years, so that one can be deleted too.

packages/core-utils/src/query-params.js
```diff
@@ -162,119 +153,6 @@ const queryParams = [
...
-    ],
-    itineraryRewrite: value => ({
-      maxWalkDistance: value,
-      // ensures that the value is repopulated when loaded from URL params
-      maxBikeDistance: value
-    })
-  },
-
   {
     /* optimize -- how to optimize a trip (non-bike, non-micromobility trips) */
```
binh-dam-ibigroup, line 157: Are `optimize` and `optimizeBike` still in use?

packages/core-utils/src/query-params.js
```diff
@@ -324,198 +202,6 @@ const queryParams = [
...
-    itineraryRewrite: value => ({
-      maxWalkDistance: value,
-      // ensures that the value is repopulated when loaded from URL params
-      maxEScooterDistance: value
-    })
-  },
-
   {
     /* bikeSpeed -- the user's bikeSpeed speed in m/s */
     name: "watts",
```
binh-dam-ibigroup, line 207: Are the device power settings (`watts`) still in use?

packages/core-utils/src/query-params.js
binh-dam-ibigroup: Is `numItineraries` supposed to come from config for each mode combination later on? If so, it can be removed from here.

binh-dam-ibigroup: How about all the params below `otherThanPreferredRoutesPenalty`?

binh-dam-ibigroup (changes requested):
Can we delete more old params and get rid of the old routing types?

packages/core-utils/src/query-params.js
daniel-heppner-ibigroup: I think the file should still be removed entirely, but probably not because of settingsselector anymore.

daniel-heppner-ibigroup: Yes! I was worried someone was using those, like TriMet, but I think you're right that we can remove them.

daniel-heppner-ibigroup:
I deleted a lot more of the query parameters, but that caused some problems in the SettingsSelectorPanel, which is still being used in some instances and we aren't ready to fully deprecate, even with a breaking change. So I had to revert those changes, I think at this point we can merge this to fix the original bug.

binh-dam-ibigroup (changes requested):
Main ask is to remove the deleted query params from `defaultParams`. But should we also remove query parameters not supported by the mode selector?

daniel-heppner-ibigroup:
I'm closing the PR, let's try again without all the jank from the revert commits.

PR #713, opened 2024-02-09
Vehicle Overlay: Fix zoom bug fix

miles-grant-ibigroup:
Previously the zoom was only being checked on render. This means if you didn't click anything the layer wouldn't update and the cool ferris wheel behavior functioned very strangely. By moving zoom state into react state we do things more sensibly. Most importantly, the zoom behavior is fixed!

### DIFF
--- a/packages/vehicle-rental-overlay/src/index.tsx
+++ b/packages/vehicle-rental-overlay/src/index.tsx
@@ -95,7 +95,7 @@ const VehicleRentalOverlay = ({
   visible
 }: Props): JSX.Element => {
   const { current: map } = useMap();
-  const [zoom, setZoom] = useState(0);
+  const [zoom, setZoom] = useState(map?.getZoom());
 
   const layerId = `rental-vehicles-${id}`;
   const [clickedVehicle, setClickedVehicle] = useState(null);
--- a/packages/vehicle-rental-overlay/src/index.tsx
+++ b/packages/vehicle-rental-overlay/src/index.tsx
@@ -23,25 +23,6 @@ const getColorForStation = (v: Station) => {
   return "gray";
 };
 
-const checkIfPositionInViewport = (
-  bounds: mapboxgl.LngLatBounds,
-  lat: number,
-  lng: number
-): boolean => {
-  const PADDING = 0.001;
-  // @ts-expect-error types appear to be wrong? version issue?
-  // eslint-disable-next-line no-underscore-dangle
-  const [sw, ne] = [bounds._sw, bounds._ne];
-  if (!sw || !ne) return false;
-
-  return (
-    lat >= sw.lat - PADDING &&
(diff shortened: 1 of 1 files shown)


amy-corson-ibigroup (approved):
Makes sense! Thank you!

packages/vehicle-rental-overlay/src/index.tsx
```diff
@@ -114,8 +95,7 @@ const VehicleRentalOverlay = ({
   visible
 }: Props): JSX.Element => {
   const { current: map } = useMap();
-  const zoom = map?.getZoom();
-  const bounds = map?.getBounds();
+  const [zoom, setZoom] = useState(0);
```
binh-dam-ibigroup, line 98: Can you initialize the zoom state based on the map's zoom?

packages/vehicle-rental-overlay/src/index.tsx
```diff
@@ -192,44 +179,40 @@ const VehicleRentalOverlay = ({
         </Source>
       )}
       {zoom >= DETAILED_MARKER_CUTOFF &&
-        stations
-          .filter(station =>
-            checkIfPositionInViewport(bounds, station.y, station.x)
```
binh-dam-ibigroup, line 197: Was this filter causing rendering issues?

binh-dam-ibigroup (commented):
See comments

packages/vehicle-rental-overlay/src/index.tsx
miles-grant-ibigroup: It was. But it's not needed anymore. Vector tiles are much more performant than the raster ones

binh-dam-ibigroup: How about rendering extra HTML that won't get shown except for the small portion of bike stations within the viewport?

miles-grant-ibigroup: Why would we do this?

binh-dam-ibigroup: Removing that filter results in more markup sent to the browser that is not directly visible (actually more markup sent to maplibre for rendering). Was that the intent? Perhaps one case for removing the filter is to prevent re-rendering when slightly panning the map.

miles-grant-ibigroup: If we render/unrender markup using react we are now relying on react to do a map refresh instead of letting maplibre handle it. Maplibre does a good job at deciding whether to render something or not.

binh-dam-ibigroup: Good deal.

binh-dam-ibigroup (approved):
Seems to work, and thanks for the clarifications!

PR #727, opened 2024-03-19
fix(core-utils): remove query params conflicting with new mode selector

daniel-heppner-ibigroup:
BREAKING CHANGE: removes wheelchair and walkReluctance query parameters

Really keeping it simple with this one... Only removing the two parameters that are handled by the new mode selector settings dropdowns. 

Please check the general settings panel story. Does this matter? The settings selector panel and trip options stuff is unchanged.


binh-dam-ibigroup:
@daniel-heppner-ibigroup What was the reason for not removing the max walk/bike distance and some of the other no-longer supported items from #712?

PR #728, opened 2024-04-02
fix: support `maxVehicleAge` realtime vehicle filtering based on last update

miles-grant-ibigroup:
RT Vehicle updates have a lastUpdated field that we previously always ignored. This sometimes resulted in a bunch of busses piling up in the yard because they hadn't sent an update in forever. This PR introduces a new prop that can be used to hide vehicles that haven't updated their location in a certain amount of time

cc @fpurcell

### DIFF
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -104,6 +104,7 @@ export type TransitVehicle = {
 
   status?: string;
   reportDate?: string;
+  lastUpdated?: number;
   seconds?: number;
 
   stopSequence?: number;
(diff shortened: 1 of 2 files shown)


amy-corson-ibigroup (approved):
I for one will miss knowing the buses are safe in their homes!

packages/types/src/index.ts
```diff
@@ -104,6 +104,7 @@ export type TransitVehicle = {
 
   status?: string;
   reportDate?: string;
+  lastUpdated?: number;
```
binh-dam-ibigroup, line 107: Sort props.

binh-dam-ibigroup (approved):

PR #724, opened 2024-03-13
OTP2 Elevation Data Compatibility 

miles-grant-ibigroup:
We did not support elevation data from OTP2. This PR changes our query to work with our OTP1 compatible elevation UI

PR #729, opened 2024-04-03
Update elevation profile type to new GraphQL shape

daniel-heppner-ibigroup:
This is building a better fix for what https://github.com/opentripplanner/otp-ui/pull/724 is solving.


packages/types/src/index.ts
```diff
@@ -217,7 +217,7 @@ export type Step = {
   area: boolean;
   bogusName: boolean;
   distance: number;
-  elevation: ElevationData;
+  elevationProfile: ElevationData[];
```
binh-dam-ibigroup, line 220: To avoid breaking change and keep CI happy, let's keep the old field along with the new one.

packages/types/src/index.ts
```diff
@@ -193,9 +193,9 @@ export type EncodedPolyline = {
 };
 
 export type ElevationData = {
```
binh-dam-ibigroup, line 195: Keep the old `ElevationData` type, and create a new `ElevationProfile` type with the distance and elevation fields.

binh-dam-ibigroup (changes requested):
Please revert the breaking change and create a `feat` commit with the new elevation types and fields along with the existing ones.

packages/types/src/index.ts
daniel-heppner-ibigroup: @binh-dam-ibigroup Hmm... I liked this but we already have an `ElevationProfile` type. It could be `StepElevationProfile` or `ElevationProfileComponent` Since the elevation profile itself is an array of these, the second one might make sense? Although it could be confused with the existing ElevationProfile type which is something else. The first option is okay except that in the GraphQL schema there's no rigid connection between the `Step` type and the `ElevationProfile` type. Let me know what you think...

binh-dam-ibigroup: > I liked this but we already have an ElevationProfile type. It could be StepElevationProfile or ElevationProfileComponent [...]

Sure, pick a name that fits.

daniel-heppner-ibigroup:
I force pushed to remove the breaking change comment from the earlier commit. No more breaking change!

binh-dam-ibigroup (approved):

binh-dam-ibigroup:
@daniel-heppner-ibigroup Please remove the mention of a breaking change in the PR description.

miles-grant-ibigroup (approved):
Wonderful!

PR #730, opened 2024-04-03
Update elevation profile function for new elevation type

daniel-heppner-ibigroup:
This follows #729 to implement the new type.


daniel-heppner-ibigroup:
Types package needs to be updated before merging.

PR #732, opened 2024-04-08
feat(itinerary-body): Fix alerts a11y

amy-corson-ibigroup:
- Creates a `ul` and `li` for each transit alert
- Creates a clear alert hyperlink that opens in external window (if url is provided)
- Adds more invisible context for AT

### DIFF
--- a/packages/itinerary-body/i18n/en-US.yml
+++ b/packages/itinerary-body/i18n/en-US.yml
@@ -90,8 +90,10 @@ otpUi:
     viewOnMap: View on map
   TransitLegBody:
     AlertsBody:
+      alertLinkText: See alert on agency website
       effectiveDate: Effective as of {dateTime, date, long}
       effectiveTimeAndDate: Effective as of {dateTime, time, short}, {day}
+      externalLink: (External link)
       today: Today
       tomorrow: Tomorrow
       yesterday: Yesterday
--- a/packages/itinerary-body/i18n/en-US.yml
+++ b/packages/itinerary-body/i18n/en-US.yml
@@ -90,11 +90,12 @@ otpUi:
     viewOnMap: View on map
   TransitLegBody:
     AlertsBody:
-      agency: agency
       alertLinkText: See alert on {agency} website
       effectiveDate: Effective as of {dateTime, date, long}
       effectiveTimeAndDate: Effective as of {dateTime, time, short}, {day}
       externalLink: (External link)
+      linkOpensNewWindow: (Opens new window)
+      noAgencyAlertLinkText: See alert on agency website
       today: Today
       tomorrow: Tomorrow
       yesterday: Yesterday
--- a/packages/itinerary-body/i18n/en-US.yml
+++ b/packages/itinerary-body/i18n/en-US.yml
@@ -90,6 +90,7 @@ otpUi:
     viewOnMap: View on map
   TransitLegBody:
     AlertsBody:
+      agency: agency
       alertLinkText: See alert on {agency} website
       effectiveDate: Effective as of {dateTime, date, long}
       effectiveTimeAndDate: Effective as of {dateTime, time, short}, {day}
--- a/packages/itinerary-body/i18n/en-US.yml
+++ b/packages/itinerary-body/i18n/en-US.yml
@@ -90,7 +90,7 @@ otpUi:
     viewOnMap: View on map
   TransitLegBody:
     AlertsBody:
-      alertLinkText: See alert on agency website
+      alertLinkText: See alert on {agency} website
       effectiveDate: Effective as of {dateTime, date, long}
       effectiveTimeAndDate: Effective as of {dateTime, time, short}, {day}
       externalLink: (External link)
--- a/packages/itinerary-body/i18n/en-US.yml
+++ b/packages/itinerary-body/i18n/en-US.yml
@@ -93,7 +93,6 @@ otpUi:
       alertLinkText: See alert on {agency} website
       effectiveDate: Effective as of {dateTime, date, long}
       effectiveTimeAndDate: Effective as of {dateTime, time, short}, {day}
-      externalLink: (External link)
       linkOpensNewWindow: (Opens new window)
       noAgencyAlertLinkText: See alert on agency website
       today: Today
--- a/packages/itinerary-body/src/TransitLegBody/alerts-body.tsx
+++ b/packages/itinerary-body/src/TransitLegBody/alerts-body.tsx
@@ -3,7 +3,7 @@ import { toDate, utcToZonedTime } from "date-fns-tz";
 import coreUtils from "@opentripplanner/core-utils";
 import { Alert } from "@opentripplanner/types";
 import React, { FunctionComponent, ReactElement } from "react";
-import { FormattedMessage, useIntl } from "react-intl";
+import { FormattedMessage } from "react-intl";
 
 import { ExternalLinkAlt } from "@styled-icons/fa-solid";
 import * as S from "../styled";
@@ -70,10 +70,6 @@ export default function AlertsBody({
   timeZone = getUserTimezone()
 }: Props): ReactElement {
   if (typeof alerts !== "object") return null;
-  const intl = useIntl();
-  const defaultAgencyName = intl.formatMessage({
-    id: "otpUi.TransitLegBody.AlertsBody.agency"
-  });
   return (
     <S.TransitAlerts>
       {alerts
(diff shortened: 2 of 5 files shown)


packages/itinerary-body/src/styled.tsx
```diff
@@ -660,6 +660,13 @@ export const TransitAlert = styled.a`
   text-decoration: none;
 `;
 
+export const TransitAlertExternalLink = styled.a`
```
miles-grant-ibigroup, line 663: No underline on hover?

packages/itinerary-body/i18n/en-US.yml
```diff
@@ -90,8 +90,10 @@ otpUi:
     viewOnMap: View on map
   TransitLegBody:
     AlertsBody:
+      alertLinkText: See alert on agency website
```
miles-grant-ibigroup, line 93: Can we add a slot to insert operator name?

packages/itinerary-body/src/TransitLegBody/alerts-body.tsx
```diff
@@ -131,6 +132,24 @@ export default function AlertsBody({
                       }}
                     />
                   )}
+                  <br />
```
miles-grant-ibigroup, line 135: Can we add padding-top to S.TransitAlertExternalLink instead?

miles-grant-ibigroup (changes requested):
A few cleanup suggestions. This new behavior also NEEDS to be configurable. This is a huge behavior change that I think many may dislike

amy-corson-ibigroup:
Wait @miles-grant-ibigroup why would this need to be configurable? Is there something specific that you're anticipating is going to get pushback? A lot of these changes seem like pretty straightforward improvements to me.

miles-grant-ibigroup:
> Wait @miles-grant-ibigroup why would this need to be configurable? Is there something specific that you're anticipating is going to get pushback? A lot of these changes seem like pretty straightforward improvements to me.

Many OTP-UI implementers expect the entire alert to be clickable

packages/itinerary-body/src/TransitLegBody/alerts-body.tsx
```diff
@@ -5,12 +5,14 @@ import { Alert } from "@opentripplanner/types";
...
 import { FormattedMessage } from "react-intl";
 
+import { ExternalLinkAlt } from "@styled-icons/fa-solid";
 import * as S from "../styled";
 import { defaultMessages } from "../util";
 
 const { getUserTimezone, getCurrentDate } = coreUtils.time;
 
 interface Props {
+  agencyName: string;
```
miles-grant-ibigroup, line 15: `agencyName` is not guaranteed. We should make this optional and have a fallback ("agency")

packages/itinerary-body/src/TransitLegBody/alerts-body.tsx
```diff
@@ -131,6 +134,24 @@ export default function AlertsBody({
                       }}
                     />
                   )}
+                  {alertUrl.trim() && (
+                    <S.TransitAlertExternalLink href={alertUrl} target="_blank">
+                      <ExternalLinkAlt height={10} />
```
miles-grant-ibigroup, line 139: We shouldn't specify a height here? Or if we do we should flex align to baseline so that it is better aligned

packages/itinerary-body/src/TransitLegBody/alerts-body.tsx
```diff
@@ -131,6 +134,24 @@ export default function AlertsBody({
...
+                      <FormattedMessage
+                        defaultMessage={
+                          defaultMessages[
+                            "otpUi.TransitLegBody.AlertsBody.alertLinkText"
+                          ]
+                        }
+                        description="Text with the date an alert takes effect"
+                        id="otpUi.TransitLegBody.AlertsBody.alertLinkText"
+                        values={{ agency: agencyName }}
+                      />{" "}
```
miles-grant-ibigroup, line 149: Can we put the space inside the invisible details?

packages/itinerary-body/src/styled.tsx
```diff
@@ -660,6 +660,18 @@ export const TransitAlert = styled.a`
...
+  align-items: center;
```
miles-grant-ibigroup, line 664: This should be `baseline`

miles-grant-ibigroup (changes requested):
Everyone is excited about this! Just a few small tweaks and then I think we should finally be good to go

miles-grant-ibigroup:
If I were being really nitpicky I'd say let's also reduce the padding on the date so that the date and link are closer together... but I think that is not that importnat

packages/itinerary-body/src/TransitLegBody/alerts-body.tsx
```diff
@@ -131,6 +138,25 @@ export default function AlertsBody({
...
+                            "otpUi.TransitLegBody.AlertsBody.alertLinkText"
+                          ]
+                        }
+                        description="Text with the date an alert takes effect"
+                        id="otpUi.TransitLegBody.AlertsBody.alertLinkText"
+                        values={{ agency: agencyName || defaultAgencyName }}
+                      />
+                      <S.InvisibleAdditionalDetails>
+                        {" "}
+                        <FormattedMessage id="otpUi.TransitLegBody.AlertsBody.externalLink" />
```
binh-dam-ibigroup, line 156: In this context, I feel that "external link" is redundant with "See alert on agency's website", unless it means "opens a new tab/window".

packages/itinerary-body/src/TransitLegBody/index.tsx
```diff
@@ -303,12 +303,22 @@ class TransitLegBody extends Component<Props, State> {
...
                 )}
+                <S.InvisibleAdditionalDetails>
+                  <FormattedMessage
+                    defaultMessage={
+                      defaultMessages["otpUi.TransitLegBody.expandDetails"]
+                    }
+                    description="Screen reader text added to expand steps"
+                    id="otpUi.TransitLegBody.expandDetails"
+                  />
+                </S.InvisibleAdditionalDetails>
```
binh-dam-ibigroup, line 314: The "expand" hint should be inside the conditional block above, so it is shown only if the toggle is also shown.

packages/itinerary-body/src/styled.tsx
```diff
@@ -660,6 +660,18 @@ export const TransitAlert = styled.a`
...
+export const TransitAlertExternalLink = styled.a`
+  align-items: baseline;
+  display: flex;
+  gap: 5px;
+  margin-top: 0.5em;
+  text-decoration: none;
+
+  &:hover {
+    text-decoration: underline;
+  }
```
binh-dam-ibigroup, line 672: Delete these default behaviors of `<a>` tags.

packages/itinerary-body/src/styled.tsx
```diff
@@ -684,7 +696,7 @@ export const TransitAlertIconContainer = styled.div`
   font-size: 18px;
 `;
 
-export const TransitAlerts = styled.div`
+export const TransitAlerts = styled.ul`
```
binh-dam-ibigroup, line 699: Remove the left margin that is typically included with `<ul>` tags.

binh-dam-ibigroup (changes requested):
See various change requests.

packages/itinerary-body/src/TransitLegBody/alerts-body.tsx
```diff
@@ -131,6 +138,25 @@ export default function AlertsBody({
                       }}
                     />
                   )}
+                  {alertUrl.trim() && (
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Good catch, thank you!

packages/itinerary-body/src/TransitLegBody/alerts-body.tsx
```diff
@@ -131,6 +138,25 @@ export default function AlertsBody({
...
+                      <ExternalLinkAlt height={10} />
+                      <FormattedMessage
+                        defaultMessage={
+                          defaultMessages[
+                            "otpUi.TransitLegBody.AlertsBody.alertLinkText"
+                          ]
+                        }
+                        description="Text with the date an alert takes effect"
+                        id="otpUi.TransitLegBody.AlertsBody.alertLinkText"
+                        values={{ agency: agencyName || defaultAgencyName }}
```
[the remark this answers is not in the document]
amy-corson-ibigroup: That might make more sense depending on the language, changed in [579f01e](https://github.com/opentripplanner/otp-ui/pull/732/commits/579f01e7ae630118e9b77e25ad22ea225bf0eca5)

packages/itinerary-body/src/TransitLegBody/alerts-body.tsx
amy-corson-ibigroup: Good point!

packages/itinerary-body/src/TransitLegBody/index.tsx
amy-corson-ibigroup: Good point, I also changed this toggle to a `div` in cases where the list cannot be collapsed.

packages/itinerary-body/src/styled.tsx
amy-corson-ibigroup: By default there is an `underline`  on the `<a>` regardless of hover. I think I prefer it this way, where the underline only appears on hover, but I don't mind making it just the default behavior if you'd prefer!

binh-dam-ibigroup: Oh okay that's right, then keep that CSS.

packages/itinerary-body/i18n/en-US.yml
```diff
@@ -90,8 +90,12 @@ otpUi:
     viewOnMap: View on map
   TransitLegBody:
     AlertsBody:
+      alertLinkText: See alert on {agency} website
       effectiveDate: Effective as of {dateTime, date, long}
       effectiveTimeAndDate: Effective as of {dateTime, time, short}, {day}
+      externalLink: (External link)
```
binh-dam-ibigroup, line 96: Remove `externalLink`, it does not seem to be used in the code.

binh-dam-ibigroup (approved):
Much better with the changes! Please cleanup the English file to make the tests pass.

PR #733, opened 2024-04-11
Support `search` for OTP Geocoder

miles-grant-ibigroup:
Right now it causes crashes. We use autocomplete to achieve the same behavior


daniel-heppner-ibigroup (approved):
Looks fine, but should we add a warning or some documentation that this is just forwarding to the autocomplete query?

amy-corson-ibigroup (approved):

PR #734, opened 2024-04-12
Add localization and restore walk speed selector to TripOptions component

grant-humphries:
Our team at TriMet is preparing to release trimet.org in Spanish and we found that the UI elements in `TripOptions` that are created from the `GeneralSettingsPanel` could not be localized.  This work adds that localization following the pattern used in `SettingsSelectorPanel`.  This also restores the walk speed selector to the `TripOptions` component.  We're still on version 1 of the OTP engine, though we've started work to move to v2, and would like to retain this feature until we upgrade the engine


packages/core-utils/src/query-params.js
```diff
@@ -365,6 +365,30 @@ const queryParams = [
...
+      {
+        text: "3 MPH",
+        value: 1.34
+      },
+      {
+        text: "4 MPH",
+        value: 1.79
+      }
+    ]
+  },
```
grant-humphries, line 391: hi @daniel-heppner-ibigroup 👋 , I saw that you removed this section of the code last August in [this commit](https://github.com/opentripplanner/otp-ui/commit/2438812f059a325d3f25b030f32d993d4250b6ee).  At TriMet we're still on OTP 1 for now and this UI element is still in the current version of our app so I'm hoping to add it back to the `TripOptions` component only, but want to make sure it doesn't break anything elsewhere in the code base.  

I added some logic to the `applicable` method to exclude it if a query param like `otp2: true` is passed, but not sure if that addresses the reason it was removed.  I also confirmed that it doesn't show up in the `SettingSelectorPanel` with these additions (at least in the Storybook stories for that component)

grant-humphries:
One other thing for the reviewers: this work makes a change to `core-utils` so `trip-form` will need to rely on a forth-coming release of that package. Is there anything I need to do to update the core-utils dependency in trip-form or is that an automated part of the release process?

packages/core-utils/src/query-params.js
grant-humphries: per Miles request I'm now overriding the version of `core-utils` that `trip-form` uses in the trimet.org app to go back to 9.0.3, the last that had the `walkSpeed` param.  So this is now a moot point and there are no longer changes to `core-utils` in this PR

grant-humphries:
> One other thing for the reviewers: this work makes a change to `core-utils` so `trip-form` will need to rely on a forth-coming release of that package. Is there anything I need to do to update the core-utils dependency in trip-form or is that an automated part of the release process?

There are no longer changes to `core-utils` in this PR so this is no longer an issue, but still curious how this works...

miles-grant-ibigroup:
> > One other thing for the reviewers: this work makes a change to `core-utils` so `trip-form` will need to rely on a forth-coming release of that package. Is there anything I need to do to update the core-utils dependency in trip-form or is that an automated part of the release process?
> 
> There are no longer changes to `core-utils` in this PR so this is no longer an issue, but still curious how this works...

We resolve this with a follow up PR that updates the internal dependencies. It's not ideal, but it works

__snapshots__/storybook.test.ts.snap
```diff
@@ -276361,27 +276361,27 @@ exports[`Storyshots TripOptions Company First Mixed Category 2`] = `
             <option
               value={160.9}
             >
-              1/10 mile
+              0.1 miles
```
miles-grant-ibigroup, line 276364: Are these snapshot updates still needed?

miles-grant-ibigroup (changes requested):
Working for me! Just a quick question about these snapshot updates

miles-grant-ibigroup (approved):
Sounds good to me!

packages/trip-form/i18n/es.yml
```diff
@@ -39,6 +39,9 @@ otpUi:
     use: Usar
     useCompanies: Usar las empresas
     walkOnly: Solo caminar
+  TripOptions:
+    transitOnly: Ir en transporte público
+    openApp: Abrir aplicación
```
grant-humphries, line 44: @miles-grant-ibigroup I found a couple of places where there was hard-coded text in `TripOptions` and hooked that into the i18n setup.  

How do you handle generating the strings in all of the languages that you support?  I've only done English and Spanish so far and used ChatGPT to generate the latter

miles-grant-ibigroup: Just saw. Looks good! Thanks.

We pay professional translators

binh-dam-ibigroup:
@grant-humphries This CI process enforces French content completeness. The new text for French is:
```
  TripOptions:
    transitOnly: En transports publics
    openApp: Ouvrir l'application
```

packages/trip-form/i18n/es.yml
grant-humphries: I just added French, also via ChatGPT, to get the CI to pass.  Does this cover our bases with the i18n strings enough for this to be merged?

binh-dam-ibigroup: @grant-humphries Should do for this time, however in other instances, you should check that the produced text fits in the space available, or is not disproportionally longer than the original text. (French and Spanish can produce long sentences on direct translations.)

packages/trip-form/i18n/fr.yml
```diff
@@ -39,6 +39,9 @@ otpUi:
     use: Utiliser
     useCompanies: Prestataires
     walkOnly: À pied uniquement
+  TripOptions:
+    transitOnly: Aller en transport en commun
```
binh-dam-ibigroup, line 43: "Transport" is used in plural form in French, and to save space, you may strip the verb "go" which is implied.
```suggestion
  transitOnly: En transports en commun
```

grant-humphries: Thanks Binh, just updated to what you supplied.  Does the Spanish look ok to you?

binh-dam-ibigroup: It doesn't look obviously wrong, but you should confirm with a competent person if that is going on the TriMet site.

grant-humphries: we don't actually use the non-english files from this project in our i18n implementation, but I ran it by a Spanish speaker from our team and they said it looks ok

PR #735, opened 2024-04-18
Implement the new GraphQL elevationProfile type 

daniel-heppner-ibigroup:
This implements the type from #729. It also adds a test to ensure backwards compatibility is working properly.


daniel-heppner-ibigroup:
@binh-dam-ibigroup this is ready for a review!

packages/core-utils/src/itinerary.ts
```diff
@@ -329,30 +330,42 @@ export function getElevationProfile(
...
   steps.forEach(step => {
-    if (!step.elevation || step.elevation.length === 0) {
+    // Support for old REST response data (in step.elevation)
+    const stepElevationProfile =
+      step.elevationProfile ||
+      (Array.isArray(step.elevation) &&
+        step.elevation?.map<ElevationProfileComponent>(elev => ({
+          distance: elev.first,
+          elevation: elev.second
+        })));
```
binh-dam-ibigroup, line 343: Extract a method for this conversion, and reuse in the test.

binh-dam-ibigroup (commented):
Elevation code changes look good, but did we need to touch all the package.json files? The reason I ask is there is only one `feat` commit for core-utils, and as a result, no new binaries will be generated for the other modified packages.

binh-dam-ibigroup (approved):
It is actually fine if no binaries are released for the other packages.

daniel-heppner-ibigroup:
That's intentional! Just ran `update-internal-dependencies`

packages/core-utils/src/itinerary.ts
```diff
@@ -329,30 +340,41 @@ export function getElevationProfile(
...
+      const elev = stepElevationProfile[i];
       if (previous) {
-        const diff = (elev.second - previous.second) * unitConversion;
+        const diff = (elev.elevation - previous.elevation) * unitConversion;
         if (diff > 0) gain += diff;
         else loss += diff;
       }
-      if (i === 0 && elev.first !== 0) {
+      if (i === 0 && elev.distance !== 0) {
         // console.warn(`No elevation data available for step ${stepIndex}-${i} at beginning of segment`, elev)
```
amy-corson-ibigroup, line 366: Is there a point to keeping this commented out `console.warn`?

amy-corson-ibigroup (approved):
Looks good! Even just changing `first` and `second` to `distance` and `elevation` is so much more clear

packages/core-utils/src/itinerary.ts
daniel-heppner-ibigroup: IDK but it's not really relevant to this PR so I don't wanna touch it

PR #737, opened 2024-04-29
fix(location-field): Add overflow to results container

amy-corson-ibigroup:
Set a max-height on the location field results so that users can scroll in cases where the list is too long

### DIFF
--- a/packages/location-field/src/styled.ts
+++ b/packages/location-field/src/styled.ts
@@ -46,8 +46,10 @@ export const MenuItemList = styled.ul.attrs({
   list-style: none;
   margin: 2px 0 0;
   min-width: 160px;
+  max-height: 75vh;
   padding: 5px 0;
   position: absolute;
+  overflow: scroll;
   text-align: left;
   top: 100%;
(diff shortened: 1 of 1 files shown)


packages/location-field/src/styled.ts
```diff
@@ -46,8 +46,10 @@ export const MenuItemList = styled.ul.attrs({
   list-style: none;
   margin: 2px 0 0;
   min-width: 160px;
+  max-height: 75vh;
```
miles-grant-ibigroup, line 49: Does this work if you increase the number of responses shown?

amy-corson-ibigroup: I don't think that would impact it! I tried with a few different environments, and the container stayed consistent regardless of the geocoder results. Let me know if there's something specific I can test though

miles-grant-ibigroup (approved):
A little antsy about 75vh but have tested and nothing seemed to blow up so I think good enough for now

daniel-heppner-ibigroup (approved):
Hope this doesn't cause any unforeseen problems!

PR #743, opened 2024-05-30
Add Focus Traps to all map popups

amy-corson-ibigroup:
Updates all map popups to include focus traps.

### DIFF
--- a/packages/endpoints-overlay/src/endpoint.tsx
+++ b/packages/endpoints-overlay/src/endpoint.tsx
@@ -15,6 +15,7 @@ import { Times } from "@styled-icons/fa-solid/Times";
 import coreUtils from "@opentripplanner/core-utils";
 import flatten from "flat";
 import React, { ComponentType, useState } from "react";
+import { FocusTrapWrapper } from "@opentripplanner/map-popup/lib";
 
 import * as S from "./styled";
 
@@ -176,7 +177,10 @@ const Endpoint = (props: Props): JSX.Element => {
           latitude={location.lat}
           longitude={location.lon}
         >
-          <div>
+          <FocusTrapWrapper
+            id="endpoint-overlay"
+            closePopup={() => setShowPopup(false)}
+          >
             <strong>
               <UserLocationIcon type={icon} />
               {location.name}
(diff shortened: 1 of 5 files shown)


packages/endpoints-overlay/src/endpoint.tsx
```diff
@@ -176,7 +177,10 @@ const Endpoint = (props: Props): JSX.Element => {
           latitude={location.lat}
           longitude={location.lon}
         >
-          <div>
+          <FocusTrapWrapper
```
binh-dam-ibigroup, line 180: Not seeing the popup for `Endpoint`, also not on the master branch. Should we fix that?

amy-corson-ibigroup: Do you mean the popup isn't showing up in the story?

binh-dam-ibigroup: Correct

amy-corson-ibigroup: Fixed in [2e969ce](https://github.com/opentripplanner/otp-ui/pull/743/commits/2e969ce70f0c33f05265da21da83f7dde0953917)

packages/endpoints-overlay/src/endpoint.tsx
```diff
@@ -15,6 +15,7 @@ import { Times } from "@styled-icons/fa-solid/Times";
 import coreUtils from "@opentripplanner/core-utils";
 import flatten from "flat";
 import React, { ComponentType, useState } from "react";
+import { FocusTrapWrapper } from "@opentripplanner/map-popup/lib";
```
binh-dam-ibigroup, line 18: Move this import to after line 15 (this repo imports by package name).

binh-dam-ibigroup:
> Sorry, looks like node caught a circular dependency in the package references. Probably `FocusTrapWrapper` shouldn't be used from within base-map.

Alternately (dilemma?), the map-popup package can duplicate some styles it imports from base-map so that base-map can use `FocusTrapWrapper`.

binh-dam-ibigroup (approved):
Seems to work well now, thanks for the changes!

josh-willis-arcadis (approved):
LGTM!

PR #744, opened 2024-05-30
Create Alert Component

amy-corson-ibigroup:
Creates a basic customizable alert in building blocks 

image
image
image


packages/building-blocks/src/Alert.tsx
```diff
@@ -0,0 +1,124 @@
...
+interface Props {
+  alertHeader: string;
+  alertSubheader?: string;
+  backgroundColor?: string;
+  collapsable?: boolean;
+  children?: ReactChild;
+  Icon?: any;
+}
+
+const AlertContainer = styled.div<{
```
miles-grant-ibigroup, line 17: Should we consider going more semantic and using a [details](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) tag?

packages/building-blocks/src/stories/alerts.story.tsx
```diff
@@ -0,0 +1,76 @@
...
+  }
+];
+
+const AlertContent = () => {
+  return (
+    <>
+      {alerts.map((alert, i) => {
+        return (
+          <>
+            <div key={alert.alertDescriptionText}>
```
miles-grant-ibigroup, line 42: Let's maybe just use the counter/UUID. Using a huge body might be funky

packages/building-blocks/src/Alert.tsx
```diff
@@ -0,0 +1,124 @@
...
+import { Bell } from "@styled-icons/bootstrap/Bell";
+import blue from "./colors/blue";
+
+interface Props {
+  alertHeader: string;
+  alertSubheader?: string;
+  backgroundColor?: string;
+  collapsable?: boolean;
+  children?: ReactChild;
+  Icon?: any;
```
miles-grant-ibigroup, line 14: This can probably be a `JSX.Element` or something right?

packages/building-blocks/package.json
```diff
@@ -27,5 +28,8 @@
   "bugs": {
     "url": "https://github.com/opentripplanner/otp-ui/issues"
   },
-  "gitHead": "0af1b7cda60bd4252b219dcf893e01c2acb2ed5d"
+  "gitHead": "0af1b7cda60bd4252b219dcf893e01c2acb2ed5d",
+  "dependencies": {
+    "styled-icons": "^10.47.1"
```
miles-grant-ibigroup, line 33: This is causing a huge `yarn.lock` increase. We should probably only depend on the one icon set we need

miles-grant-ibigroup (changes requested):
Looks great! The only blocking thing for me is the huge `yarn.lock` size increase

packages/building-blocks/src/Alert.tsx
amy-corson-ibigroup: Hmm I think this would actually make the component less customizable, since `details` always has a toggle, and this alert currently only has a toggle if `collapsible` is true

packages/building-blocks/src/Alert.tsx
amy-corson-ibigroup: Whoops!! Yes

miles-grant-ibigroup (approved):
Safe enough for now!

packages/building-blocks/src/Alert.tsx
daniel-heppner-ibigroup: is it possible to hide the toggle with CSS?

packages/building-blocks/src/stories/alerts.story.tsx
```diff
@@ -0,0 +1,78 @@
...
+  },
+  {
+    alertUrl: "http://trimet.org/alerts/",
+    effectiveStartDate: 1572827580,
+    alertDescriptionText:
+      "The west elevators at the Washington Park MAX Station are out of service. Please use east elevators to access street level and platforms. "
+  }
+];
+
+const alertKey = crypto.randomUUID();
```
daniel-heppner-ibigroup, line 36: Why not just use `i` for the key? Does that break something?

daniel-heppner-ibigroup: actually yeah this is bad because all the keys will all be the same!

daniel-heppner-ibigroup (changes requested):
can we fix the issue with the key?

packages/building-blocks/src/stories/alerts.story.tsx
amy-corson-ibigroup: Good catch!! I moved this into the `map` so it actually does what it's supposed to do, although I guess `i` would work in this case because the order of the mock alerts in this story isn't going to change? Happy to change it if someone feels strongly?

daniel-heppner-ibigroup: Actually I didn't realize this is just the story. It's fine, stories don't have to be perfect!

daniel-heppner-ibigroup (approved):
Looks good!

PR #746, opened 2024-06-04
fix(core-utils): support `unpreferred` query param

miles-grant-ibigroup:
We even have this in our plan query file! How did we not support setting this!

### DIFF
--- a/packages/core-utils/src/query-gen.ts
+++ b/packages/core-utils/src/query-gen.ts
@@ -223,9 +223,9 @@ export function generateOtp2Query(
     modeSettings,
     numItineraries,
     preferred,
-    unpreferred,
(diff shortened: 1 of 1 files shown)


packages/core-utils/src/query-gen.ts
```diff
@@ -222,6 +223,7 @@ export function generateOtp2Query(
     modeSettings,
     numItineraries,
     preferred,
+    unpreferred,
```
binh-dam-ibigroup, line 226: Sort attributes (2x).

binh-dam-ibigroup (approved):
Cool

amy-corson-ibigroup (approved):
makes sense!

PR #747, opened 2024-06-04
fix(core-utils): optimize flex combination generation

daniel-heppner-ibigroup:
This accomplishes two things. First, it only allows FLEX_DIRECT on its own, without transit or anything else. Although this should work, OTP has a bug that causes issues with this configuration. It doesn't make sense to combine them anyway.

Second, it makes FLEX (access/egress) *only* comaptible with TRANSIT. As far as I can tell, OTP doesn't support FLEX combined with bikeshare or anything else. This should reduce the number of queries we send to OTP, which will improve performance.

Lastly, it improves the test output to make it easier to read.

### DIFF
--- a/packages/core-utils/src/__tests__/query-params.ts
+++ b/packages/core-utils/src/__tests__/query-params.ts
@@ -22,6 +22,12 @@ function modeStrToTransportMode(m): TransportMode {
     qualifier: splitVals?.[1] || null
   };
 }
+function transportModeToModeStr(m: TransportMode): string {
+  if (m.qualifier) {
+    return `${m.mode}_${m.qualifier}`;
+  }
+  return m.mode;
+}
 
 const mockLatLon = {
   lat: 1,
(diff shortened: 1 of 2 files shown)


miles-grant-ibigroup:
A few things

1) FLEX_DIRECT + BIKE is definitely a combination I've seen before. Flex and scooters as well, thats' why they're in the unit test!
2) I agree flex access/egress should only be compatible with transit, but I don't think this approach is the way to do it. Can't we just reclassify these two in the same category as scooters/bikes?

I am hesitant to make these fundamental changes to the core of otp-ui when this is a performance problem that needs to be solved within OTP. Reducing the number of requests to the server isn't going to help performance, especially when we support load balancers

daniel-heppner-ibigroup:
Not sure about 1. Will have to investigate more.

For (2) I think this is a good approach, since it always felt like a hack to me to bundle flex with scooters and bike share. It really feels like its own thing. Especially when FLEX_DIRECT has to be treated separately. 

As for performance, I sort of agree, but it is actually quite expensive to spin up more of these large instances that are used for flex-enabled OTP. We need to approach the performance problem from every angle, and especially this one if the extra requests aren't yielding any different results.

daniel-heppner-ibigroup:
I think we've confirmed now that FLEX_DIRECT should also be alone. It doesn't make sense to add other modes since it's specifically door to door flex. For ACCESS and EGRESS, we might want to combine it with other access and egress modes.

PR #754, opened 2024-06-25
Add SEPTA Icons

miles-grant-ibigroup:
Septa METRO has some exciting icons that are now available to all OTP UI users. 
Screenshot 2024-06-25 at 12 38 22 PM

Also adds some itinerary body styling support


josh-willis-arcadis (approved):
LGTM

miles-grant-ibigroup:
@daniel-heppner-ibigroup how do we do it with the icons again will this support Storybook 7?

daniel-heppner-ibigroup (approved):
LGTM, the key for making this work with Storybook 7 was that the file name has to match the default export. Also, why not make these .tsx files while we're at it so that we get types when they're imported?

PR #755, opened 2024-07-03
remove metro mode on hover floating dropdown

josh-willis-arcadis:
remove metro mode on hover floating dropdown. This functionality will be moved to the advanced settings panel.

### DIFF
--- a/packages/trip-form/src/MetroModeSelector/index.tsx
+++ b/packages/trip-form/src/MetroModeSelector/index.tsx
@@ -1,13 +1,10 @@
 import {
   arrow,
-  FloatingFocusManager,
   offset,
-  safePolygon,
   shift,
   useClick,
   useDismiss,
   useFloating,
-  useHover,
   useInteractions,
   useRole
 } from "@floating-ui/react";
@@ -18,7 +15,7 @@ import React, { ReactElement, useCallback, useRef, useState } from "react";
 import { FormattedMessage, useIntl } from "react-intl";
 import styled, { css } from "styled-components";
 
-import SubSettingsPane, { defaultMessages } from "./SubSettingsPane";
+import { defaultMessages } from "./SubSettingsPane";
 import generateModeButtonLabel from "./i18n";
 
 const invisibleCss = css`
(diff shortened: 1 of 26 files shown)


packages/trip-form/src/MetroModeSelector/index.tsx
```diff
@@ -220,38 +186,18 @@ function ModeButton({
...
-      // Enable hover only if no popup has been triggered via keyboard.
-      // (This is to avoid focus being stolen by hovering out of another button.)
-      enabled: itemWithKeyboard === null && hoverEnabled,
-      handleClose: safePolygon({
-        blockPointerEvents: false,
-        restMs: 500,
-        buffer: 0
-      })
-    }),
+  const { getReferenceProps } = useInteractions([
```
daniel-heppner-ibigroup, line 195: I think you can remove ALL of this! We don't need the floating UI stuff at all anymore since it was only being used for this. I think even the dependency can go!

yarn.lock
```diff
@@ -3246,6 +3246,11 @@
   resolved "https://registry.yarnpkg.com/@open-draft/until/-/until-1.0.3.tgz#db9cc719191a62e7d9200f6e7bab21c5b848adca"
   integrity sha512-Aq58f5HiWdyDlFffbbSjAlv596h/cOnt2DO1w3DOC7OJ5EHs0hd/nycJfiu9RJbT6Yk6F1knnRRXNSpxoIVZ9Q==
 
+"@opentripplanner/types@6.5.0":
```
daniel-heppner-ibigroup, line 3249: that's weird... maybe run update-internal-dependencies while you're at it

packages/trip-form/src/MetroModeSelector/index.tsx
```diff
@@ -1,13 +1,10 @@
...
   offset,
-  safePolygon,
   shift,
   useClick,
   useDismiss,
   useFloating,
-  useHover,
   useInteractions,
   useRole
 } from "@floating-ui/react";
```
daniel-heppner-ibigroup, line 10: delete it all!

daniel-heppner-ibigroup (changes requested):
I think there's a lot more that can be removed!

daniel-heppner-ibigroup (changes requested):
Check out the results of `yarn check:i18n-en-fr`. It looks like there are extra i18n values that need to be cleaned up. I also think you need a breaking change commit in here. Feel free to edit one of your previous commit messages and force push it, or just do it with your i18n changes in trip-form.

PR #760, opened 2024-07-25
fix(core-utils): Remove extra trailing bracket from GraphQL place vars

binh-dam-ibigroup:
We are (inadvertently?) imserting an extra bracket into the GraphQL `fromPlace` and `toPlace` parameters, which can cause parsing issues. Proposing to remove these extra bracket with this PR.


josh-willis-arcadis (approved):
LGTM!

amy-corson-ibigroup (approved):
Oooh good catch!

PR #761, opened 2024-07-30
fix(location-field): set overflow to auto

amy-corson-ibigroup:
Sorry for tiny tiny PR but this will fix the appearance of the scroll bar on windows! I also tweaked some of the padding within the menu.

### DIFF
--- a/packages/location-field/src/styled.ts
+++ b/packages/location-field/src/styled.ts
@@ -49,7 +49,7 @@ export const MenuItemList = styled.ul.attrs({
   max-height: 75vh;
   padding: 5px 0;
   position: absolute;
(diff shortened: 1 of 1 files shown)


binh-dam-ibigroup (approved):
Approving this because it looks good. At some point, in a separate PR, could you address making items visible during keyboard navigation?

amy-corson-ibigroup:
Looking into this, it seems like the easiest way to handle scrolling on keyboard navigation will be to add focus to the `activeIndex` element, which may be best served by using `FocusTrapWrapper`. I'll ticket this for a future PR

josh-willis-arcadis (approved):
Looks good!

PR #762, opened 2024-07-31
fix(itin-body): support duration in walk leg string override

miles-grant-ibigroup:
The itinerary body walk legs currently show distance. If a user wants to override the string to display duration instead, they'll need that value. Sadly, it's only in seconds. To fix this, we need a formatDuration method. That method needs to be passed in. Therefore this PR is quite large for doing functionally nothing, but it opens the door to exciting features in the future


josh-willis-arcadis (approved):
Looks great!

amy-corson-ibigroup (approved):
Beautiful

PR #763, opened 2024-08-07
fix: don’t show stops that are within stations

miles-grant-ibigroup:
Sometimes stops are a part of stations. OTP now exposes this, so we can filter them out.


josh-willis-arcadis (approved):
looks great!

binh-dam-ibigroup (approved):

PR #764, opened 2024-08-09
Map Popup show company name in all map popup titles

miles-grant-ibigroup:
It was a little confusing that there was no title in the station names. Especially when feeds had terrible station names. We resolve this. As a result we remove company name from the other strings. However, the car isn't localized, so it's kept for now. In the future we might want to get rid of that

### DIFF
--- a/packages/map-popup/i18n/en-US.yml
+++ b/packages/map-popup/i18n/en-US.yml
@@ -3,7 +3,8 @@ otpUi:
     availableBikes: "Available bikes: {value}"
     availableDocks: "Available docks: {value}"
     floatingBike: "Free-floating bike: {name}"
-    floatingCar: "{company} {name}"
+    floatingCar: "Car: {name}"
     floatingEScooter: "E-scooter: {name}"
+    popupTitle: "{stationNetwork, select, false {{name}} other {{stationNetwork} {name}}}"
     stopId: "Stop ID: {stopId}"
     stopViewer: Stop Viewer
--- a/packages/map-popup/i18n/fr.yml
+++ b/packages/map-popup/i18n/fr.yml
@@ -3,7 +3,7 @@ otpUi:
     availableBikes: "Vélos disponibles : {value}"
     availableDocks: "Bornes disponibles : {value}"
     floatingBike: "Vélo flottant : {name}"
-    floatingCar: "{company} {name}"
+    floatingCar: "Voiture : {name}"
     floatingEScooter: Trottinette {name}
     popupTitle: "{stationNetwork, select, false {{name}} other {{stationNetwork} {name}}}"
     stopId: Arrêt n°{stopId}
--- a/packages/map-popup/i18n/fr.yml
+++ b/packages/map-popup/i18n/fr.yml
@@ -5,5 +5,6 @@ otpUi:
     floatingBike: "Vélo flottant : {name}"
     floatingCar: "{company} {name}"
     floatingEScooter: Trottinette {name}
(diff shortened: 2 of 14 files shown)


miles-grant-ibigroup:
@daniel-heppner-ibigroup I need your help updating the snapshtos

daniel-heppner-ibigroup (approved):
Good change!

packages/map-popup/i18n/en-US.yml
```diff
@@ -3,7 +3,8 @@ otpUi:
     availableBikes: "Available bikes: {value}"
     availableDocks: "Available docks: {value}"
     floatingBike: "Free-floating bike: {name}"
-    floatingCar: "{company} {name}"
+    floatingCar: "Car: {name}"
```
binh-dam-ibigroup, line 6: Should this change apply to other languages too?

packages/map-popup/i18n/fr.yml
```diff
@@ -5,5 +5,6 @@ otpUi:
     floatingBike: "Vélo flottant : {name}"
     floatingCar: "{company} {name}"
     floatingEScooter: Trottinette {name}
+    popupTitle: "{stationNetwork, select, false {{name}} other {{stationNetwork} {name}}}"
```
binh-dam-ibigroup, line 8: If `popupTitle` is going to be telegraphic style at first, does it need to be in the language files?

binh-dam-ibigroup (changes requested):
I think you need to apply the "Car" labeling to the other languages.

packages/map-popup/i18n/en-US.yml
miles-grant-ibigroup: I made the change backwards compatible so I don't have to make all the translations. I don't feel confident translating for all those languages

packages/map-popup/i18n/fr.yml
miles-grant-ibigroup: Same as above. I've made it backwards compatible so languages can adjust when they're ready. Should we consider removing the strings from other languages?

packages/map-popup/i18n/en-US.yml
binh-dam-ibigroup, line 6: Is the colon needed?

miles-grant-ibigroup: Without the colon things can get messy especially if the company name is silly

packages/map-popup/src/MapPopup.story.tsx
binh-dam-ibigroup, line 1: Add a story for the car case, for instance:
```js
export const FloatingCarEntity = (): JSX.Element => (
  <MapPopupContents entity={FLOATING_CAR} setLocation={action("setLocation")} />
);
```
where
```js
const FLOATING_CAR = {
  "stroke-width": 1,
  allowDropoff: false,
  allowPickup: true,
  color: "#f00",
  id: "car_6861",
  isCarStation: false,
  isFloatingCar: true,
  name: "0541",
  networks: ["CARONDEMAND"],
  realTimeData: true,
  spacesAvailable: 0,
  x: -122.70486,
  y: 45.525486666666666
};
```

binh-dam-ibigroup (approved):
See if you can add the additional story to illustrate the updated floating car case. Otherwise I am good with the changes.

packages/map-popup/src/MapPopup.story.tsx
miles-grant-ibigroup: Good call! Added

miles-grant-ibigroup:
@daniel-heppner-ibigroup I'm going to need you to please update the snapshots again

PR #765, opened 2024-08-09
feat(building-blocks): add focus trap to building blocks

amy-corson-ibigroup:
Housing Focus Trap Wrapper in map popup doesn't make a ton of sense and is causing circular dependencies and duplicated code.

This version of focus trap also allows for some configuration of which elements are focusable within the trap. For example, if you're creating non-standard focusable elements with `tabIndex`, those can be included.

### DIFF
--- a/packages/building-blocks/src/dropdown/index.tsx
+++ b/packages/building-blocks/src/dropdown/index.tsx
@@ -7,7 +7,10 @@ import React, {
   useState
 } from "react";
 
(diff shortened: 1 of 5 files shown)


binh-dam-ibigroup (commented):
Are there settings that changed that are causing the build to fail?

binh-dam-ibigroup (approved):

josh-willis-arcadis (approved):
LGTM

PR #768, opened 2024-08-27
fix(geocoder/otp): more clever agency name selection

miles-grant-ibigroup:
Instead of picking the first agency (randomly selected) we instead look at other factors to generate an excellently fitting agency name.

### DIFF
--- a/packages/geocoder/src/geocoders/otp.ts
+++ b/packages/geocoder/src/geocoders/otp.ts
@@ -6,8 +6,20 @@ import Geocoder from "./abstract-geocoder";
 import { OTPGeocoderResponse } from "../apis/otp";
 
 const generateLabel = stop => {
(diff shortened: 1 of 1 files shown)


amy-corson-ibigroup (approved):
This makes sense to me, but it's a little hard to tell how it's going to work in practice when there aren't any stories or typing that illustrate this part of the code. Is there a type for this `stop` object anywhere?

miles-grant-ibigroup:
> This makes sense to me, but it's a little hard to tell how it's going to work in practice when there aren't any stories or typing that illustrate this part of the code. Is there a type for this `stop` object anywhere?

Stories will come once we update internal dependencies and use this within the location field

josh-willis-arcadis (approved):
beautiful!

PR #770, opened 2024-09-04
Update internal focus traps

amy-corson-ibigroup:
Focus trap has moved from map-popup to building blocks, and this change reflects that


packages/map-popup/src/index.tsx
```diff
@@ -113,6 +115,7 @@ export function MapPopup({ configCompanies, entity, getEntityName, setLocation,
 
   return (
     <S.MapOverlayPopup>
+      <FocusTrapWrapper closePopup={closePopup} id={`${entity.id}-popup`}>
```
binh-dam-ibigroup, line 118: This addition is producing an undesired side effect. When the "Stop Viewer" button or the "From Here"/"To Here" button get focused, the focus is stuck on that button. I think that's because `FocusTrapWrapper` only moves the focus to siblings, and the focus has to stay within a containing element.

Not sure what the fix for this should be without fixing FocusTrapWrapper, so should we revert that for now?

packages/map-popup/src/index.tsx
```diff
@@ -142,11 +145,13 @@ export function MapPopup({ configCompanies, entity, getEntityName, setLocation,
...
     </S.MapOverlayPopup>
   );
 }
 
 export default MapPopup;
 
 // Rename styled components for export.
-export { S as Styled, FocusTrapWrapper };
\ No newline at end of file
+export { S as Styled };
```
binh-dam-ibigroup, line 157: There should be a breaking change commit for the removed export.

binh-dam-ibigroup (changes requested):
`FocusTrapWrapper` does not play well with components in `map-popup`, do you want to revert that?

packages/map-popup/src/index.tsx
amy-corson-ibigroup: Actually, I think the issue was coming from double quotes inside the mock data ids: 
`id: '"hub_1580"',`
I wrote a line in map-popup to remove these for the purposes of the focus trap wrapper and the problem seems to be solved.... maybe it's not in the right spot? Let me know what you think

binh-dam-ibigroup: Good find! I'm good with that.

binh-dam-ibigroup (approved):

binh-dam-ibigroup:
@amy-corson-ibigroup You might need to update snapshots.

josh-willis-arcadis (approved):
LGTM!

PR #772, opened 2024-09-06
Support stop filtering

miles-grant-ibigroup:
Adds a new prop to the OTP2 tile layer to support a stop whitelist. If this prop is active, only stops with this id will be shown. Zoom level is also overridden.

### DIFF
--- a/packages/otp2-tile-overlay/src/index.tsx
+++ b/packages/otp2-tile-overlay/src/index.tsx
@@ -24,6 +24,7 @@ const OTP2TileLayerWithPopup = ({
   onMapClick,
   setLocation,
   setViewedStop,
+  stopsWhitelist,
   type
 }: {
   color?: string;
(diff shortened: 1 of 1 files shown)


packages/otp2-tile-overlay/src/index.tsx
```diff
@@ -135,6 +141,9 @@ const OTP2TileLayerWithPopup = ({
   if (type === "stops" || type === "areaStops") {
     filter = ["all", ["!", ["has", "parentStation"]], ["!=", ["get", "routes"], ["literal", "[]"]]]
   }
+  if (stopsWhitelist) {
+    filter = ["in", ["get", "gtfsId"], ["literal", stopsWhitelist]]
+  }
```
daniel-heppner-ibigroup, line 146: will it ever be a problem if this is overriding the previous filters set or is this intentional?

daniel-heppner-ibigroup (changes requested):
Looks good but I can't see anything in the story. Can you check on the story and if possible add a knob for this?

packages/otp2-tile-overlay/src/index.tsx
miles-grant-ibigroup: We want to disable the old ones. We want to show all stops even if there is a parent station! And showing stops without routes is impossible (since we are generating the whitelist from route data) but even if not I like that the whitelist overrides existing filters

daniel-heppner-ibigroup (approved):
Looks good

josh-willis-arcadis (approved):
LGTM

PR #776, opened 2024-09-13
fix: `isFlex` truthy more of the time

miles-grant-ibigroup:
isFlex isn't super reliable at the minute. This PR repairs this by increasing the number of conditions we use to accept a leg is flexy


binh-dam-ibigroup (approved):
LGTM

josh-willis-arcadis (approved):
LGTM

PR #777, opened 2024-09-16
Fix Map Popups (and focus traps)

amy-corson-ibigroup:
Fixes IDs that are breaking due to stop codes, passes `closePopup` to `MapPopup` through `otp2-tile-layer`, prevent crash on null `closePopup`

### DIFF
--- a/packages/map-popup/src/index.tsx
+++ b/packages/map-popup/src/index.tsx
@@ -101,7 +101,7 @@ function entityIsStation(entity: Entity): entity is Station {
 /**
  * Renders a map popup for a stop, scooter, or shared bike
  */
-export function MapPopup({ closePopup = null, configCompanies, entity, getEntityName, setLocation, setViewedStop }: Props): JSX.Element {
+export function MapPopup({ closePopup = () => null, configCompanies, entity, getEntityName, setLocation, setViewedStop }: Props): JSX.Element {
   const intl = useIntl()
   if (!entity) return <></>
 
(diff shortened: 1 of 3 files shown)


packages/map-popup/src/index.tsx
```diff
@@ -101,7 +101,7 @@ function entityIsStation(entity: Entity): entity is Station {
 /**
  * Renders a map popup for a stop, scooter, or shared bike
  */
-export function MapPopup({ closePopup = null, configCompanies, entity, getEntityName, setLocation, setViewedStop }: Props): JSX.Element {
+export function MapPopup({ closePopup = () => null, configCompanies, entity, getEntityName, setLocation, setViewedStop }: Props): JSX.Element {
```
miles-grant-ibigroup, line 104:
```suggestion
export function MapPopup({ closePopup = () => void, configCompanies, entity, getEntityName, setLocation, setViewedStop }: Props): JSX.Element {
```
Should this be void instead?

miles-grant-ibigroup (changes requested):
Small but crucial nitpick

packages/map-popup/src/index.tsx
amy-corson-ibigroup: This isn't a type it's a default!

miles-grant-ibigroup: We have a noOp function you can import

miles-grant-ibigroup (approved):
This looks great! But CI seems insanely broken? What's going on

josh-willis-arcadis (approved):

PR #778, opened 2024-09-18
feat(core-utils): add isTransitLeg function that uses transitLeg prop

daniel-heppner-ibigroup:
This adds a new function (to avoid a breaking change) that checks for transit legs using the OTP2 `transitLeg` property. We could add extra functionality to this in the future to check for transit legs.

To avoid breaking changes in the future, is there any chance that we would want to support other arguments or return types in the future?


miles-grant-ibigroup (approved):
Smashing

binh-dam-ibigroup (approved):

PR #779, opened 2024-09-23
Transfer leg support

miles-grant-ibigroup:
Supports rendering "transfer" legs correctly. They are just informational, there's no directions that should be shown


packages/itinerary-body/src/defaults/access-leg-description.tsx
```diff
@@ -111,6 +113,23 @@ export default function AccessLegDescription({
...
+          description="Summarizes an transfer leg"
+          id="otpUi.AccessLegBody.transfer"
+          values={{
+            duration: intl.formatMessage(
+              {
+                id: "otpUi.ItineraryBody.common.durationShort"
+              },
+              {
+                approximatePrefix: false,
+                ...coreUtils.time.toHoursMinutesSeconds(leg.duration)
```
binh-dam-ibigroup, line 128: Nit: I see a call to `formatDuration` above, could we use that here?

binh-dam-ibigroup (approved):
LGTM, see minor comments.

packages/itinerary-body/src/defaults/access-leg-description.tsx
miles-grant-ibigroup: Both behave slightly differently! Also `formatDuration` is not guaranteed to be present

packages/itinerary-body/src/defaults/access-leg-description.tsx
```diff
@@ -111,6 +113,23 @@ export default function AccessLegDescription({
             place: placeContent
           }}
         />
+      ) : isTransferLeg ? (
+        <FormattedMessage
+          defaultMessage="Transfer, wait {duration}"
+          description="Summarizes an transfer leg"
```
daniel-heppner-ibigroup, line 119:
```suggestion
          description="Summarizes a transfer leg"
```

daniel-heppner-ibigroup (approved):
Code looks good, wondering if we should really display the walk icon for transfer legs.

miles-grant-ibigroup:
What icon are you looking for?

miles-grant-ibigroup:
So the snapshots are making it seem like there's some weird stuff going on with TNC legs, but this seems to be a snapshot issue more than an actual issue.

PR #781, opened 2024-09-25
Make query-gen agnostic to OTP plan query params

binh-dam-ibigroup:
This PR makes query-gen agnostic to OTP query params it doesn't process.

When merged, #696 could be closed.


daniel-heppner-ibigroup (approved):
This is *so* much better. Can't believe I didn't do this originally, looks so obvious now.

PR #784, opened 2024-09-27
fix build issues by using alpha version

daniel-heppner-ibigroup:
The master branch is completely messed up, so we need to use an alpha build to fix it. After this PR is merged we can eliminate the alpha dependency. 

we don't need a semantic release fix commit because it will release based on previous commits already in master once the build succeeds.

### DIFF
--- a/packages/otp2-tile-overlay/package.json
+++ b/packages/otp2-tile-overlay/package.json
@@ -16,10 +16,10 @@
     "react-map-gl": "^7.0.15"
   },
   "dependencies": {
-    "@opentripplanner/map-popup": "^4.0.0"
+    "@opentripplanner/map-popup": "^v3.2.0-alpha.1"
   },
   "devDependencies": {
-    "@opentripplanner/base-map": "^3.2.2",
+    "@opentripplanner/base-map": "^4.0.0",
     "@opentripplanner/types": "^6.5.2"
   }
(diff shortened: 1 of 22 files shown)


packages/otp2-tile-overlay/package.json
```diff
@@ -16,10 +16,10 @@
     "react-map-gl": "^7.0.15"
   },
   "dependencies": {
-    "@opentripplanner/map-popup": "^4.0.0"
+    "@opentripplanner/map-popup": "^v3.2.0-alpha.1"
```
daniel-heppner-ibigroup, line 19: this is the fix

amy-corson-ibigroup (approved):
thank you! And you don't need to update snapshots, right?

daniel-heppner-ibigroup:
So, it looks like Storybook was finding the .story files buried inside node modules. This was happening because of the copy all files flag we added to Babel to support the custom image in transitive overlay. I clamped down the glob used to find stories a bit by telling it to only look in the src directories of packages. Is this too restrictive? Let me know maybe we can adjust it. 

That image in the overlay keeps biting us

miles-grant-ibigroup (approved):
Great fix thanks. That glob wasn't reviewed properly when we approved it! This one looks good to me.

PR #786, opened 2024-10-09
deps: update react-dom

daniel-heppner-ibigroup:
Someone noticed that we have some old react-dom versions in our packages. This is bad! We need to get all those in sync with the react version (which is 18).


miles-grant-ibigroup (approved):
Great!

amy-corson-ibigroup (approved):
looking great!

PR #788, opened 2024-10-10
fix(EndPointsOverlay): Make setLocation optional

binh-dam-ibigroup:
In response to https://github.com/opentripplanner/otp-react-redux/pull/1272#discussion_r1779001128 , this PR makes the `setLocation` prop in the `EndpointsOverlay` optional (puts in the default `noop` handler as with other event handlers).

### DIFF
--- a/packages/endpoints-overlay/src/EndpointsOverlay.story.tsx
+++ b/packages/endpoints-overlay/src/EndpointsOverlay.story.tsx
@@ -46,11 +46,7 @@ export default {
 } as ComponentMeta<typeof EndpointsOverlay>;
 
 export const EndpointsOverlayWithoutUserSettings: ComponentStory<typeof EndpointsOverlay> = () => (
-  <EndpointsOverlay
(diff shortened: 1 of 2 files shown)


daniel-heppner-ibigroup (approved):
I feel less bad about requesting this change now that I see we already have `noop` as a default in here.

binh-dam-ibigroup:
> I feel less bad about requesting this change now that I see we already have `noop` as a default in here for other things

Maybe we'll have to remove the noops from here eventually?

daniel-heppner-ibigroup:
> > I feel less bad about requesting this change now that I see we already have `noop` as a default in here for other things
> 
> Maybe we'll have to remove the noops from here eventually?

Yeah It's a really wacky approach...

miles-grant-ibigroup (approved):
I think a noop function is much cleaner than doing weird existence checks elsewhere! It's also computationally more efficient (fewer jumps)

PR #789, opened 2024-10-11
fix: add combined stops/stations layer

miles-grant-ibigroup:
Requires a few hacks to get click handling working, but allows us to render stations/stops seamless as one layer

### DIFF
--- a/packages/otp2-tile-overlay/src/index.tsx
+++ b/packages/otp2-tile-overlay/src/index.tsx
@@ -16,6 +16,7 @@ import { generateLayerPaint, ROUTE_COLOR_EXPRESSION } from "./util"
 
 const SOURCE_ID = "otp2-tiles"
 const AREA_TYPES = ["areaStops"]
+const STOPS_AND_STATIONS_TYPE = "OTP-UI-stopsAndStations"
 
 const OTP2TileLayerWithPopup = ({
   color,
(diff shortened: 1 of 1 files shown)


packages/otp2-tile-overlay/src/index.tsx
```diff
@@ -224,14 +254,19 @@ const OTP2TileLayerWithPopup = ({
...
-  layers: { color?: string; name?: string; network?: string; type: string, initiallyVisible?: boolean }[],
+  layers: { color?: string; name?: string; network?: string; type: string, initiallyVisible?: boolean, overrideType?: string }[],
   endpoint: string,
   setLocation?: (location: MapLocationActionArg) => void,
   setViewedStop?: (stop: Stop) => void,
   stopsWhitelist?: string[],
   configCompanies?: ConfiguredCompany[],
   getEntityPrefix?: (entity: Stop | Station) => JSX.Element
 ): JSX.Element[] => {
+  const fakeOtpUiLayerIndex = layers.findIndex(l=>l.type === STOPS_AND_STATIONS_TYPE)
```
amy-corson-ibigroup, line 265: Can you add a comment here explaining why we're calling this `fakeOtpUiLayerIndex`

amy-corson-ibigroup (changes requested):
Only issue I'm seeing is this is not playing well with the entity prefix. Are you anticipating fixing that in OTP-RR or is there a way to not pass it to stations?
image

packages/otp2-tile-overlay/src/index.tsx
miles-grant-ibigroup: It's not a real layer! The others are all linked to actual maplibre layers

miles-grant-ibigroup:
Bug you mention is fixed here! https://github.com/opentripplanner/otp-react-redux/pull/1284

amy-corson-ibigroup (approved):
OTP-RR draft looks great thanks for the explanation!

josh-willis-arcadis (approved):
looks great!

PR #790, opened 2024-10-15
fix(itinerary-body): indicate rented bikes in class name

miles-grant-ibigroup:
Adds some class names that are helpful when styling

### DIFF
--- a/packages/itinerary-body/src/ItineraryBody/place-row.tsx
+++ b/packages/itinerary-body/src/ItineraryBody/place-row.tsx
@@ -78,7 +78,9 @@ export default function PlaceRow({
 
   return (
     <S.PlaceRowWrapper
-      className={`place-row-wrapper ${leg.transitLeg ? "transit" : ""}`}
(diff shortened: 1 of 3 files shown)


daniel-heppner-ibigroup (approved):
These interpolated classnames kind of bug me. We end up with random spaces in the result too. I'll approve it to not hold you up but what do you think about adding a cn function to our utilities like this? 
https://github.com/lukeed/clsx/blob/master/src/lite.js

miles-grant-ibigroup:
> These interpolated classnames kind of bug me. We end up with random spaces in the result too. I'll approve it to not hold you up but what do you think about adding a cn function to our utilities like this?
> https://github.com/lukeed/clsx/blob/master/src/lite.js

I agree. Cleaner approach here would be to get the mode from OTP and then use that as the classname. Unfortunately there's a lot of ancient code that requires the class names as currently exist

josh-willis-arcadis (approved):

PR #791, opened 2024-10-22
fix(trip-form): close mode button before setting submode options to default

josh-willis-arcadis:
Uses `setTimeout` to wait for the mode button animation to complete before setting submode options to their default values.


amy-corson-ibigroup (changes requested):
Code looks good, and it looks good in OTP-RR, but for some reason it's looking pretty weird on storybook? Is there a way we could fix the animation?

amy-corson-ibigroup (approved):
Beautiful!

daniel-heppner-ibigroup (approved):
Looks good! Not the worst hack I've seen

PR #793, opened 2024-11-04
Update Date/Time Selector

amy-corson-ibigroup:
Updates the D/T selector styling so that date time inputs are always visible, and the departure option (leave now, depart at, arrive by) is a dropdown.

If a user updates the date or time input when the option "leave now" is selected, the dropdown will change to "depart at".

| Before | After |
|--------|-------|
|image|image|

### DIFF
--- a/packages/trip-form/src/DateTimeSelector/index.tsx
+++ b/packages/trip-form/src/DateTimeSelector/index.tsx
@@ -1,24 +1,15 @@
 import CSS from "csstype";
 import { format, parse } from "date-fns";
-import flatten from "flat";
 import coreUtils from "@opentripplanner/core-utils";
-import React, { ChangeEvent, ReactElement, ReactNode, useCallback } from "react";
-import { FormattedMessage, useIntl } from "react-intl";
+import React, { ChangeEvent, ReactElement, useCallback } from "react";
+import { useIntl } from "react-intl";
 
-import ModeButton from "../ModeButton";
+import colors, { Dropdown } from "@opentripplanner/building-blocks";
 import * as S from "../styled";
 
 // eslint-disable-next-line prettier/prettier
 import type { QueryParamChangeEvent } from "../types";
 
-// Load the default messages.
-import defaultEnglishMessages from "../../i18n/en-US.yml";
-
-// HACK: We should flatten the messages loaded above because
-// the YAML loaders behave differently between webpack and our version of jest:
-// - the yaml loader for webpack returns a nested object,
-// - the yaml loader for jest returns messages with flattened ids.
-const defaultMessages: Record<string, string> = flatten(defaultEnglishMessages);
 
 const {
   getCurrentDate,
@@ -27,6 +18,8 @@ const {
   OTP_API_DATE_FORMAT,
   OTP_API_TIME_FORMAT
 } = coreUtils.time;
+
+const { blue } = colors;
 
 type DepartArriveValue = "NOW" | "DEPART" | "ARRIVE";
 
@@ -78,7 +71,7 @@ interface DateTimeSelectorProps {
 
 interface DepartArriveOption {
   isSelected?: boolean;
-  text: ReactNode;
+  text: string;
   type: DepartArriveValue;
 }
 
@@ -137,6 +130,26 @@ export default function DateTimeSelector({
   timeZone = getUserTimezone()
 }: DateTimeSelectorProps): ReactElement {
   const intl = useIntl()
+  const baseColor = S.baseColor()
+
+  const departureOptions: DepartArriveOption[] = [
+    {
+      // Default option.
+      type: "NOW",
+      text: intl.formatMessage({ id: "otpUi.DateTimeSelector.now" })
+    },
+    {
+      type: "DEPART",
+      text: intl.formatMessage({ id: "otpUi.DateTimeSelector.depart" })
+    },
+    {
+      type: "ARRIVE",
+      text: intl.formatMessage({ id: "otpUi.DateTimeSelector.arrive" })
+    }
+  ];
+  departureOptions.forEach(opt => {
+    opt.isSelected = departArrive === opt.type;
+  });
 
   const handleQueryParamChange = useCallback(
     (queryParam: QueryParamChangeEvent): void => {
@@ -150,8 +163,12 @@ export default function DateTimeSelector({
   const handleInputChange = (key: string) => useCallback(
     (evt: ChangeEvent<HTMLInputElement>): void => {
       handleQueryParamChange({ [key]: evt.target.value });
+      // If the user changes the time, it doesn't make sense for them to be departing now.
+      if (departArrive === "NOW") {
+        handleQueryParamChange({ departArrive: "DEPART" });
+      }
     },
-    [onQueryParamChange, key]
+    [onQueryParamChange, key, departArrive]
   );
 
   const handleDateChange = handleInputChange("date");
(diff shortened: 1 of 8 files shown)


miles-grant-ibigroup (changes requested):
We definitely need this to be a breaking change! Please add a breaking change commit. Stories look great! Haven't looked at the code yet

packages/trip-form/src/DateTimeSelector/index.tsx
```diff
@@ -236,21 +218,19 @@ export default function DateTimeSelector({
       className={className}
       role="group"
       style={style}
+      baseColor={baseColor}
     >
-      <S.DateTimeSelector.DepartureRow>
+        <Dropdown alignMenuLeft id="date-time-depart-arrive" text={departureOptions.find(opt => opt.isSelected).text} buttonStyle={{ backgroundColor: S.baseColor() || blue[900], borderRadius: "3px 0px 0px 3px", color: "white", height: "45px", border: "0px", padding:"5px 7px" }}>
```
miles-grant-ibigroup, line 223: Can we make the default gray instead of blue?

miles-grant-ibigroup: Also some of these values are pretty fixed and are causing layout problems in some implementations. Any way we can make this more reactive to its environment?

packages/trip-form/src/DateTimeSelector/index.tsx
```diff
@@ -137,6 +130,26 @@ export default function DateTimeSelector({
...
+      text: intl.formatMessage({ id: "otpUi.DateTimeSelector.depart" })
+    },
+    {
+      type: "ARRIVE",
+      text: intl.formatMessage({ id: "otpUi.DateTimeSelector.arrive" })
+    }
+  ];
+  departureOptions.forEach(opt => {
+    opt.isSelected = departArrive === opt.type;
+  });
```
miles-grant-ibigroup, line 152: This works but it's pretty inefficient... Anyway we can replace/generate `isSelected` when we're iterating through the array anyway?

miles-grant-ibigroup (changes requested):
First round of fixes! There's some layout problems in some contexts and some code cleanup possible too

packages/trip-form/src/DateTimeSelector/index.tsx
amy-corson-ibigroup: We approved the designs with the blue, also I think generally we've been using blue[900] to represent the styled base color, so using grey seems a little inconsistent.

Can you elaborate on which values are causing issues? I'm seeing the height but a lot of form elements have fixed heights. Took out the padding though because it was redundant!!

packages/trip-form/src/DateTimeSelector/index.tsx
amy-corson-ibigroup: Added! We do have to add the check in a few places, let me know what you think!

packages/trip-form/src/DateTimeSelector/index.tsx
miles-grant-ibigroup: Try with any non-blue config! The blue looks very out of place. As for getting render issues try loading a config that was migrated from the batch ui and then resize the window

packages/trip-form/src/DateTimeSelector/index.tsx
```diff
@@ -236,21 +214,19 @@ export default function DateTimeSelector({
       className={className}
       role="group"
       style={style}
+      baseColor={baseColor}
     >
-      <S.DateTimeSelector.DepartureRow>
+        <Dropdown alignMenuLeft id="date-time-depart-arrive" text={departureOptions.find(opt => opt.type === departArrive).text} buttonStyle={{ backgroundColor: S.baseColor() || blue[900], borderRadius: "3px 0px 0px 3px", color: "white", height: "45px", border: "0px" }}>
```
miles-grant-ibigroup, line 219: We can probably set extract `S.baseColor() || blue/gray`?

miles-grant-ibigroup (changes requested):
Mostly looking good, just the blue is still bothering me. Still have to do some more integration tests as well

packages/trip-form/src/DateTimeSelector/index.tsx
```diff
@@ -150,8 +159,12 @@ export default function DateTimeSelector({
   const handleInputChange = (key: string) => useCallback(
     (evt: ChangeEvent<HTMLInputElement>): void => {
       handleQueryParamChange({ [key]: evt.target.value });
+      // If the user changes the time, it doesn't make sense for them to be departing now.
+      if (departArrive === "NOW") {
```
miles-grant-ibigroup, line 163: nitpick: if departureOptions were an enum then we could reuse that instead of using a magic string

miles-grant-ibigroup (approved):
Code looks good! Haven't been able to test otp-rr integration yet let's hope that goes well

packages/trip-form/src/DateTimeSelector/index.tsx
amy-corson-ibigroup: Would you change `departureOptions` to 
```
departureOptions = {
now: "NOW"
etc
```
And then render the text strings dynamically?

miles-grant-ibigroup: yes! but it's such a nitpick

packages/trip-form/src/types.ts
```diff
@@ -107,11 +107,14 @@ export interface ModeSelectorOptionSet {
   tertiary?: ModeSelectorOption[];
 }
 
-type SimpleStyledDiv = StyledComponent<"div", any>;
+type SimpleStyledDiv = StyledComponent<
+  "div",
+  any,
+  { baseColor?: string | undefined }
```
binh-dam-ibigroup, line 113: Remove `undefined` (`undefined` is implied in the `?:` syntax).

```suggestion
  { baseColor?: string }
```

packages/trip-form/src/styled.ts
```diff
@@ -21,29 +28,26 @@ export const FloatingSettingLabel = styled(SettingLabel)`
...
-export const DateTimeSelector: DateTimeSelectorAndSubComponents = styled.div``;
-
-DateTimeSelector.DepartureRow = styled.div`
-  box-sizing: border-box;
-  > * {
-    box-sizing: border-box;
-    width: 33.333333%;
-    padding: 0px 5px;
-  }
+export const DateTimeSelector: DateTimeSelectorAndSubComponents = styled.div<{ baseColor: string | undefined }>`
```
binh-dam-ibigroup, line 31:
```suggestion
export const DateTimeSelector: DateTimeSelectorAndSubComponents = styled.div<{ baseColor?: string }>`
```

binh-dam-ibigroup (approved):
A few CSS tweaks, but seems fine otherwise.

binh-dam-ibigroup (changes requested):
Actually, you need to update the snapshots.

packages/trip-form/src/DateTimeSelector/index.tsx
```diff
@@ -216,7 +218,7 @@ export default function DateTimeSelector({
       style={style}
       baseColor={baseColor}
     >
-        <Dropdown alignMenuLeft id="date-time-depart-arrive" text={departureOptions.find(opt => opt.type === departArrive).text} buttonStyle={{ backgroundColor: S.baseColor() || grey[700], borderRadius: "3px 0px 0px 3px", color: "white", height: "45px", border: "0px" }}>
+        <Dropdown alignMenuLeft id="date-time-depart-arrive" text={departureOptions.find(opt => opt.type === departArrive).text} buttonStyle={buttonStyle}>
```
binh-dam-ibigroup, line 221: Breakup this long line.

binh-dam-ibigroup (approved):
Two small styling things but good to go nonetheless

amy-corson-ibigroup:
This is being closed in favor of a different design approach

PR #794, opened 2024-11-08
fix(transit-vehicle-overlay): use correct+consistent time units

miles-grant-ibigroup:
Maintains great OTP1 support


amy-corson-ibigroup (approved):
Another battle won in the war against unix timestamps

daniel-heppner-ibigroup (approved):
I don't want to know why

PR #797, opened 2024-11-20
fix(location-field): remove usecallbacks

amy-corson-ibigroup:
My bad, these callbacks are breaking Location Field sorry


josh-willis-arcadis (approved):

miles-grant-ibigroup (approved):
I've always been a callback hater

PR #798, opened 2024-11-26
route viewer overlay: avoid MultiPolygon crash

miles-grant-ibigroup:
Previously the multi polygon data structure broke the bounds checking. Since we don't support rendering that anyway, this PR catches this case and avoids a crash. The whole thing needs refactoring anyway. GeoJSON was built for functional programming and that's what should be done here.

### DIFF
--- a/packages/route-viewer-overlay/src/index.tsx
+++ b/packages/route-viewer-overlay/src/index.tsx
@@ -143,7 +143,7 @@ const RouteViewerOverlay = (props: Props): JSX.Element => {
               ? bounds.extend(coordinates)
               : new LngLatBounds(coordinates, coordinates);
           });
-        } else if (geoJson.type === "MultiPolygon") {
+        } else if (geoJson?.type === "MultiPolygon") {
           // We can't currently render this, so let's fix that before fixing the bounds behavior
           // TODO: handle this case
         } else if (geoJson) {


daniel-heppner-ibigroup:
typescript is failing

packages/route-viewer-overlay/src/index.tsx
```diff
@@ -127,9 +127,34 @@ const RouteViewerOverlay = (props: Props): JSX.Element => {
             reduceBounds,
             bounds || new LngLatBounds(coordsArray[0], coordsArray[0])
           );
+        } else if (geoJson?.type === "GeometryCollection") {
+          const { geometries } = geoJson;
```
daniel-heppner-ibigroup, line 131: I don't think this is correct. Should it be `stop.geometries.geometries`? `geoJson` is coming from `stop.geometries` above

daniel-heppner-ibigroup (changes requested):
uh

packages/route-viewer-overlay/src/index.tsx
miles-grant-ibigroup: Fixed it!

daniel-heppner-ibigroup: I think we can do better than this! If this isn't always going to be a Polygon then why not update the type in the types package? This type is coming from the Stop type on line 502 of types.

daniel-heppner-ibigroup (changes requested):

miles-grant-ibigroup:
Typescript is really broken here I think. I can type line 131 as GeoJSON.Geometry but TS still thinks there's no `coordinates` as children. What did we do here?

daniel-heppner-ibigroup (approved):
Raising the white flag to typescript here, do whatever you need to do, I can't unravel it.

amy-corson-ibigroup (changes requested):
Looks like this breaks a few of the stories, namely the default, path styling, and changing path

miles-grant-ibigroup:
Good catch thanks!

amy-corson-ibigroup (approved):

PR #799, opened 2024-12-04
feat(itinerary-body): add and use showAlertEffectiveDateTimeText to alert components

josh-willis-arcadis:
Make the 'Effective as of [date]' line from the alert message configurable, as requested from the client.

### DIFF
--- a/packages/itinerary-body/src/TransitLegBody/alerts-body.tsx
+++ b/packages/itinerary-body/src/TransitLegBody/alerts-body.tsx
@@ -107,9 +107,9 @@ export default function AlertsBody({
                   <S.TransitAlertHeader>{header}</S.TransitAlertHeader>
                 )}
                 <S.TransitAlertBody>{description}</S.TransitAlertBody>
-                <S.TransitAlertEffectiveDate>
-                  {showAlertEffectiveDateTimeText &&
-                    (Math.abs(dayDiff) <= 1 ? (
+                {showAlertEffectiveDateTimeText && (
+                  <S.TransitAlertEffectiveDate>
+                    {Math.abs(dayDiff) <= 1 ? (
                       <FormattedMessage
                         defaultMessage={
                           defaultMessages[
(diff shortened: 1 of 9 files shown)


packages/itinerary-body/src/TransitLegBody/alerts-body.tsx
```diff
@@ -106,34 +108,35 @@ export default function AlertsBody({
...
-                      }
-                      description="Text with the date an alert takes effect"
-                      id="otpUi.TransitLegBody.AlertsBody.effectiveDate"
-                      values={{
-                        dateTime: effectiveStartDate * 1000
-                      }}
-                    />
-                  )}
+                  {showAlertEffectiveDateTimeText &&
+                    (Math.abs(dayDiff) <= 1 ? (
```
daniel-heppner-ibigroup, line 112: What happens if you wrap the parent component `S.TransitAlertEffectiveDate` with the new boolean? I think this would make it a bit clearer just because it would break up this bit conditional block here that can be a bit hard to follow. At first I thought it was `showAlertEffectiveDateTimeText && Math.abs(dayDiff) <= 1` because I didn't see the `(` so I think breaking it up would be good just for clarity.

packages/itinerary-body/src/types.ts
```diff
@@ -201,6 +201,8 @@ interface ItineraryBodySharedProps {
   setViewedTrip: SetViewedTripFunction;
   /** If true, will show agency information in transit legs */
   showAgencyInfo?: boolean;
+  /** If true, will show the effect date/time text in the alert body */
```
daniel-heppner-ibigroup, line 204: Since it's default to true, maybe change the comment to "if false..."

daniel-heppner-ibigroup (approved):
LGTM!

packages/itinerary-body/src/types.ts
josh-willis-arcadis: I agree. Good catch!

packages/itinerary-body/src/TransitLegBody/alerts-body.tsx
josh-willis-arcadis: addressed in dc9fb59

amy-corson-ibigroup (approved):
LGTM! Thanks for including a story!!

PR #800, opened 2024-12-05
Support Alternate Geocoder Methods for Auto-Search

miles-grant-ibigroup:
Using a new field in the geocoderConfig is a bit unclean... Curious if someone has a better idea that doesn't involve drilling.


daniel-heppner-ibigroup (approved):
This looks fine to me but I'm curious what the problem is with adding it to the config? Doesn't that seem like a reasonable place for it?

PR #802, opened 2024-12-12
fix(itinerary-body): Optimize icon import in AlertsBody.

binh-dam-ibigroup:
This PR imports a single icon from fa-solid instead of "requiring" the entire library. This change can help reduce the main OTP-RR production bundle size by as much as 2+MB! (See screenshot before/after)

image


daniel-heppner-ibigroup (approved):
Nice catch. I think these imports can sneak in easily through an auto import. Probably a smarter bundle tool would catch this automatically.

PR #803, opened 2024-12-19
Add 'boundsFitting' prop to manage camera manipulation via zoomToGeoJSON method

lightsage88:
This pull request addresses a need for Trimet's instance of the maplibre map to not get the camera's placement hijacked by the _zoomToGeoJSON_ method which is causing some issues with how we want to frame the transitive-overlay with respect to the side-panel we house much of our tools in.

Essentially, we pass in _boundsFitting_ from the client and into packages/transitive-overlay/src/index.tsx. If it happens to be **false** then when _zoomToGeoJSON_ is called then it returns immediately with no effect on the map. Since this was set up to be active by default, I made the _boundsFitting_ prop be defaultly **true**.

### DIFF
--- a/package.json
+++ b/package.json
@@ -284,5 +284,6 @@
   },
   "dependencies": {
     "@storybook/icons": "^1.2.9"
-  }
+  },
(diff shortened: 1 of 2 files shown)


binh-dam-ibigroup (approved):

miles-grant-ibigroup:
Blocked by https://github.com/opentripplanner/otp-ui/pull/804. I'll review once that's merged. I have a feeling snapshots might have to get updated

PR #809, opened 2025-01-28
feat(package.json): move maplibre-gl to peer dependencies

miles-grant-ibigroup:
BREAKING CHANGE: maplibre-gl is no longer a direct dependency

@c-harper

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -11,6 +11,8 @@ jobs:
           node-version: 21.x
       - name: Install npm packages using cache
         uses: bahmutov/npm-install@v1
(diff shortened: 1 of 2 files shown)


miles-grant-ibigroup:
Looks like this may only be possible with a move to pnpm: https://github.com/opentripplanner/otp-ui/pull/795

In that case we definitely need the yarn warning/alias. Since how else will people know to use pnpm instead of yarn

miles-grant-ibigroup:
@daniel-heppner-ibigroup

miles-grant-ibigroup:
Let's go ahead and close in favor of #795 sorry about the chaos all!

PR #810, opened 2025-02-05
fix(itinerary-body): show alert links if showAlertEffectiveDateTimeText is false

josh-willis-arcadis:
When `showAlertEffectiveDateTimeText` is false, still show alert urls.


daniel-heppner-ibigroup (approved):
LGTM nice catch

PR #813, opened 2025-03-03
feat(core-utils): add walkTolerance param and restore walkSpeed param for walk-only trips

themapwizard:
This PR adds walkTolerance and walkSpeed to core-utils\src\query-params.js and adds them to the default params in core-utils\src\query.js.

This is for trimet.org and will add a walk tolerance dropdown to the TripOptions UI for transit+walk modes and adds back walkSpeed for walk only modes. We needed to add back walkSpeed because latest versions 12.0.2 do not contain walkSpeed and upgrading in trimet.org would render no settings in the walk only panel (see screenshots). 

walk + transit:
Screenshot 2025-03-03 115112

walk only:
Screenshot 2025-03-03 115125


packages/core-utils/src/query-params.js
```diff
@@ -163,6 +163,65 @@ const queryParams = [
     default: true
   },
 
+  {
+    /* walkTolerance -- simplified UI for walkReluctance */
+    name: "walkTolerance",
+    routingTypes: ["ITINERARY"],
+    applicable: query =>
+      !query.otp2 &&
```
miles-grant-ibigroup, line 171: Are you sure you want this to only trigger when `otp2` is not present?

packages/core-utils/src/query-params.js
```diff
@@ -163,6 +163,65 @@ const queryParams = [
...
+        text: "Normal walking",
+        value: 2
+      },
+      {
+        text: "Less walking",
+        value: 4
+      }
+    ],
+    itineraryRewrite: value => ({
+      walkReluctance: value
```
miles-grant-ibigroup, line 193: If we're not using `walkTolerance` as a separate value, why not just use `walkReluctance` and re-name the label?

miles-grant-ibigroup (changes requested):
Looking at the failing snapshots, it looks like these 2 variables are being applied in many cases where they perhaps should not be. This PR would send a default 2 walkReluctance along in many cases where it perhaps should not be. If this is desired, then I'd consider this a change that might make more sense within the trimet-specific map form component.

packages/core-utils/src/query-params.js
themapwizard: I wasn't sure if the existing walkReluctance slider was being used anywhere.

packages/core-utils/src/query-params.js
themapwizard: You're right, we don't need it. I've removed the `!query.otp2` condition

packages/core-utils/src/query-params.js
themapwizard: I've removed walkTolerance, we are going to modify our configs and just change the label. We still need this PR to add back walkSpeed for the walk only mode.

- https://maps.trimet.org/otp2-test/config.yaml
- src/otp-ui-config.js

themapwizard:
> Looking at the failing snapshots, it looks like these 2 variables are being applied in many cases where they perhaps should not be. This PR would send a default 2 walkReluctance along in many cases where it perhaps should not be. If this is desired, then I'd consider this a change that might make more sense within the trimet-specific map form component.

this should no longer be an issue, i've update the PR to only add back walkSpeed

miles-grant-ibigroup (approved):
If you go ahead and update the snapshots then I think this is good to go!

themapwizard:
> If you go ahead and update the snapshots then I think this is good to go!

updated

miles-grant-ibigroup:
It doesn't look like that quite worked! Did you try running `yarn unit -u`?

themapwizard:
> It doesn't look like that quite worked! Did you try running `yarn unit -u`?

I originally ran it with jest directly, i just re-ran with `yarn unit -u`

daniel-heppner-ibigroup:
Sorry @themapwizard `yarn unit -u` is actually outdated, we now have a script for updating the storybook snapshots because they require that the storybook be built to HTML. You can run `yarn update-snapshots`

daniel-heppner-ibigroup:
I was able to do it myself but I can't push to your branch so you'll have to do it yourself. Sorry for the runaround.

PR #814, opened 2025-03-07
Fix route sorting comparator

amy-corson-ibigroup:
OTP1 returns a route `sortOrder` value of `-999` if it is not defined by the agency. But OTP2 returns it as `null`. All our `sortOrder` route sorting was assuming we'd see default values of `-999` and as a result it does not work for OTP2. The route sort order sorting has been fixed to work for both otp1 and otp2. PLEASE NOTE: This means that in OTP1 environments `-999` is an _invalid_ `sortOrder` value and is treated as `null`, and in OTP2 environments `-999` is a _valid_ sortOrder value and will not be overwritten (and therefore probably will be at the top of the list). The tests have been rewritten to reflect this.

This was also really annoying to debug so I made a route sort comparator story with checkboxes so users can see how each sort comparator is impacting the results.

The operator and route data comes from Sound Transit deployment, however I have added operator orders to the data to better show how those impact the sort.

image


packages/core-utils/src/route.ts
```diff
@@ -282,6 +284,13 @@ function alphabeticShortNameComparator(a: Route, b: Route): number {
   return 0;
 }
 
+const undefinedNullOrNaN = (val: any): boolean => {
+  /* Note: Using the global version of isNaN (the Number version behaves differently. */
+  // eslint-disable-next-line no-restricted-globals
+  if (typeof val === "number" && isNaN(val)) return true;
+
+  return val === null || val === undefined;
+};
```
miles-grant-ibigroup, line 293: I don't think this function is needed. `isNaN(undefined)` is true

packages/core-utils/src/route.ts
```diff
@@ -304,19 +313,21 @@ function alphabeticShortNameComparator(a: Route, b: Route): number {
 export function makeNumericValueComparator(
   objGetterFn?: (item: Route) => number
 ) {
-  /* Note: Using the global version of isNaN (the Number version behaves differently. */
-  /* eslint-disable no-restricted-globals */
   return (a: number, b: number): number => {
     const { aVal, bVal } = getSortValues(objGetterFn, a, b);
-    if (typeof aVal !== "number" || typeof bVal !== "number") return 0;
```
miles-grant-ibigroup, line 311: Instead of removing this line, should we just set the value to not be null?

packages/core-utils/src/route.ts
```diff
@@ -350,14 +361,15 @@ export function makeStringValueComparator(
...
  * is encountered, assume that it actually means that the routeSortOrder is not
  * set in the GTFS.
  *
  * See https://github.com/opentripplanner/OpenTripPlanner/issues/2938
  * Also see https://github.com/opentripplanner/otp-react-redux/issues/122
+ * This was updated in OTP2 TO be empty by default. https://docs.opentripplanner.org/en/v2.3.0/OTP2-MigrationGuide/#:~:text=the%20Alerts-,Changes%20to%20the%20Index%20API,-Error%20handling%20is
  */
-function getRouteSortOrderValue(val: number): number {
+export function getOTP1RouteSortOrderValue(val: number): number {
   return val === -999 ? undefined : val;
```
miles-grant-ibigroup, line 373: Instead of the NaN horrors above should we just add a check in here that checks for `val === null`?

packages/core-utils/src/route.ts
```diff
@@ -420,7 +432,9 @@ export function makeRouteComparator(
 ): (a: number, b: number) => number {
   return makeMultiCriteriaSort(
     makeTransitOperatorComparator(transitOperators),
-    makeNumericValueComparator(obj => getRouteSortOrderValue(obj.sortOrder)),
+    makeNumericValueComparator(obj =>
+      getOTP1RouteSortOrderValue(obj.sortOrder)
```
miles-grant-ibigroup, line 436: I like that we are clarifying about OTP1, but could we do it in comments instead of the method name?

packages/core-utils/src/fake-route-data.json
```diff
@@ -0,0 +1,5425 @@
+{
```
miles-grant-ibigroup, line 1: Can we put these files in a test folder? that way they don't get bundled into a release

miles-grant-ibigroup (changes requested):
A few things to consider. My biggest worry is the failing unit tests, but I think if we avoid changes to our `isNaN` and instead adjust our null handling earlier in the process we should get the result we want without breaking previous tests. 

I'd also say we should exempt the new story from the snapshots. I think it causes more problems than it solves

packages/core-utils/src/route.ts
amy-corson-ibigroup: I think having `null` as a default is best because in OTP2 any numeric value should be a valid sort order, even -999. An agency should be able to use -999 with OTP2 and push routes to the top of the list if they want.

packages/core-utils/src/route.ts
amy-corson-ibigroup: I added a n`ull` check along with the `NaN` check here just to be safe? But I also think I updated the other function to filter out cases of `undefined` so maybe it's overkill, lmk

packages/core-utils/src/fake-route-data.json
amy-corson-ibigroup: I added a `.story` suffix to them which I think should prevent them from being bundled. An alternative idea is using the fake routes in the __mocks__ folder instead of creating this new set of fake data? Lmk what you think.

amy-corson-ibigroup:
This PR has been updated to reflect the idea that -999 should be valid in OTP2 environments and invalid in OTP1 environments. I have written new tests and updated snapshots. Also, some of the existing tests were broken so I updated those too. More details are in the PR description.

@miles-grant-ibigroup I'd really prefer to keep the story in. This sorting logic is complicated. It took me hours to parse it out, and I work on this project full time. If we want these `core-utils` to be more accessible to the OTP community, this feels like a good step, plus we can use it to debug if OTP changes in the future. But to be fair, I don't know what problems you're referring to, so feel free to elaborate!

packages/core-utils/src/route.ts
miles-grant-ibigroup: Makes sense!

miles-grant-ibigroup: That being said why are we using null instead of 0? Isn't 0 more precise?

packages/core-utils/src/fake-route-data.json
miles-grant-ibigroup: the `.story` suffix is great! I'm agreed we should keep the story I think it's good. But the mock data needs to go in a dedicated testing folder! IS it possible to use the itinerary-body package as a guide?

miles-grant-ibigroup:
Agreed the story is good. I've responded to your comments I think we're getting there

packages/core-utils/src/route.ts
amy-corson-ibigroup: 0 is not only a valid number, it's actively being used as a `sortOrder` value in some deployments!

packages/core-utils/src/fake-route-data.json
amy-corson-ibigroup: Added the fake data to a new mocks folder, that seemed to be how we handled it in itinerary body! Let me know if that's not what you were thinking

packages/core-utils/src/route.ts
```diff
@@ -350,15 +362,23 @@ export function makeStringValueComparator(
...
+ * it in OTP2, it's a valid value, so we should return it.
  *
  * See https://github.com/opentripplanner/OpenTripPlanner/issues/2938
  * Also see https://github.com/opentripplanner/otp-react-redux/issues/122
+ * This was updated in OTP2 TO be empty by default. https://docs.opentripplanner.org/en/v2.3.0/OTP2-MigrationGuide/#:~:text=the%20Alerts-,Changes%20to%20the%20Index%20API,-Error%20handling%20is
  */
-function getRouteSortOrderValue(val: number): number {
-  return val === -999 ? undefined : val;
+export function getRouteSortOrderValue(route: Route): number {
+  const isOTP1 = !!route.agencyId;
```
miles-grant-ibigroup, line 374: This line gives me pause because I know we are sometimes converting OTP2 data into OTP1 data but I think that's ok because in that case the behavior would be unchanged from the way it is now.

miles-grant-ibigroup (approved):
Tested and working! And nervous

packages/core-utils/src/core-utils.story.tsx
```diff
@@ -55,3 +66,192 @@ RouteColorTester.parameters = {
...
+    height: 30px;
+    width: 30px;
+  }
+`;
+
+const GridCell = styled.span`
+  border: 1px solid black;
+  padding: 2px;
+`;
+export const RouteSortingLogic = (): JSX.Element => {
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Good point, comment added in [0078fdd](https://github.com/opentripplanner/otp-ui/pull/814/commits/0078fdd6436b34a21310f2c267b0e32360769a60)

binh-dam-ibigroup (changes requested):
Convert the table of routes to an actual `<table>`, please, plus some other nits as indicated.

packages/core-utils/src/core-utils.story.tsx
```diff
@@ -55,3 +66,199 @@ RouteColorTester.parameters = {
...
+      <div
+        style={{
+          display: "flex",
+          flexDirection: "column",
+
+          width: "100%"
+        }}
+      >
+        <RouteRow>
+          <GridCell>Logo</GridCell>
```
binh-dam-ibigroup, line 162: This should be `<th>`.

packages/core-utils/src/core-utils.story.tsx
```diff
@@ -55,3 +66,199 @@ RouteColorTester.parameters = {
...
+// Route sort logic story:
+
+const Columns = styled.div`
+  position: relative;
+  display: flex;
+  flex-direction: row;
+  width: 100%;
+`;
+
+const RouteRow = styled.div`
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Cannot believe I forgot about `table`...

packages/core-utils/src/core-utils.story.tsx
```diff
@@ -138,54 +146,39 @@ export const RouteSortingLogic = (): JSX.Element => {
...
-      <div
-        style={{
-          display: "flex",
-          flexDirection: "column",
-
-          width: "100%"
-        }}
-      >
+      <StyledTable>
         <RouteRow>
```
binh-dam-ibigroup, line 156: Convert `<RouteRow>` to `<tr>` (move that CSS under `StyledTable`) (2 instances).

binh-dam-ibigroup (changes requested):
One more (minor) change, please, and update the snapshot.

binh-dam-ibigroup (approved):
Thanks for the changes!

PR #816, opened 2025-03-19
feat(itinerary-body): use routeDurationToMinute to remove 0min durations

amy-corson-ibigroup:
Previously, leg durations under 60 seconds would read as "0 mins" which is a little confusing. This introduces a small util function to round all access leg durations to at least one minute.


packages/itinerary-body/src/util.ts
```diff
@@ -24,6 +24,13 @@ export const toSafeRouteColor = (routeColor: string): string => {
   );
 };
 
+/**
+ * If a duration is less than 60 seconds, round it to one minute, to avoid a duration
+ * of 0 minutes on a leg.
+ */
+export const roundDurationToMinute = (duration: number): number =>
```
binh-dam-ibigroup, line 31: Are you planning to reuse this in OTP-RR? If so, should this function reside in core-utils/time? (Module `util` is currently not exported in the package's index file, so we won't be able to import that cleanly in OTP-RR, and that can result in duplicate code in the bundle)

binh-dam-ibigroup (commented):
@amy-corson-ibigroup  Are you planning to reuse the rounding function in OTP-RR?

packages/itinerary-body/src/util.ts
amy-corson-ibigroup: That's a good idea, moved in [2b509e1](https://github.com/opentripplanner/otp-ui/pull/816/commits/2b509e14cf3c3701739331819f4039ed07e37f66) and [29145fc](https://github.com/opentripplanner/otp-ui/pull/816/commits/29145fce36077c5f6ede6cf0eb33d5c3bdccb5f5)

binh-dam-ibigroup (approved):
Might need to update the required core-utils version in itinerary-body's package.json after the new core-utils version is available.

amy-corson-ibigroup:
I cherry-picked the core-utils commits to https://github.com/opentripplanner/otp-ui/pull/819 and assigned you both! Once that is merged I will update the version of core-utils here.

PR #817, opened 2025-03-21
fix(transitive): render route label background in webgl

miles-grant-ibigroup:
Our current route background hack was having some issues where sometimes just the front/back ground of the layer would render. This PR resolves this by rendering the background as part of the foreground layer

### TODO: 
- padding seems a bit high? It should render the labels at lower zoom levels
- configurable? Are we ok with all this as default behavior?


packages/transitive-overlay/src/images/pill.png
binh-dam-ibigroup, line 1: Is this image with a route number used for a demo or something?

miles-grant-ibigroup: SDF ignores all data other than the alpha channel

binh-dam-ibigroup: Okay, then could you put a similar image without numbers? (saves a few bytes)

packages/transitive-overlay/src/route-layers.ts
```diff
@@ -60,11 +50,33 @@ export function patternToRouteFeature(
...
+    "text-font": ["Open Sans Bold", "Arial Unicode MS Bold"],
     "text-field": ["get", textField],
     "text-ignore-placement": true,
+    "text-justify": "center",
+    "text-overlap": "always",
     "text-rotation-alignment": "viewport",
+    "icon-rotation-alignment": "viewport",
+    "text-padding": 0,
+    "text-allow-overlap": true,
+    "icon-text-fit-padding": [0, 0, 0, 0],
```
binh-dam-ibigroup, line 78: Sort attributes. Could you make this padding a prop?

packages/transitive-overlay/src/route-layers.ts
```diff
@@ -60,11 +50,33 @@ export function patternToRouteFeature(
...
+    // "icon-image": "debug",
+    "icon-image": [
+      "case",
+      ["==", ["length", ["get", textField]], 1],
+      "circle-1",
+      ["==", ["length", ["get", textField]], 2],
+      "circle-2",
+      ["==", ["length", ["get", textField]], 3],
+      "circle-3",
+      "rect"
```
binh-dam-ibigroup, line 62: I think the default behavior should be a rectangle so it is like the current behavior. If circles/pills are needed, how about adding a prop that takes a lookup, falling back on the rectangle if none:
```
// length-based lookup
{
  1: "circle-1",
  2: "circle-2",
  // etc
  // then convert into style-specification syntax
}
```

packages/transitive-overlay/src/index.tsx
```diff
@@ -129,9 +133,43 @@ const TransitiveCanvasOverlay = ({
...
+      url: routeArrow,
+      options: { sdf: true }
     });
   }
+  mapImages.push({
+    id: "circle-1",
+    url: circle,
+    options: {
+      sdf: true,
+      content: [440, 370, 560, 630]
```
binh-dam-ibigroup, line 145: Is this the content box? Add a comment to compute this. If padding goes in here, then a prop of type `number[4]` would probably do.

binh-dam-ibigroup (changes requested):
Nice attempt that also brings interesting visuals. Could you:
- Add props to conditionally set circles based on the content length
- Add a prop for padding if that makes any sense
- Add a `feat` commit as we are adding functionality.

binh-dam-ibigroup (changes requested):
The part the computes padding is very obscure, could you explain how the magic numbers came to be please?

packages/transitive-overlay/src/route-layers.ts
```diff
@@ -60,10 +50,58 @@ export function patternToRouteFeature(
...
+    "icon-optional": false,
+    // @ts-expect-error maplibre is not typed correctly
+    "icon-overlap": "always",
+    "icon-rotation-alignment": "viewport",
+    "icon-text-fit-padding": [0, 0, 0, 0],
+    "icon-text-fit": "both",
     "symbol-placement": "line-center",
     "text-allow-overlap": true,
     "text-field": ["get", textField],
+    "text-font": ["Open Sans Bold", "Arial Unicode MS Bold"],
```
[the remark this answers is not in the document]
miles-grant-ibigroup: It has to be fonts that are explicitly supported by MapLibre GL.

miles-grant-ibigroup:
I have a new approach for the padding generation that is much cleaner but that might have weird edge cases I haven't found yet. Curious for your experience

binh-dam-ibigroup (changes requested):
Could you use loops to load images and generate the `case` logic, and add a comment regarding the padding for the narrow shapes.

packages/transitive-overlay/src/index.tsx
```diff
@@ -129,10 +147,141 @@ const TransitiveCanvasOverlay = ({
...
+    }
+  });
+  mapImages.push({
+    id: "17",
+    url: seventeen,
+    options: {
+      sdf: true
+    }
+  });
+  mapImages.push({
```
binh-dam-ibigroup, line 277: Add a story to illustrate a very long route name.

packages/transitive-overlay/src/index.tsx
```diff
@@ -152,128 +186,66 @@ const TransitiveCanvasOverlay = ({
...
-    options: {
-      sdf: true
+
+    if (width > 12) {
+      topPad = 175;
     }
-  });
+
+    // Each image has height 533
+    return [topPad, 0, WIDTH_IMAGE_SIZES[width] - topPad, 533];
```
binh-dam-ibigroup, line 213: Extract a constant for the image height.

packages/transitive-overlay/src/index.tsx
```diff
@@ -152,128 +186,66 @@ const TransitiveCanvasOverlay = ({
...
     options: {
+      // These paddings are specifically set so that the circle appears circular
+      // despite non-circular padding
+      content: [0, 40, 500, 460],
       sdf: true
     }
   });
+
+  // Generate each capsule image from 3 - 17
+  for (let i = 3; i < 18; i++) {
```
binh-dam-ibigroup, line 238: Much better!

packages/transitive-overlay/src/route-layers.ts
```diff
@@ -49,47 +49,25 @@ export function patternToRouteFeature(
  * Obtains common layout options for route label layers.
  */
 export function getRouteLayerLayout(textField: string): SymbolLayout {
+  // Generates a single icon based on the string length
+  function generateIcon(length: number) {
+    return [["==", ["length", ["get", textField]], length], `${length}`];
+  }
+
+  // Generates every icon length from 1-17. Anything higher renders a rectangle
```
binh-dam-ibigroup, line 57: Much better!

packages/transitive-overlay/src/index.tsx
```diff
@@ -15,24 +15,58 @@ import bbox from "@turf/bbox";
...
-import nine from "./images/09.png";
-import ten from "./images/10.png";
-import eleven from "./images/11.png";
-import twelve from "./images/12.png";
-import thirteen from "./images/13.png";
-import fourteen from "./images/14.png";
-import fifteen from "./images/15.png";
-import sixteen from "./images/16.png";
-import seventeen from "./images/17.png";
+import capsule1 from "./images/01.png";
```
binh-dam-ibigroup, line 18: For subsequent PR: The images are being downloaded all, even though only one/a subset is actually used.

binh-dam-ibigroup (approved):
Much better with the changes that make the code more concise. The loading of all images can be dealt with in another PR.

daniel-heppner-ibigroup:
I fixed the type issues, but I had to change some of the return type in getRouteLayerLayout. I couldn't find documentation for two of the properties you had there so I removed them, because I don't think they did anything. I replaced the icon overlap one with one that looks like it does what you intended? Double check [aa930d2](https://github.com/opentripplanner/otp-ui/pull/817/commits/aa930d2bf1e5bd52544bdcd4b6a276b702734a2b).

daniel-heppner-ibigroup (approved):
Looks good! really nice work! Check out my commit that fixes the types, otherwise we can revert it.

PR #818, opened 2025-03-26
fix(otp geocoder): don’t send “stop id ” to otp

miles-grant-ibigroup:
This makes it easier to search for stop codes without getting street addresses mixed in

### DIFF
--- a/packages/geocoder/src/apis/otp/index.ts
+++ b/packages/geocoder/src/apis/otp/index.ts
@@ -59,8 +59,14 @@ async function autocomplete({
   url,
   text
 }: AutocompleteQuery): Promise<OTPGeocoderResponse> {
+  let shortenedText = text;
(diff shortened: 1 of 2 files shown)


binh-dam-ibigroup (approved):

daniel-heppner-ibigroup (changes requested):
Do we really want this to be hard coded in and not configurable? And not translated? Also why did yarn.lock change

miles-grant-ibigroup:
> Do we really want this to be hard coded in and not configurable? And not translated? Also why did yarn.lock change

Yep and yep! It's a silly feature so I'm trying to avoid spending too many lines of code on it.

Yarn.lock has changed because one of the non-version locked packages has changed I think? It's happening pretty consistently so it'll have to happen in some PR soon

daniel-heppner-ibigroup (approved):
okay, sounds good enough to me.

PR #820, opened 2025-03-28
feat(trip-details): remove time spent active

amy-corson-ibigroup:
Remove the time spent active line from the itinerary details


amy-corson-ibigroup:
Actually this might need to be a breaking change?

amy-corson-ibigroup:
Closing for now, might re-open pending more research

PR #821, opened 2025-03-28
Update "Recently Searched" to "Recent places" in location field

amy-corson-ibigroup:
Update the "recently searched" section to "recent places" so as not to be confused with the recent trip searches which are also called "recently searched"


amy-corson-ibigroup:
Closed in favor of https://github.com/opentripplanner/otp-ui/pull/823

PR #823, opened 2025-03-31
fix(location-field): Change recent searches to recent places

amy-corson-ibigroup:
Change recent searches to recent places. Seemed like a good idea to change the translation key to accurately reflect what it was representing.


miles-grant-ibigroup (approved):
We should figure out what our thinking/policy is on renaming i18n keys but this is fine for now.

amy-corson-ibigroup:
Good point about the breaking change, also discussing offline it sounds like we're not on the same page about whether this string change is actually desired? Closing in favor of a configurable solution!

PR #824, opened 2025-04-08
Fix compile issue with maplibre

daniel-heppner-ibigroup:
This should have been in the pnpm PR but I forgot to merge it over from the alpha branch after I fixed it there. This PR does that, it also releases every dependent package by modifying the readmes. Not elegant but at least we can release a bunch of packages at once now without having to stagger the PRs!


miles-grant-ibigroup (approved):
Great work

binh-dam-ibigroup (approved):

PR #825, opened 2025-04-10
OTP2 Tilelayer MinZoom Support 

miles-grant-ibigroup:
Allows implementers to define a minZoom override, to avoid being stuck with `14`.


daniel-heppner-ibigroup (approved):
LGTM, thanks for the pack all change too

daniel-heppner-ibigroup:
I think this should be a feat not fix btw

miles-grant-ibigroup:
> I think this should be a feat not fix btw

Why? It's not a breaking change. No behavior changes

amy-corson-ibigroup (approved):
love it

PR #826, opened 2025-04-17
fix(location-field): Fix keyboard navigation

amy-corson-ibigroup:
Splitting the geocoded features into categories broke the way we applied `itemIndex` to the list of menu items, which subsequently broke keyboard navigation. Instead of tracking the options by index, and also using a location function lookup, now we use `indexedOptionLookup` to create an array of all the ids and location handlers, and use that to iterate through the options in `onKeyDown`

Also added `generateOptionId` to create bespoke consistent ids for each element (open to feedback on this, it's pretty hacky)

### DIFF
--- a/packages/location-field/src/index.tsx
+++ b/packages/location-field/src/index.tsx
@@ -38,16 +38,6 @@ import defaultEnglishMessages from "../i18n/en-US.yml";
 
 type IndexedOptionLookup = Array<{ id: string; locationSelected: () => void }>;
 
-const optionIdPrefix = "otpui-locf-option";
-
-/**
- * Formats the option id based on its given index position.
- * This assumes only one location dropdown is shown at a time.
- */
-function getOptionId(index: number): string {
-  return `${optionIdPrefix}-${index}`;
-}
-
 function generateOptionId(optionPrefix, feature) {
   // Use a default id if the feature does not have an id.
   const featureId =
@@ -726,7 +716,7 @@ const LocationField = ({
         matchingLocations.map(userLocation =>
           makeUserOption(
             userLocation,
-            generateOptionId("userLocationResults", userLocation),
+            generateOptionId("user-saved-results", userLocation),
             activeIndex,
             pushToIndexedOptions,
             indexedOptionLookup
@@ -747,7 +737,7 @@ const LocationField = ({
     const geocodedFeaturesWithId = geocodedFeatures.map(feature => {
       return {
         ...feature,
-        id: generateOptionId("geocodedResults", feature),
+        id: generateOptionId("geocoder", feature),
         locationSelected: () => setLocationSelected(feature)
       };
     });
@@ -927,10 +917,7 @@ const LocationField = ({
           setLocation(sessionLocation, "SESSION");
         };
 
-        const locationId = generateOptionId(
-          "recentSearchResults",
-          sessionLocation
-        );
+        const locationId = generateOptionId("recent", sessionLocation);
 
         // Add to the selection handler lookup (for use in onKeyDown)
         pushToIndexedOptions(locationSelected, locationId);
@@ -970,7 +957,7 @@ const LocationField = ({
       userLocationRenderData.map(userLocation =>
         makeUserOption(
           userLocation,
-          generateOptionId("userLocations", userLocation),
+          generateOptionId("user-saved", userLocation),
           activeIndex,
           pushToIndexedOptions,
           indexedOptionLookup
@@ -1061,7 +1048,7 @@ const LocationField = ({
   const textControl = (
     <S.Input
       aria-activedescendant={
-        activeIndex !== null ? getOptionId(activeIndex) : null
+        activeIndex !== null ? indexedOptionLookup[activeIndex]?.id : null
       }
       aria-autocomplete="list"
       aria-controls={listBoxId}
--- a/packages/location-field/src/index.tsx
+++ b/packages/location-field/src/index.tsx
@@ -46,8 +46,27 @@ function getOptionId(index: number): string {
   return `${optionIdPrefix}-${index}`;
 }
 
-// FIXME have a better key generator for options
-let optionKey = 0;
+function generateOptionId(feature, featureItems) {
+  // Use a default id if the feature does not have an id.
+  let id = "000";
+
+  // User saved locations do not have an id, so use the label.
+  if (feature.displayName) {
+    id = feature.displayName.replace(/\s/g, "");
+  }
+
+  // Some geocoders will return an id. Use that if it exists.
+  if (feature.properties) {
+    id = feature.properties.id || feature.properties.label.replace(/\s/g, "");
+  }
+
+  // There are some cases, such as when we're using user-saved locations, where
+  // the id is not unique. In this case, append a number to the end of the id.
+  if (featureItems.find(item => item.id === id)) {
+    id = `${id}0`;
+  }
+  return id;
+}
 
 function DefaultLocationIcon({
   locationType
@@ -135,16 +154,22 @@ function getFeaturesByCategoryWithLimit(
 /**
  * Helper to render and register a user-saved location.
  */
-function makeUserOption(userLocation, index, key, activeIndex, selectHandlers) {
+function makeUserOption(
+  userLocation,
+  id,
+  activeIndex,
+  pushToFeaturedItems,
+  featureItems
+) {
   const { displayName, icon, locationSelected } = userLocation;
   // Add to the selection handler lookup (for use in onKeyDown)
-  selectHandlers[index] = locationSelected;
+  pushToFeaturedItems(locationSelected, id, displayName);
   return (
     <Option
       icon={icon}
-      id={getOptionId(index)}
-      isActive={index === activeIndex}
-      key={key}
+      id={id}
+      isActive={featureItems[activeIndex]?.id === id}
+      key={id}
       onClick={locationSelected}
       title={displayName}
     />
@@ -152,40 +177,18 @@ function makeUserOption(userLocation, index, key, activeIndex, selectHandlers) {
 }
 
 const renderFeature = (
-  itemIndex,
   layerColorMap,
   feature,
   operatorIconMap,
-  setLocation,
-  addLocationSearch,
   showSecondaryLabels,
-  locationSelectedLookup,
   activeIndex,
   GeocodedOptionIconComponent,
-  geocoderConfig
+  featureItems
 ) => {
   // generate the friendly labels for this feature
   const { main, secondary } = generateLabel(feature.properties);
 
   // Create the selection handler
-  const locationSelected = () => {
-    getGeocoder(geocoderConfig)
-      .getLocationFromGeocodedFeature(feature)
-      .then(geocodedLocation => {
-        // add the friendly location labels for use later on
-        geocodedLocation.main = main;
-        geocodedLocation.secondary = secondary;
-        geocodedLocation.name = getCombinedLabel(feature.properties);
-        // Set the current location
-        setLocation(geocodedLocation, "GEOCODE");
-        // Add to the location search history. This is intended to
-        // populate the sessionSearches array.
-        addLocationSearch({ location: geocodedLocation });
-      });
-  };
-
-  // Add to the selection handler lookup (for use in onKeyDown)
-  locationSelectedLookup[itemIndex] = locationSelected;
 
   // Extract GTFS/POI info and assign to class
   const { id, layer, secondaryLabels, source } = feature.properties;
@@ -210,10 +213,10 @@ const renderFeature = (
       classes={classNames.join(" ")}
       color={layerColorMap[layer]}
       icon={operatorIcon || <GeocodedOptionIconComponent feature={feature} />}
-      id={getOptionId(itemIndex)}
-      isActive={itemIndex === activeIndex}
-      key={optionKey++}
-      onClick={locationSelected}
+      id={feature.id}
+      isActive={featureItems[activeIndex]?.id === feature.id}
+      key={feature.id}
+      onClick={feature.locationSelected}
       title={main}
       subTitle={secondary}
       secondaryLabels={secondaryLabels}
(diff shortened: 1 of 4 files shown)


packages/location-field/src/index.tsx
```diff
@@ -240,20 +245,16 @@ const FeatureHeader = ({
...
   headingType,
-  itemIndex,
   layerColorMap,
-  locationSelectedLookup,
   operatorIconMap,
-  setLocation,
   showSecondaryLabels,
-  title
+  title,
+  indexedOptionLookup
```
binh-dam-ibigroup, line 257: Sort props

packages/location-field/src/index.tsx
```diff
@@ -698,14 +712,18 @@ const LocationField = ({
...
-  const locationSelectedLookup = {}; // maps itemIndex to a location selection handler (for use by the onKeyDown method)
+  const indexedOptionLookup: IndexedOptionLookup = []; // array of menu item ids and associated locationSelected handlers for onKeyDown (does not include non-selectable items e.g. dividers/headings)
   const userLocationRenderData = showUserSettings
     ? userLocationsAndRecentPlaces.map(loc =>
         getRenderData(loc, setLocation, UserLocationIconComponent, intl)
       )
     : [];
 
+  const pushToIndexedOptions = (locationSelected, id) => {
+    const obj = { id, locationSelected };
```
binh-dam-ibigroup, line 723: Nit - Maybe inline this variable.

packages/location-field/src/index.tsx
```diff
@@ -715,27 +733,41 @@ const LocationField = ({
     if (matchingLocations.length) {
       // Iterate through any saved locations
       menuItems = menuItems.concat(
-        matchingLocations.map(userLocation =>
-          makeUserOption(
+        matchingLocations.map(userLocation => {
+          const featureId = generateOptionId(userLocation, indexedOptionLookup);
```
binh-dam-ibigroup, line 737: Nit - Maybe inline this variable and use the previous return syntax.

packages/location-field/src/index.tsx
```diff
@@ -747,16 +779,15 @@ const LocationField = ({
...
         setLocation,
         addLocationSearch,
         showSecondaryLabels,
-        locationSelectedLookup,
         activeIndex,
         GeocodedOptionIconComponent,
         layerColorMap,
-        geocoderConfig
+        geocoderConfig,
+        indexedOptionLookup
```
binh-dam-ibigroup, line 790: Nit - Sort attributes.

packages/location-field/src/index.tsx
```diff
@@ -983,23 +1025,24 @@ const LocationField = ({
     statusMessages.push(optionTitle);
   }
 
-  // Add to the selection handler lookup (for use in onKeyDown)
-  locationSelectedLookup[itemIndex] = locationSelected;
+  const optionId = `current-position`;
+
+  // Add to the selection handler lookup (for use in onKeyDown
```
binh-dam-ibigroup, line 1030: Typo - add closing parenthesis.

packages/location-field/src/index.tsx
```diff
@@ -46,8 +48,27 @@ function getOptionId(index: number): string {
...
+
+  // User saved locations do not have an id, so use the label.
+  if (feature.displayName) {
+    id = feature.displayName.replace(/\s/g, "");
+  }
+
+  // Some geocoders will return an id. Use that if it exists.
+  if (feature.properties) {
+    id = feature.properties.id || feature.properties.label.replace(/\s/g, "");
+  }
```
binh-dam-ibigroup, line 63: I am in favor of trying to look for a feature id first, then fallback on using the feature label. What do you think?

packages/location-field/src/index.tsx
```diff
@@ -46,8 +48,27 @@ function getOptionId(index: number): string {
...
+  let id = "000";
+
+  // User saved locations do not have an id, so use the label.
+  if (feature.displayName) {
+    id = feature.displayName.replace(/\s/g, "");
+  }
+
+  // Some geocoders will return an id. Use that if it exists.
+  if (feature.properties) {
+    id = feature.properties.id || feature.properties.label.replace(/\s/g, "");
```
binh-dam-ibigroup, line 62: What is the reason for removing blank spaces from the ids? (Add comment)

packages/location-field/src/index.tsx
```diff
@@ -938,15 +979,16 @@ const LocationField = ({
 
     // Iterate through any saved locations
     menuItems = menuItems.concat(
-      userLocationRenderData.map(userLocation =>
-        makeUserOption(
+      userLocationRenderData.map(userLocation => {
+        const locationId = generateOptionId(userLocation, indexedOptionLookup);
```
binh-dam-ibigroup, line 983: Nit - Maybe inline this variable and use the previous return syntax.

packages/location-field/src/index.tsx
```diff
@@ -860,20 +898,19 @@ const LocationField = ({
         };
 
         // Add to the selection handler lookup (for use in onKeyDown)
-        locationSelectedLookup[itemIndex] = locationSelected;
+        pushToIndexedOptions(locationSelected, stopId);
 
         // Create and return the option menu item
         const option = (
```
binh-dam-ibigroup, line 904: Nit - inline this return.

packages/location-field/src/index.tsx
```diff
@@ -902,22 +939,26 @@ const LocationField = ({
...
+        const locationId = generateOptionId(
+          sessionLocation,
+          indexedOptionLookup
+        );
+
         // Add to the selection handler lookup (for use in onKeyDown)
-        locationSelectedLookup[itemIndex] = locationSelected;
+        pushToIndexedOptions(locationSelected, locationId);
         // Create and return the option menu item
         const option = (
```
binh-dam-ibigroup, line 950: Nit - inline this return.

packages/location-field/src/index.tsx
```diff
@@ -698,14 +712,18 @@ const LocationField = ({
 
   const statusMessages = [];
   let menuItems = []; // array of menu items for display (may include non-selectable items e.g. dividers/headings)
-  let itemIndex = 0; // the index of the current location-associated menu item (excluding non-selectable items)
-  const locationSelectedLookup = {}; // maps itemIndex to a location selection handler (for use by the onKeyDown method)
+  const indexedOptionLookup: IndexedOptionLookup = []; // array of menu item ids and associated locationSelected handlers for onKeyDown (does not include non-selectable items e.g. dividers/headings)
```
binh-dam-ibigroup, line 715: Place the comment above the code to be able to break up the long comment.

packages/location-field/src/index.tsx
```diff
@@ -46,8 +48,27 @@ function getOptionId(index: number): string {
...
+
+  // Some geocoders will return an id. Use that if it exists.
+  if (feature.properties) {
+    id = feature.properties.id || feature.properties.label.replace(/\s/g, "");
+  }
+
+  // There are some cases, such as when we're using user-saved locations, where
+  // the id is not unique. In this case, append a number to the end of the id.
+  if (indexedOptionLookup.find(item => item.id === id)) {
+    id = `${id}0`;
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Could you clarify in the comment above the if block whether the "unique" refer to that a user location might have the same id as another feature, or the same id as another user location? A better way to generate the id may come from that answer.

binh-dam-ibigroup (commented):
Could you clarify the piece regarding unique ids for user locations, why whitespaces are removed in ids, and your thoughts on using a feature id before falling back on the display name, please?

packages/location-field/src/index.tsx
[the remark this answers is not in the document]
amy-corson-ibigroup: The unique id issue comes from user locations appearing in the list twice, as happens sometimes when a query matches a user location. Screenshot of this below.
image

packages/location-field/src/index.tsx
amy-corson-ibigroup: No technical reason, I just rarely see ids that contain whitespace. I can leave it in if it's unneeded!

packages/location-field/src/index.tsx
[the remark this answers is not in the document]
binh-dam-ibigroup: @amy-corson-ibigroup So what would be the element ids for the two highlighted items in your screenshot?

packages/location-field/src/index.tsx
amy-corson-ibigroup: I agree, isn't that what will happen here? Wont the `||` operator only come into play when there's no id in `properties`?
If you're talking about the user locations, I can add that check but in general I was not seeing any ids for those

binh-dam-ibigroup: Sorry, I meant to fallback on `feature.displayName` if `feature.properties` is not provided!

packages/location-field/src/index.tsx
```diff
@@ -712,16 +712,17 @@ const LocationField = ({
 
   const statusMessages = [];
   let menuItems = []; // array of menu items for display (may include non-selectable items e.g. dividers/headings)
-  const indexedOptionLookup: IndexedOptionLookup = []; // array of menu item ids and associated locationSelected handlers for onKeyDown (does not include non-selectable items e.g. dividers/headings)
+  /* array of menu item ids and associated locationSelected handlers for onKeyDown
```
binh-dam-ibigroup, line 715: Use `//` comments.

binh-dam-ibigroup (approved):
Approving, consider addressing https://github.com/opentripplanner/otp-ui/pull/826/files#r2050802384. Thanks for addressing the nits. I think the IDs should be okay.

packages/location-field/src/index.tsx
```diff
@@ -46,8 +48,27 @@ function getOptionId(index: number): string {
   return `${optionIdPrefix}-${index}`;
 }
 
-// FIXME have a better key generator for options
-let optionKey = 0;
+function generateOptionId(feature, indexedOptionLookup) {
```
josh-willis-arcadis, line 51: nit: should these be typed?

packages/location-field/src/index.tsx
```diff
@@ -46,8 +48,27 @@ function getOptionId(index: number): string {
...
+    id = feature.displayName.replace(/\s/g, "");
+  }
+
+  // There are some cases, such as when we're using user-saved locations, where
+  // the id is not unique. In this case, append a number to the end of the id.
+  if (indexedOptionLookup.find(item => item.id === id)) {
+    id = `${id}-0`;
+  }
+  return id;
+}
```
josh-willis-arcadis, line 71: This function works perfectly as is, however I see a way to simplify the logic flow. Does this to make it seem less hacky?
```
function generateOptionId(feature, indexedOptionLookup) {
  // Get the base ID from the most appropriate source
  const baseId = feature.properties?.id || 
                 feature.properties?.label?.replace(/\s/g, "") ||
                 feature.displayName?.replace(/\s/g, "") ||
                 "000";

  // Check for duplicates and append number if needed
  if (!indexedOptionLookup.some(item => item.id === baseId)) {
    return baseId;
  }

  // Find the next available number to append
  let counter = 0;
  let newId = `${baseId}-${counter}`;
  
  while (indexedOptionLookup.some(item => item.id === newId)) {
    counter++;
    newId = `${baseId}-${counter}`;
  }

  return newId;
}
```

josh-willis-arcadis (approved):
Just a few nits. LGTM!

packages/location-field/src/index.tsx
amy-corson-ibigroup: Unfortunately the way `newId` is being referenced here is causing the `counter` not to iterate properly, so it's causing this issue: 
image
https://eslint.org/docs/latest/rules/no-loop-func#options That's probably fixable though!


Secondly, genuinely curious if using a loop here might be less efficient? The way this is written, each new non-unique id is going to take an additional iteration from the id before it. For example, if you're working with a list of 25 entities without ids, the last time this function is called you're going to have to loop over the `while` block 25 times before it generates `000-24`.  I'll defer to @binh-dam-ibigroup though, I might be wrong!

binh-dam-ibigroup: Using a counter to generate an id was not ideal to begin with. (It goes back to when this component was initially created.) We could get rid of the counter and magic `000` altogether and replace it with an ID scheme such as `<category>-<feature-id/name>` (e.g. `stop-Angle Lake`, `favorites-Home`, `geocoder-f456ae1304`, `recent-123 Main Street`), what do you think?

amy-corson-ibigroup: Thanks for the idea Binh, instead of the counter I added a option prefix to the function that should fix the issue discussed above (multiple identical user saved locations in the list). If there are multiple features without ids or display names in a given category we probably have bigger issues.

amy-corson-ibigroup: @binh-dam-ibigroup @josh-willis-arcadis if you could both confirm this works and no outstanding issues, I will merge!

josh-willis-arcadis: This ID scheme seems like a better solution. Looping could get pretty inefficient overtime. @amy-corson-ibigroup

binh-dam-ibigroup (changes requested):
Can you please remove the "000" that remains?

packages/location-field/src/index.tsx
```diff
@@ -36,19 +36,19 @@ import {
...
-  return `${optionIdPrefix}-${index}`;
+type IndexedOptionLookup = Array<{ id: string; locationSelected: () => void }>;
+
+function generateOptionId(optionPrefix, feature) {
+  // Use a default id if the feature does not have an id.
+  const featureId =
+    feature.properties?.id ||
+    feature.properties?.label ||
+    feature.displayName ||
+    "000";
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Good idea, added in [d7f70a4](https://github.com/opentripplanner/otp-ui/pull/826/commits/d7f70a4d1296089c512ec95e38b535be187445fb)

binh-dam-ibigroup (approved):
Thanks for all the changes!

amy-corson-ibigroup:
Thanks so much to everyone for the discussion and feedback!!

PR #829, opened 2025-04-22
fix(base-map): fix firefox printing bug

amy-corson-ibigroup:
Trying to print the printable itinerary in firefox resulted in the map being empty. Added `preserveDrawingBuffer` per this  thread  https://github.com/mapbox/mapbox-gl-js/issues/9810

| Before | After |
|--------|-------|
|image|image|

### DIFF
--- a/packages/base-map/src/index.tsx
+++ b/packages/base-map/src/index.tsx
@@ -161,6 +161,7 @@ const BaseMap = ({
       }}
       onTouchCancel={clearLongPressTimer}
       onTouchEnd={clearLongPressTimer}
+      preserveDrawingBuffer
       style={style}
       zoom={viewState.zoom}
     >


binh-dam-ibigroup (approved):
Good find!

packages/base-map/src/index.tsx
```diff
@@ -161,6 +161,7 @@ const BaseMap = ({
       }}
       onTouchCancel={clearLongPressTimer}
       onTouchEnd={clearLongPressTimer}
+      preserveDrawingBuffer
```
binh-dam-ibigroup, line 164: FYI: This prop is no longer supported in maplibre-gl 5.x: https://github.com/maplibre/maplibre-gl-js/releases/tag/v5.0.0. If this feature is important, that change to maplibre will limit our upgrade options.

binh-dam-ibigroup: The current `preserveDrawingBuffer` prop touches the same attribute in `MapOptions`, however that attribute is elsewhere in maplibre-gl v5.x, and react-map-gl v7.1 and above removed support for `preserveDrawingBuffer` without adding an alternate way to touch that attribute in maplibre-gl.

PR #831, opened 2025-04-25
fix(location-field): if no user text then clear geocoder results

josh-willis-arcadis:
Address a bug to clear geocoder results when there is no text in the search bar.


amy-corson-ibigroup (approved):
Nice!

PR #832, opened 2025-04-29
fix: Create mutableLeg for utility operation flexibility

lightsage88:
While working on integrating a new provider into our trimet.org application, I encountered an instance of a flexRoute which within the existing code calls upon the runtime for assignments to a property of an object that is a element going through iteration.

Typescript does not allow this and it was crashing my application. In this pull-request, I import clonedeep from lodash and create a 'mutableLeg' -- that is a copy of leg -- right before the if-statements that wrap the commands to assign data to iteration elements.


packages/transitive-overlay/src/util.ts
```diff
@@ -356,73 +360,79 @@ export function itineraryToTransitive(
...
+              mutableLeg.interStopGeometry[i].points
           });
         });
       }
 
       // Add the "to" end of transit legs to the list of stops.
       // (Do not label stop names if they repeat.)
       const lastCoord = hasLegGeometry && legCoords[legCoords.length - 1];
-      const toStop = makeStop(leg.to, lastCoord);
+      const toStop = makeStop(mutableLeg.to, lastCoord);
```
binh-dam-ibigroup, line 394: I would create a method that clones the original `to` stop and includes the flex stuff from lines 328/330. The main benefit is that you don't rename/clone the entire leg just to change a small piece of it, given that the cloned/modified leg is not used anywhere else other than here.

packages/transitive-overlay/src/util.ts
```diff
@@ -318,32 +319,35 @@ export function itineraryToTransitive(
...
       // these stops still need to be rendered separately!
-      if (isFlex(leg)) {
-        if (typeof leg.to.stop === "object") {
-          leg.to.stop.id = `${leg.to.stop.id}_flexed_to`;
+      if (isFlex(mutableLeg)) {
+        if (typeof mutableLeg.to.stop === "object") {
+          mutableLeg.to.stop.id = `${mutableLeg.to.stop.id}_flexed_to`;
         } else {
-          leg.to.stopId = `${leg.to.stopId}_flexed_to`;
+          mutableLeg.to.stopId = `${mutableLeg.to.stopId}_flexed_to`;
```
binh-dam-ibigroup, line 330: Move this to a new method that creates a clone of the `to` stop.

binh-dam-ibigroup (changes requested):
It is a good thing to not modify props. I think cloning the entire leg without further use is overkill. Please modify the implementation per https://github.com/opentripplanner/otp-ui/pull/832/files#r2067255428 to minimize copying fields and renaming vars while keeping the origina `leg` prop untouched.

packages/transitive-overlay/src/util.ts
```diff
@@ -177,6 +177,27 @@ function makeStop(stop: Place, coordinate?: number[]) {
...
+    toStop.stop = { ...toStop.stop };
+  }
+
+  if (isFlex(leg)) {
+    if (typeof toStop.stop === "object") {
+      toStop.stop.id = `${toStop.stop.id}_flexed_to`;
+    } else {
+      toStop.stopId = `${toStop.stopId}_flexed_to`;
+    }
+  }
```
miles-grant-ibigroup, line 196: This behavior is not related to stop cloning, and should therefore not be in this function

packages/transitive-overlay/src/util.ts
```diff
@@ -177,6 +177,27 @@ function makeStop(stop: Place, coordinate?: number[]) {
     : stop;
 }
 
+/**
+ * Helper function that clones the 'to' stop and applies flex logic
+ */
+function cloneToStop(leg: Leg): Place {
+  const toStop = { ...leg.to };
```
miles-grant-ibigroup, line 184: We don't know that `leg.to` will be a stop! Let's make that clear in the variable name

miles-grant-ibigroup (changes requested):
A few cleanup opportunities!

packages/transitive-overlay/src/util.ts
```diff
@@ -221,6 +242,10 @@ export function itineraryToTransitive(
   newPlaces.push(makeFromToPlace(itin.legs[itin.legs.length - 1].to, "to"));
 
   itin.legs.forEach((leg, idx) => {
+    // Create shallow copy of the leg
+    const mutableLeg = { ...leg };
```
binh-dam-ibigroup, line 246: After introducing the function `cloneToStop`, it is no longer necessary to create a shallow clone of the leg. A better to do that is to call `cloneToStop` at line 411 by writing:
```js
// Line 411 below
      const toStop = makeStop(cloneToStop(leg), lastCoord);
```

packages/transitive-overlay/src/util.ts
```diff
@@ -323,10 +348,10 @@ export function itineraryToTransitive(
...
       if (isFlex(leg)) {
-        if (typeof leg.to.stop === "object") {
-          leg.to.stop.id = `${leg.to.stop.id}_flexed_to`;
+        if (typeof modifiedToStop.stop === "object") {
+          modifiedToStop.stop.id = `${modifiedToStop.stop.id}_flexed_to`;
         } else {
-          leg.to.stopId = `${leg.to.stopId}_flexed_to`;
+          modifiedToStop.stopId = `${modifiedToStop.stopId}_flexed_to`;
         }
       }
```
binh-dam-ibigroup, line 356: This block is covered in new function `cloneToStop` and should now be removed from here.

packages/transitive-overlay/src/util.ts
```diff
@@ -323,10 +348,10 @@ export function itineraryToTransitive(
       // Flex routes sometimes have the same from and to IDs, but
       // these stops still need to be rendered separately!
```
binh-dam-ibigroup, line 349: Move this comment to new function `cloneToStop`.

packages/transitive-overlay/src/util.ts
binh-dam-ibigroup: @lightsage88 Rename this function to `makeToStopWithFlexSupport`, and this block will definitely belong here, and that was the intent for extracting this method in the first place. @miles-grant-ibigroup

packages/transitive-overlay/src/util.ts
lightsage88: I renamed it to _flexSupportStop_ in my soon to be pushed commit!

packages/transitive-overlay/src/util.ts
lightsage88: I am renaming it to _makeStopWithFlexSupport_ as a response to this comment: https://github.com/opentripplanner/otp-ui/pull/832#discussion_r2068627210

lightsage88:
Okay guys, I think I covered all your questions and I've confirmed it's working for my application. I can't seem to avoid changes to the pnpm-lock.yaml. In fact, I even pasted the one you have in master into my branch to ensure there'd be no divergences but it seems to flag that there are changes within it still. Odd, I'm not sure what to do about that.

binh-dam-ibigroup (approved):
Looks better after the suggested changes.

packages/transitive-overlay/src/util.ts
```diff
@@ -177,6 +177,29 @@ function makeStop(stop: Place, coordinate?: number[]) {
     : stop;
 }
 
+/**
+ * Helper function that clones the 'to' stop and applies flex logic
+ * Flex routes sometimes have the same from and to IDs, but
+ * these stops still need to be rendered separately!
+ */
+function makeStopWithFlexSupport(leg: Leg): Place {
+  const flexSupportStop = { ...leg.to };
```
miles-grant-ibigroup, line 186: This is not abstract enough. Can the function clone any place its given, and then have it accept `leg.to`?

miles-grant-ibigroup (changes requested):
Can be cleaner

packages/transitive-overlay/src/util.ts
binh-dam-ibigroup: @miles-grant-ibigroup I thought about suggesting this change, however there is a call to `isFlex(leg)` in the function's body.

miles-grant-ibigroup: Hmm that's right. Could we then rename this method to `renameLegFlexStops`?

miles-grant-ibigroup (changes requested):

packages/transitive-overlay/src/util.ts
lightsage88: Yes, I'll do that right away!

binh-dam-ibigroup (approved):

PR #835, opened 2025-05-07
Highlight Interlines

miles-grant-ibigroup:
Allows implementers the option to highlight interlined legs by adding a class name when legs are interlined


josh-willis-arcadis (approved):
LGTM

amy-corson-ibigroup (approved):

PR #836, opened 2025-05-14
feat(trip-form): Update Date Time Selector and add seperate dropdown

amy-corson-ibigroup:
BREAKING CHANGE: major change to d/t selector

### DIFF
--- a/packages/trip-form/src/DateTimeSelector/index.tsx
+++ b/packages/trip-form/src/DateTimeSelector/index.tsx
@@ -1,24 +1,15 @@
 import CSS from "csstype";
 import { format, parse } from "date-fns";
-import flatten from "flat";
 import coreUtils from "@opentripplanner/core-utils";
-import React, { ChangeEvent, ReactElement, ReactNode, useCallback } from "react";
-import { FormattedMessage, useIntl } from "react-intl";
+import React, { ChangeEvent, ReactElement, useCallback } from "react";
+import { useIntl } from "react-intl";
 
-import ModeButton from "../ModeButton";
+import { Dropdown } from "@opentripplanner/building-blocks";
 import * as S from "../styled";
 
 // eslint-disable-next-line prettier/prettier
 import type { QueryParamChangeEvent } from "../types";
 
-// Load the default messages.
-import defaultEnglishMessages from "../../i18n/en-US.yml";
-
-// HACK: We should flatten the messages loaded above because
-// the YAML loaders behave differently between webpack and our version of jest:
-// - the yaml loader for webpack returns a nested object,
-// - the yaml loader for jest returns messages with flattened ids.
-const defaultMessages: Record<string, string> = flatten(defaultEnglishMessages);
 
 const {
   getCurrentDate,
@@ -47,6 +38,10 @@ interface DateTimeSelectorProps {
    * The initial setting determining whether a trip should start or end at a given time.
    */
   departArrive?: DepartArriveValue;
+  /**
+   * If true, only render the date/time inputs. The depart/arrive selector will be handled by the parent component.
+   */
+  externalDepartArriveSelector?: boolean;
   /**
    * If true, forces legacy mode and uses `<input type="text">`
    * instead of the native date/time pickers found on modern browsers.
@@ -77,8 +72,7 @@ interface DateTimeSelectorProps {
 }
 
 interface DepartArriveOption {
-  isSelected?: boolean;
-  text: ReactNode;
+  text: string;
   type: DepartArriveValue;
 }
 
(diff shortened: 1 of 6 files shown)


packages/trip-form/src/styled.ts
```diff
@@ -8,6 +9,12 @@ import type {
   SubmodeSelectorAndSubComponents
 } from "./types";
 
+const { grey } = colors;
+
+export const baseColor = () => getComputedStyle(document.documentElement).getPropertyValue(
+  "--main-base-color"
+)
+
```
miles-grant-ibigroup, line 17: Don't we have this function inside core utils already?

packages/trip-form/src/styled.ts
```diff
@@ -21,31 +28,28 @@ export const FloatingSettingLabel = styled(SettingLabel)`
...
-  box-sizing: border-box;
-  > * {
-    box-sizing: border-box;
-    width: 33.333333%;
-    padding: 0px 5px;
-  }
-`;
+export const DateTimeSelector: DateTimeSelectorAndSubComponents = styled.div`
+  border-radius: 5px;
+  background: ${grey[50]};
```
miles-grant-ibigroup, line 33: Why are we not using the base color and text color from the variables? I think it's possible to set a fallback too

miles-grant-ibigroup (changes requested):
This looks so good! I'm really excited for this to deploy. I have a few nitpicks for now but overall this looks great!

packages/trip-form/src/styled.ts
amy-corson-ibigroup: We do not :(

packages/trip-form/src/styled.ts
amy-corson-ibigroup: I forgot we talked about this! Added

packages/trip-form/src/DateTimeSelector/index.tsx
```diff
@@ -99,11 +95,78 @@ function isInputTypeSupported(type: string): boolean {
...
+    type: "DEPART",
+    text: intl.formatMessage({ id: "otpUi.DateTimeSelector.depart" })
+  },
+  {
+    type: "ARRIVE",
+    text: intl.formatMessage({ id: "otpUi.DateTimeSelector.arrive" })
+  }
+];
+
+const handleQueryParamChange = (onQueryParamChange, queryParam) => {
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Removed! Thanks

packages/trip-form/src/DateTimeSelector/index.tsx
```diff
@@ -138,20 +202,19 @@ export default function DateTimeSelector({
...
+
+
 
   const handleInputChange = (key: string) => useCallback(
     (evt: ChangeEvent<HTMLInputElement>): void => {
-      handleQueryParamChange({ [key]: evt.target.value });
+      handleQueryParamChange(onQueryParamChange, { [key]: evt.target.value });
+      // If the user changes the time, it doesn't make sense for them to be departing now.
+      if (departArrive === "NOW") {
+        handleQueryParamChange(onQueryParamChange, { departArrive: "DEPART" });
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Huh, originally I think i created the `handleQueryParamChange` function to fix a typeerror but it doesn't seem to be an issue? I didn't add optional chaining but I can!

daniel-heppner-ibigroup (approved):
Looks great to me! Thanks for the fixes. I noticed that the default value on onQueryParamChange wasn't needed so I just removed it and pushed it myself. 

Before merging can you get @miles-grant-ibigroup to sign off on my suggested default for the departArriveDropdown default? I think it just makes logical sense that the default setting would be for this to be just a date time selector, and only display that dropdown if we need it. But it is a departure from the previous philosophy so wanted to see if Miles agrees!

amy-corson-ibigroup:
Thanks Daniel!!

Noting also that [3da40a9](https://github.com/opentripplanner/otp-ui/pull/836/commits/3da40a9c01d3a92d2a0ede5b6a6211eaa08f8403) is a breaking change commit just so everyone's aware

miles-grant-ibigroup (changes requested):
I like the new props! But I think the stories should be more clear about how the component works and what the props do.

Also, could we edit the breaking change commit to have a more meaningful message? Something like "api changed, dropdown now optional"

packages/trip-form/src/components.story.tsx
```diff
@@ -131,13 +131,31 @@ export const dateTimeSelector = makeStory(Core.DateTimeSelector, {
...
+  dateFormatLegacy: "YY-M-d",
+  departArrive: "NOW",
+  forceLegacy: false,
+  onQueryParamChange,
+  time: "14:17",
+  timeFormatLegacy: "HH:mm",
+  timeZone: "America/New_York"
+});
+
+export const departArriveDropdown = makeStory(Core.DepartArriveDropdown, {
```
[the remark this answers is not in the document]
amy-corson-ibigroup: I've updated the d/t selector story to include the `departArriveDropdown` control. The `DepartArriveDropdown` stroy should remain separate because it's technically its own component

PR #837, opened 2025-05-19
fix(itinerary-body):  Correctly apply aria-expanded to alerts button

amy-corson-ibigroup:
The alert button inside the itinerary body needed an `aria-expanded` prop, however we can't add that to a static `div`. This splits the alerts button into `button` and `div` elements and applies the appropriate props to each.

It also centers the alert label, which was previously off-center

### DIFF
--- a/packages/itinerary-body/src/TransitLegBody/index.tsx
+++ b/packages/itinerary-body/src/TransitLegBody/index.tsx
@@ -184,6 +184,20 @@ class TransitLegBody extends Component<Props, State> {
         defaultFareSelector.riderCategoryId
       );
 
+    const alertLabelContents = (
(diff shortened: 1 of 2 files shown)


binh-dam-ibigroup (changes requested):
Looks good, but could you simplify the alert toggle render as indicated and update the snapshots?

binh-dam-ibigroup (approved):

miles-grant-ibigroup (changes requested):
This story appears to be broken! Screenshot 2025-05-23 at 3 10 28 PM

I'm also a little unsure about this one

Screenshot 2025-05-23 at 3 10 44 PM

PR #838, opened 2025-05-19
fix: Clean up temporary directory

daniel-heppner-ibigroup:
Clean up the temporary directory before extracting the tar to eliminate problems when the version changes or files are removed.


miles-grant-ibigroup (approved):
Working well

PR #839, opened 2025-05-21
fix: Add new optional property to Place type

lightsage88:
This is meant to be step one of a two part pull-request situation.

This adds an optional value to the _Place_ type:
```
vehicleRentalStation?: { rentalNetwork: { networkId: string } };
```

The follow up pull request will add the following code in  the getCompanyFromLeg method in _core-utils/src/itinerary.ts_

```
  if (from.vehicleRentalStation && from.vehicleRentalStation.rentalNetwork) {
    return from.vehicleRentalStation.rentalNetwork.networkId;
  }
```

And all of this is so that in our client-code at Trimet, we can avoid having to do the following:
```
  const company =
    coreUtils.itinerary.getCompanyFromLeg(leg) ||
    leg.from.vehicleRentalStation?.rentalNetwork.networkId;
```
and instead just have this:
```
const company = coreUtils.itinerary.getCompanyFromLeg(leg);
```

### DIFF
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -251,6 +251,7 @@ export type Place = {
   name: string;
   networks?: string[];
   rentalVehicle?: { network: string };
+  vehicleRentalStation?: { rentalNetwork: { networkId: string } };
   stop?: Stop;
   /**
    * @deprecated Only for OTP1 support, removal is immenent


packages/types/src/index.ts
```diff
@@ -251,6 +251,7 @@ export type Place = {
   name: string;
   networks?: string[];
   rentalVehicle?: { network: string };
+  vehicleRentalStation?: { rentalNetwork: { networkId: string } };
```
miles-grant-ibigroup, line 254: @daniel-heppner-ibigroup could you take a look at this? Doesn't this imply that the `rentalNetwork` is always there? Is there some way to read this from the graphql?

miles-grant-ibigroup (changes requested):
Paging @daniel-heppner-ibigroup to help to make sure these types are correct

packages/types/src/index.ts
lightsage88: I could put a ? by rentalNetwork. I mean the vehicleRentalStation is optional as it is and each body of data for a rentalStation is going to need to have some reference to what network it's a part of...

image


This applies to when we are planning rental trips that originate or have a leg in the trip that leverages the use of a rental hub. Free floating rental bikes have never been a problem.

lightsage88: Just let me know and I can whip those adjustments up in a jiffy! 🥜 🧈

daniel-heppner-ibigroup: Rental network and network ID are required fields in the schema so this is fine!

daniel-heppner-ibigroup: image

miles-grant-ibigroup (approved):
Thanks @daniel-heppner-ibigroup!

daniel-heppner-ibigroup:
By the way, there is no need to split these pull requests anymore. pnpm can handle having them both get updated in the same PR now!

lightsage88:
Oh...so should I include the changes I want to make to core-utils in this pull-request as well? Won't that result in multiple updates to package version numbers? If so I feel like having individual ones would be less messy, right?

daniel-heppner-ibigroup:
It's fine as is! But in the future you don't need to, since the version numbers in package.json have been replaced with `workspace`. They can both be released in the same PR without issue.

binh-dam-ibigroup (approved):

lightsage88:
Thanks for your help everyone! I have created !840 as a follow up to this pull-request. Next time, I'll do it all in one!

PR #846, opened 2025-05-29
fix(location-field): add type to location field clear button

amy-corson-ibigroup:
Fixes the clear button so it wont default to `type='submit'` when within a form


josh-willis-arcadis (approved):
great catch!

PR #847, opened 2025-05-30
fix(map-popup): remove stop id

josh-willis-arcadis:
Removes logic to display GTFS stop_id in the map popup package as it is not a public facing value.

### DIFF
--- a/packages/map-popup/src/index.tsx
+++ b/packages/map-popup/src/index.tsx
@@ -63,7 +63,7 @@ const StationHubDetails = ({ station }: { station: Station }) => {
 const StopDetails = ({ id, setViewedStop }: { id: string, setViewedStop: () => void; }) => {
   return (
     <Styled.PopupRow>
-      { id &&
+      {id &&
         <strong>
           <FormattedMessage
             defaultMessage={defaultMessages["otpUi.MapPopup.stopId"]}
@@ -115,7 +115,7 @@ export function MapPopup({ closePopup = () => {}, configCompanies, entity, getEn
 
   const bikesAvailablePresent = entityIsStation(entity)
   const entityIsStationHub = bikesAvailablePresent && entity?.bikesAvailable !== undefined && !entity?.isFloatingBike;
-  const stopId = !bikesAvailablePresent && entity?.code;  // Double quotes make the query invalid, so remove them from the id just in case
+  const stopId = !bikesAvailablePresent && entity?.code;
   const id = `focus-${encodeURIComponent(entity.id).replace(/%/g, "")}-popup`
 
   return (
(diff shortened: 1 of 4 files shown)


daniel-heppner-ibigroup (changes requested):
a few comments! 
Please also add a new story *with* a stop code so that we can see what it looks like when there's a stop code stop. then update snapshots

daniel-heppner-ibigroup (approved):
Excellent work! Thank you!

packages/map-popup/src/index.tsx
```diff
@@ -63,17 +63,18 @@ const StationHubDetails = ({ station }: { station: Station }) => {
...
-          defaultMessage={defaultMessages["otpUi.MapPopup.stopId"]}
-          description="Displays the stop id"
-          id="otpUi.MapPopup.stopId"
-          values={{
-            stopId: id
-          }}
-        />
-      </strong>
-      <ViewStopButton onClick={setViewedStop}>
+      { id &&
```
miles-grant-ibigroup, line 66: Can we remove this extra space?

```suggestion
      {id &&
```

packages/map-popup/src/index.tsx
```diff
@@ -114,9 +115,7 @@ export function MapPopup({ closePopup = () => {}, configCompanies, entity, getEn
 
   const bikesAvailablePresent = entityIsStation(entity)
   const entityIsStationHub = bikesAvailablePresent && entity?.bikesAvailable !== undefined && !entity?.isFloatingBike;
-  const stopId = !bikesAvailablePresent && entity?.code || entity.id.split(":")[1] || entity.id
-
-  // Double quotes make the query invalid, so remove them from the id just in case
+  const stopId = !bikesAvailablePresent && entity?.code;  // Double quotes make the query invalid, so remove them from the id just in case
```
miles-grant-ibigroup, line 118: You can remove this comment

```suggestion
  const stopId = !bikesAvailablePresent && entity?.code;
```

miles-grant-ibigroup (changes requested):
A few nitpicks

packages/map-popup/src/styled.ts
```diff
@@ -1,14 +1,16 @@
...
-  border-left: 1px solid #000;
+  ${props =>
+    props.stopId != null
+      ? "border-left: 1px solid #000; margin-left: 5px;"
+      : "border-left: none"};
   border-right: none;
   border-top: none;
   color: #008;
   font-family: inherit;
-  margin-left: 5px;
```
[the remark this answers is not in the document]
josh-willis-arcadis: addressed in f6726d5

miles-grant-ibigroup:
Once you update snapshots this should be good to go!

PR #851, opened 2025-06-17
fix: adapt loadImages to handle differences in maplibre versions

lightsage88:
The gist of this pull-request is to solve an issue with Trimet's application where upon using transitive-overlay version 6.0.0 with our maplibre version of the application, and an itinerary is created the bubbles that contain the name of the transit line are transparent (i.e., they don't load).

This seems to be a result of a breaking change between versions of mapLibre 1/2 vs 3/4.

1/2 use Promises to return the png images
3/4 use a callback

This pull-request refactors _loadImages_ in the packages/transitive-overlay/src/index.tsx file such that either version of mapLibre can be satisfied so that the images can appear.

To do this, however, I needed to leverage heavy use of @ts-ignore, for which I received many linting/build errors.

### DIFF
--- a/packages/transitive-overlay/src/index.tsx
+++ b/packages/transitive-overlay/src/index.tsx
@@ -1,3 +1,6 @@
+/* eslint-disable no-restricted-syntax */
+/* eslint-disable no-await-in-loop */
+/* eslint-disable @typescript-eslint/ban-ts-comment */
 import { SymbolLayout } from "mapbox-gl";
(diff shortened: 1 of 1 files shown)


lightsage88:
Hey, @binh-dam-ibigroup, @miles-grant-ibigroup mentioned that you guys were on the verge on updating MapLibre for the library. If I understand things correctly, yes I think that would do the trick!

PR #852, opened 2025-06-18
Alert AT that subsettings are expanded in transit options

amy-corson-ibigroup:
Checkboxes do not support `aria-expanded` so add an additional string to the `aria-label` to include whether an option has expanded suboptions
image

### DIFF
--- a/packages/trip-form/src/MetroModeSelector/AdvancedModeSettingsButton/index.tsx
+++ b/packages/trip-form/src/MetroModeSelector/AdvancedModeSettingsButton/index.tsx
@@ -110,18 +110,28 @@ const AdvancedModeSettingsButton = ({
   const checkboxId = `metro-submode-selector-mode-${id}`;
   const containsSubsettings = modeButton.modeSettings.length > 0;
 
+  const subsettingsStatusLabel = modeButton.enabled
+    ? intl.formatMessage({
+        id: "otpUi.ModeSelector.labels.subsettingsExpanded"
+      })
+    : intl.formatMessage({
+        id: "otpUi.ModeSelector.labels.subsettingsCollapsed"
+      });
+  const ariaLabel = containsSubsettings
+    ? label + subsettingsStatusLabel
+    : label;
+
   return (
     <SettingsContainer className="advanced-submode-container">
       <StyledModeSettingsButton
         accentColor={accentColor}
-        aria-expanded={containsSubsettings ? !!modeButton.enabled : undefined}
         className="advanced-submode-mode-button"
         fillModeIcons={fillModeIcons}
         id={modeButton.key}
         subsettings={containsSubsettings}
       >
         <input
-          aria-label={label}
+          aria-label={ariaLabel}
           checked={modeButton.enabled ?? undefined}
           id={checkboxId}
           onChange={onToggle}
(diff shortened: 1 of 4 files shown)


packages/trip-form/src/MetroModeSelector/AdvancedModeSettingsButton/index.tsx
```diff
@@ -108,17 +108,30 @@ const AdvancedModeSettingsButton = ({
...
         accentColor={accentColor}
         className="advanced-submode-mode-button"
         fillModeIcons={fillModeIcons}
         id={modeButton.key}
-        subsettings={modeButton.modeSettings.length > 0}
+        subsettings={containsSubsettings}
       >
         <input
-          aria-label={label}
+          aria-label={ariaLabel}
```
binh-dam-ibigroup, line 134: Improvement: We probably don't need `aria-label`, instead you could place the `subsettingsStatusLabel` inside an `{" "}<InvisibleA11yLabel>` within the `<label>` element below. That way, you can remove the leading space from the i18n content.

packages/trip-form/src/MetroModeSelector/AdvancedModeSettingsButton/index.tsx
```diff
@@ -130,7 +143,7 @@ const AdvancedModeSettingsButton = ({
           {modeButton.enabled && <Check2 />}
         </label>
       </StyledModeSettingsButton>
-      {modeButton.modeSettings.length > 0 && (
+      {containsSubsettings && (
         <AnimateHeight duration={300} height={modeButton.enabled ? "auto" : 0}>
```
binh-dam-ibigroup, line 147: Add `aria-expanded={modebutton.enabled}` to this element. Maybe add a `aria-describedby` or `aria-details` to the checkbox that points to this element, in addition or in place of `subsettingsStatusLabel`.

binh-dam-ibigroup (changes requested):
Could you add an `aria-expanded` prop and improve a11y labeling as suggested, please?

packages/trip-form/src/MetroModeSelector/AdvancedModeSettingsButton/index.tsx
amy-corson-ibigroup: Because `<InvisibleA11yLabel>` would be a [nested element inside a label](https://stackoverflow.com/questions/72090573/safari-voiceover-not-reading-labels-with-nested-elements-correctly), screenreaders read it as "Transit and one other item" instead of the full string. This was the easiest way I could find to make sure the full string and no other confusing information was presented.

binh-dam-ibigroup: Good point, could you add a comment about that, please?

amy-corson-ibigroup: Added in [61be29d](https://github.com/opentripplanner/otp-ui/pull/852/commits/61be29d78b927eca4dbde0a5b350ddd442afae36)

binh-dam-ibigroup (approved):
I'm good to go with what you have in this PR, if folks come up with better we'll take it.

daniel-heppner-ibigroup (approved):
Looks good! Sorry for delay

PR #853, opened 2025-06-18
chore(base-map): add translations for geolocate controls

josh-willis-arcadis:
Add translations for the Geolocate controls on the map


amy-corson-ibigroup (approved):
Looking good! Thanks for the changes

binh-dam-ibigroup (approved):

PR #854, opened 2025-06-23
Transitive Overlay Remove Un-needed refresh

miles-grant-ibigroup:
the `useEffect` was a bit too liberal, and this was causing many unnecessary requests


daniel-heppner-ibigroup (approved):
makes sense to me

PR #857, opened 2025-06-25
feat: add support for rendering agency name with the stop name in map popup

daniel-heppner-ibigroup:
This PR adds support for rendering the agency name, stop code, and stop name in the map popup used in otp2 tile overlay. It does this by accepting a prop containing a mapping of stop IDs to agencies, since this information is not available in the vector tile data. It also provides a callback prop so the component can tell its parent when a stop has been clicked on and the data is needed. 

Paired with https://github.com/opentripplanner/otp-react-redux/pull/1433

### DIFF
--- a/packages/map-popup/src/__snapshots__/MapPopup.story.tsx.snap
+++ b/packages/map-popup/src/__snapshots__/MapPopup.story.tsx.snap
@@ -385,3 +385,68 @@ exports[`Map Popup StopEntityWithEntityPrefix smoke-test 1`] = `
   </div>
 </div>
 `;
+
+exports[`Map Popup StopEntityWithFeedName smoke-test 1`] = `
+<div class="styled__MapOverlayPopup-sc-12kjso7-1 cPqJxe">
+  <div id="focus-trimet3A9526-popup-focus-trap"
+       role="presentation"
+  >
+    <header class="styled__PopupTitle-sc-12kjso7-3 jRNaQh">
+      W Burnside &amp; SW 2nd (TriMet 9526)
+    </header>
+    <p class="styled__PopupRow-sc-12kjso7-2 ckOOWr">
+      <strong>
+        Stop ID: 9526
+      </strong>
+      <button class="styled__ViewStopButton-sc-12v7ov3-0 hXaHvR">
+        Stop Viewer
+      </button>
+    </p>
+    <p class="styled__PopupRow-sc-12kjso7-2 ckOOWr">
+      <strong>
+        Plan a trip:
+      </strong>
+      <span class="styled__FromToPickerSpan-sc-vb4790-1 giBPod">
+        <span class="styled__LocationPickerSpan-sc-vb4790-0 gsVfXo">
+          <svg viewbox="0 0 512 512"
+               height="0.9em"
+               width="0.9em"
+               aria-hidden="true"
+               focusable="false"
+               fill="currentColor"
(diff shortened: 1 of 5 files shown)


packages/otp2-tile-overlay/src/index.tsx
```diff
@@ -56,6 +58,11 @@ const OTP2TileLayerWithPopup = ({
    * in accordance with the MapLibreGL api.
    */
   onMapClick?: (event: EventData) => void
+  /**
+   * A method to request the agency for a given stop ID. If this method is not passed, the agency
+   * for a given stop ID will not be displayed in the popup.
+   */
+  requestAgencyForStop?: (stopId: string) => void
```
miles-grant-ibigroup, line 65: Why do we have this as well as the map. Would it be possible to just have this instead? Feels cleaner

miles-grant-ibigroup: Can we get rid of this line then? The method is never defined

daniel-heppner-ibigroup: It's defined in the OTP-RR PR. https://github.com/opentripplanner/otp-react-redux/pull/1433/files

daniel-heppner-ibigroup:
Adding @alec-georgoff so he can take a look at how this PR accomplishes its goals. It's a good lesson in how confusing the data flow can be in OTP-RR...

packages/map-popup/src/util.ts
```diff
@@ -63,6 +66,14 @@ export function makeDefaultGetEntityName(
         },
         { name: stationName }
       );
+    } else if (
+      stopIdAgencyMap &&
+      entity.id in stopIdAgencyMap &&
+      "code" in entity
+    ) {
+      stationName = `${stationName} (${stopIdAgencyMap[entity.id].name} ${
```
binh-dam-ibigroup, line 74: Do we need to pass an entire map if the only important thing is the agency name of the stop? Can we just pass an agency id to the popup?

binh-dam-ibigroup (changes requested):
Add a story to illustrate the stop name rendering with agency name and stop code. Also, I think we could simply pass an agency id to the popup component, instead of passing an entire map.

packages/map-popup/src/__snapshots__/MapPopup.story.tsx.snap
```diff
@@ -385,3 +385,68 @@ exports[`Map Popup StopEntityWithEntityPrefix smoke-test 1`] = `
...
+<div class="styled__MapOverlayPopup-sc-12kjso7-1 cPqJxe">
+  <div id="focus-trimet3A9526-popup-focus-trap"
+       role="presentation"
+  >
+    <header class="styled__PopupTitle-sc-12kjso7-3 jRNaQh">
+      W Burnside &amp; SW 2nd (TriMet 9526)
+    </header>
+    <p class="styled__PopupRow-sc-12kjso7-2 ckOOWr">
+      <strong>
+        Stop ID: 9526
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: I think that's correct. The stop ID shows the stop code, which is correct. And the stop code exists in the title to match the geocoder output, so that's also correct. But maybe I don't understand your comment.

binh-dam-ibigroup (approved):
After internal dicussion, I'm going to approve this and ignore the duplication of the stop code in the popup title and stop code lines.

PR #858, opened 2025-06-30
Trip Details make header key optional on default fare type

daniel-heppner-ibigroup:
We don't need the header key here because it was just used on the old fare display which we don't really use anymore.


josh-willis-arcadis (approved):
LGTM!

amy-corson-ibigroup (approved):
Makes sense!

PR #860, opened 2025-07-02
feat(location-field): upgrade debouncer library

josh-willis-arcadis:
Upgrade to a new and improved debouncer library and restore `useCallback` on the `geocodeAutocomplete` function.


miles-grant-ibigroup (approved):
Huge diff has me nervous but everything seems to be working well

josh-willis-arcadis:
@daniel-heppner-ibigroup it does! I don't think we were debouncing at all with the previous solution. This library is much better!

PR #868, opened 2025-08-07
More OTP-UI new map releases

binh-dam-ibigroup:
This PR adds two breaking-change commits to trigger releases of the transit-vehicle-overlay and park-and-ride-overlay that were not updated by #848.


miles-grant-ibigroup (approved):
Exciting

PR #870, opened 2025-08-08
fix: remove duplicate stop code title

daniel-heppner-ibigroup:
This adjusts the way the feed name is rendered so you don't see duplicate info. Instead we move the feed name into the existing stop ID section, but use the parenthesis title in the location name for the from/to location picker. 
image


miles-grant-ibigroup (approved):
What happens if there is just one agency? I guess we can remove that with CSS?

amy-corson-ibigroup (approved):
Makes sense thank you!

PR #871, opened 2025-08-15
Options for transport mode function util function

alec-georgoff:
- Adds a util function to determine which mode button and/or mode settings need to be enabled for a given transport mode
- Adds unit tests for the new function
- Converts unit test file to TypeScript and updates snapshot

### DIFF
--- a/packages/trip-form/src/MetroModeSelector/utils.ts
+++ b/packages/trip-form/src/MetroModeSelector/utils.ts
@@ -212,7 +212,7 @@ export const findRequiredOptionsForTransportMode = (
   settings: ModeSetting[],
   transportMode: TransportMode
 ): RequiredOptionsForTransportMode => {
-  // If there's a mode button with the mode we need, then just return that. No mode setting necessary
+  // If there's a mode button with the mode we need, then just return that. No mode setting necessary!
   const modeButtonOnly = modeButtons.find(mb =>
     mb.modes?.some(
       m =>
--- a/packages/trip-form/src/MetroModeSelector/utils.ts
+++ b/packages/trip-form/src/MetroModeSelector/utils.ts
@@ -191,47 +191,63 @@ export const setModeButtonEnabled = (enabledKeys: string[]) => (
   };
 };
 
+/**
+ * Determines which modeButton (and potentially modeSetting) needs to be set in the query params
+ * for a certain TransportMode to be enabled. This is necessary because certain configurations
+ * have combinations of modeButtons and modeSettings that do not directly map to a given mode
+ * @param modeButtons Array of mode buttons from config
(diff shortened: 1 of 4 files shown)


packages/trip-form/src/MetroModeSelector/utils.ts
```diff
@@ -190,3 +190,64 @@ export const setModeButtonEnabled = (enabledKeys: string[]) => (
...
+export const findRequiredOptionsForTransportMode = (
+  modeButtons: ModeButtonDefinition[],
+  settings: ModeSetting[],
+  transportMode: TransportMode
+):
+  | {
+      modeSetting?: string;
+      modeButton: string;
+    }
+  | undefined => {
```
daniel-heppner-ibigroup, line 212: I'll leave this up to you to decide: what do you think is easier to handle as a consumer of this function:
- making both modeSetting and modeButton optional
- as you have now, making the entire object optional

I was thinking 1 would be cleaner here but maybe there are advantages to the consumer for the way you've done it

packages/trip-form/src/MetroModeSelector/utils.ts
```diff
@@ -190,3 +190,64 @@ export const setModeButtonEnabled = (enabledKeys: string[]) => (
...
+  // Otherwise, look for a mode setting with the mode button that we need
+  const modeSetting = settings.find(ms => {
+    if (!("addTransportMode" in ms)) return false;
+
+    return [ms.addTransportMode]
+      .flat() // addTransportMode can be a single TransportMode or an array of them
+      .some(
+        m =>
+          m.mode === transportMode.mode &&
+          m.qualifier === transportMode.qualifier
```
daniel-heppner-ibigroup, line 236: even though this check is duplicated I think it's better to just have that than extract it to a function. This is easier to read and see what's going on.

daniel-heppner-ibigroup: I'm saying this to fend off a second reviewer who suggests extracting this to a function

packages/trip-form/src/__tests__/mode-selector-utils.ts
```diff
@@ -274,4 +282,40 @@ describe("mode selector utils", () => {
...
+        qualifier: string | undefined,
+        expected: { modeSetting?: string; modeButton?: string } | undefined
+      ) => {
+        const result = findRequiredOptionsForTransportMode(
+          modeButtonDefinitions,
+          modeSettingDefinitions,
+          { mode, qualifier }
+        );
+        expect(result).toEqual(expected);
+      }
```
daniel-heppner-ibigroup, line 318: nice test!

daniel-heppner-ibigroup (approved):
Looks great

PR #872, opened 2025-08-27
`LegCalls` Based Flex Detection

miles-grant-ibigroup:
`isFlex` API is unchanged, but I am marking this as a breaking change because it significantly changes how the method behaves.

### DIFF
--- a/packages/core-utils/src/itinerary.ts
+++ b/packages/core-utils/src/itinerary.ts
@@ -98,20 +98,12 @@ export function legDropoffRequiresAdvanceBooking(leg: Leg): boolean {
  * returned by OTP when a leg is a flex leg.
  */
 export function isFlex(leg: Leg): boolean {
-  const callTypes =
-    leg?.stopCalls?.reduce((acc, cur) => {
-      // eslint-disable-next-line no-underscore-dangle
-      const callType = cur?.stopLocation?.__typename;
-      if (!acc.includes(callType)) {
-        acc.push(callType);
-      }
-
-      return acc;
-    }, []) || [];
   return (
-    // This isn't the most efficient way to do this but it's so clean and there will
-    // only ever be 3 items in the array
-    callTypes.includes("LocationGroup") || callTypes.includes("Location")
+    leg?.stopCalls?.some(call =>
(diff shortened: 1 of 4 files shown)


packages/types/src/index.ts
```diff
@@ -371,6 +371,12 @@ export type Leg = {
   serviceDate?: string;
   startTime: number | string;
   steps: Step[];
+  // Abridged version of what OTP returns
+  stopCalls?: {
+    stopLocation?: {
+      __typename?: string;
```
binh-dam-ibigroup, line 377: Shouldn't `__typename` be a list of allowed string values? What are there cases where `__typename` would be not set in the OTP response?

miles-grant-ibigroup: Good catch!

packages/core-utils/src/itinerary.ts
```diff
@@ -99,11 +99,11 @@ export function legDropoffRequiresAdvanceBooking(leg: Leg): boolean {
...
   return (
-    isReservationRequired(leg) ||
-    isCoordinationRequired(leg) ||
-    legDropoffRequiresAdvanceBooking(leg) ||
-    isAdvanceBookingRequired(leg?.pickupBookingInfo) ||
-    legContainsGeometry(leg)
+    leg?.stopCalls?.some(call =>
+      // Flex calls are "Location" or "LocationGroup"
+      // eslint-disable-next-line no-underscore-dangle
+      call?.stopLocation.__typename.includes("Location")
```
binh-dam-ibigroup, line 105: `stopLocation` is declared optional/nullable in the types but is assumed defined here. Should it be non-optional/non-nullable in the type declaration too?

binh-dam-ibigroup: Replace `.includes` with `.startsWith`.

binh-dam-ibigroup (commented):
Should `stopLocation` be a required field?

packages/core-utils/src/itinerary.ts
miles-grant-ibigroup: Great catches!

PR #873, opened 2025-08-29
Use type import syntax for RequiredOptionsForTransportMode

alec-georgoff:
Fixes an error caused by the incorrect import of `RequiredOptionsForTransportMode`


daniel-heppner-ibigroup (approved):
lgtm

PR #874, opened 2025-09-04
fix: show fares table even if no default fare defined

miles-grant-ibigroup:
Always shows fare table, even if there is no default fare type defined


binh-dam-ibigroup (approved):

daniel-heppner-ibigroup (approved):

daniel-heppner-ibigroup:
I merged this because I want to test with it today and time is limited!

PR #875, opened 2025-09-05
fix: Use type import syntax for RequiredOptionsForTransportMode

alec-georgoff:
Forgot to include a semantic release commit for #873 🙃


daniel-heppner-ibigroup (approved):
I hate semantic release

PR #876, opened 2025-09-10
fix: support flex `bookingUrl`

miles-grant-ibigroup:
This was supposed to be a larger change to add a border to the flex booking section (using the buliding blocks alert component), but that became a bit busy. For now, just adds support for rendering booking urls. The block is styled, so it should be easy to override.


amy-corson-ibigroup (approved):
Looks good! Sorry for the delay!

PR #877, opened 2025-09-15
fix(map-popup): render border only if stopID is provided

josh-willis-arcadis:
Remove border when no stopId is provided on a stop map-popup

### DIFF
--- a/packages/map-popup/src/styled.ts
+++ b/packages/map-popup/src/styled.ts
@@ -1,11 +1,11 @@
 import styled from "styled-components";
 
 /* eslint-disable-next-line import/prefer-default-export */
-export const ViewStopButton = styled.button<{ stopId?: string | boolean }>`
+export const ViewStopButton = styled.button<{ stopId?: string }>`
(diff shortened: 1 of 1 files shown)


miles-grant-ibigroup (approved):
Excited about this fix!

packages/map-popup/src/styled.ts
```diff
@@ -1,11 +1,11 @@
 import styled from "styled-components";
 
 /* eslint-disable-next-line import/prefer-default-export */
-export const ViewStopButton = styled.button<{ stopId?: string }>`
+export const ViewStopButton = styled.button<{ stopId?: string | boolean }>`
```
amy-corson-ibigroup, line 4: Wait I'm confused, if this is trying to account for no `stopId`, then couldn't it just be `stopId?: string` and the border check below is just checking if `stopId` exists? This feels a little convoluted as is

amy-corson-ibigroup (approved):
Looks good! Thanks for the changes!

PR #880, opened 2025-09-26
fix: Support array of default `mediumId` and `riderCategoryId`

miles-grant-ibigroup:
Previously we only supported one of each of these for the purpose of total fare calculation. With this PR, an array can be specified, allowing for different feeds to have different mediums and rider categories. It looks funky in the fare table, but there is no good way around that issue.

Intended to not be a popular feature. Only for those in the know. It doesn't create an excellent user experience.


amy-corson-ibigroup (approved):
Looks good!

PR #936, opened 2026-03-31
Tooltip Dynamic Placement

miles-grant-ibigroup:
Removing the `anchor` setting allows MapLibre to dynamically pick the direction the bubble goes. This allows it to dodge edges. 

Then, I've added some other fixes to make these popups less abrasive/buggy


binh-dam-ibigroup (approved):
Clever fix, but is the `popupProps.padding` attribute used in transit-vehicle-overlay having any effect?

alec-georgoff (approved):

PR #939, opened 2026-04-14
fix: add focus handlers for location field

alec-georgoff:
Adds focus handlers for the location field component, with the requisite types added to the component's props.


amy-corson-ibigroup (approved):
Looks good! 

I'm not sure this commit message is going to trigger a release, you may need to add a throwaway to make sure sem rel picks up on the change

josh-willis-arcadis (changes requested):
Def needs another commit to trigger a release

josh-willis-arcadis (approved):
I think the commit is correct to trigger a release. If not we can try again.

PR #940, opened 2026-04-21
fix(location-field): better handle input reflecting address state

amy-corson-ibigroup:
Previously, the only way to clear the `location` set by the location field was to use the clear button. This meant that occasionally the location field input `value` could be out of step with the `location` which got confusing. For example, if you cleared out the `value`, you could save the place, because the `location` had not technically been changed.

This PR clears the `location` if a user deletes the `value`. It also updates the input value `onBlur` to reflect the state of the `location`

Should this be a breaking change and should `clearLocation` be a required prop?

### DIFF
--- a/packages/location-field/src/index.tsx
+++ b/packages/location-field/src/index.tsx
@@ -570,9 +570,18 @@ const LocationField = ({
     }
   };
 
+  const onFieldBlur = () => {
+    onBlur && onBlur();
+    // Reset the input field to reflect the state onBlur
+    setValue(location?.name || "");
+  };
+
   const onTextInputChange = evt => {
     const { value } = evt.target;
     setValue(value);
+    /* If there's a location, but the user clears out the value in the input, 
+    clear the existing location */
+    location.lat && value === "" && clearLocation({ locationType });
     setMenuVisible(true);
 
     // Cancel all pending requests
@@ -1078,7 +1087,7 @@ const LocationField = ({
       className={formControlClassname}
       onChange={onTextInputChange}
       onClick={handleTextInputClick}
-      onBlur={onBlur}
+      onBlur={onFieldBlur}
       onFocus={onFocus}
       spellCheck={false}
       onKeyDown={onKeyDown}
--- a/packages/location-field/src/index.tsx
+++ b/packages/location-field/src/index.tsx
@@ -571,9 +571,9 @@ const LocationField = ({
   };
 
   const onFieldBlur = () => {
-    onBlur && onBlur();
     // Reset the input field to reflect the state onBlur
     setValue(location?.name || "");
+    onBlur && onBlur();
   };
 
   const onTextInputChange = evt => {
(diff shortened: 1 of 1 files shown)


packages/location-field/src/index.tsx
```diff
@@ -570,9 +570,18 @@ const LocationField = ({
...
+    // Reset the input field to reflect the state onBlur
+    setValue(location?.name || "");
+  };
+
   const onTextInputChange = evt => {
     const { value } = evt.target;
     setValue(value);
+    /* If there's a location, but the user clears out the value in the input, 
+    clear the existing location */
+    location.lat && value === "" && clearLocation({ locationType });
```
alec-georgoff: I think we may need to alter the existence check for `location` here...it seems like `location` could possibly be `null` (see [default value here](https://github.com/opentripplanner/otp-ui/blob/master/packages/location-field/src/index.tsx#L307) and [comment here](https://github.com/opentripplanner/otp-ui/blob/master/packages/location-field/src/index.tsx#L356)).

Could we simplify to just `location && value === "" [......]`?

If not, we should at least accommodate for the fact that `location.lat` will throw an error if `location` is null

alec-georgoff:
I'm not opposed to your suggestion of making `clearLocation` required...I think it's fair to expect a consumer of the location field to provide a callback for when location is cleared.

packages/location-field/src/index.tsx
amy-corson-ibigroup: Added optional chaining in [f350611](https://github.com/opentripplanner/otp-ui/pull/940/commits/f35061196b68e135d5211e64fa9f97245e00c8cd)! The issue with `location && value === "" ` is that location can exist but not have an actual address assigned, like in this case (a real value i was getting when i was testing):
```
location = {
   address: '',
   lat: null,
   lon: null
}

Comment (alec-georgoff) [reply packages/location-field/src/index.tsx]: Don't love that, but it's out of scope to fix something like that here. Thanks for the optional chaining!

**Review by alec-georgoff — APPROVED** [review]

File: packages/location-field/src/index.tsx [hunk 2]
```diff
@@ -570,9 +570,18 @@ const LocationField = ({
     }
   };
 
+  const onFieldBlur = () => {
+    onBlur && onBlur();
+    // Reset the input field to reflect the state onBlur
+    setValue(location?.name || "");
+  };
```
miles-grant-ibigroup: Should we reverse this? This way it's not a breaking change, right?

miles-grant-ibigroup (changes requested):
Nice fix! I'd just either mark this a breaking change or change the order of operations

packages/location-field/src/index.tsx
amy-corson-ibigroup: Good call thank you!

amy-corson-ibigroup:
@miles-grant-ibigroup do you have thoughts on making `clearLocation` a required prop?

miles-grant-ibigroup:
> @miles-grant-ibigroup do you have thoughts on making `clearLocation` a required prop?

I think I'm ok to keep it as optional. It technically works... and I can imagine some simple uses of the component that don't require having it

miles-grant-ibigroup (approved):

PR #941, opened 2026-04-28
Basic infrastructure to prepare for strict mode updates

daniel-heppner-ibigroup:
This PR is mostly focused on removing hoisted modules, which surfaced a bunch of dependencies that were missing in packages around the app. I went through and added those dependencies to the specific packages where they're required. 

This was a necessary precursor to enabling strict mode since we need to more tightly control the dependencies of each package to get the types correct.


packages/location-field/src/stories/Mobile.story.tsx
```diff
@@ -13,7 +13,7 @@ import {
   SkullCrossbones,
   Star,
   Train,
-} from "styled-icons/fa-solid";
+} from "@styled-icons/fa-solid";
```
miles-grant-ibigroup, line 16: How did this ever work lol

packages/trip-viewer-overlay/package.json
```diff
@@ -27,5 +27,8 @@
     "react": "catalog:",
     "react-map-gl": "catalog:"
   },
-  "gitHead": "0af1b7cda60bd4252b219dcf893e01c2acb2ed5d"
+  "gitHead": "0af1b7cda60bd4252b219dcf893e01c2acb2ed5d",
```
miles-grant-ibigroup, line 30: What is this even for

miles-grant-ibigroup (approved):
Unhoisted! Wow we are so on the ground

packages/trip-viewer-overlay/package.json
daniel-heppner-ibigroup: apparently this is something lerna uses... idk if we need it but I'm just going to leave it for now

alec-georgoff (approved):

PR #943, opened 2026-04-28
fix: use workspace:^ instead of workspace:* for internal deps

daniel-heppner-ibigroup:
This changes all internal dependencies to use workspace:^ instead of *, which will allow them more flexibility in the version number they use for better deduplication.


miles-grant-ibigroup (approved):
So clean!

alec-georgoff (approved):

PR #944, opened 2026-05-04
fix(offline gecoder): support unicode

miles-grant-ibigroup:
Latinizes both search strings and data. This is because uFuzzy is much faster with latin characters only.

### DIFF
--- a/packages/geocoder/src/geocoders/types.ts
+++ b/packages/geocoder/src/geocoders/types.ts
@@ -27,6 +27,7 @@ export type GeocoderConfig = {
   sources?: string;
   size?: number;
   reverseUseFeatureCollection?: boolean;
+  enableSlowFullUnicodeSupport?: boolean
 };
 
 export type ReverseQuery = {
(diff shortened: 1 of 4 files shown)


daniel-heppner-ibigroup (changes requested):
I think this still won't work with chinese and arabic (and other non-roman based character systems) right? I know we have some synonyms that include those languages. Maybe we can have a flag to enable ufuzzy's unicode support at the cost of performance? And add some tests that include chinese

miles-grant-ibigroup:
Yeah we'll probably have to add a flag for unicode support... grr

packages/geocoder/src/geocoders/types.ts
```diff
@@ -27,6 +27,7 @@ export type GeocoderConfig = {
   sources?: string;
   size?: number;
   reverseUseFeatureCollection?: boolean;
+  enableSlowFullUnicodeSupport?: boolean
```
daniel-heppner-ibigroup: run the formatter

daniel-heppner-ibigroup (approved):

amy-corson-ibigroup (approved):
LGTM

PR #946, opened 2026-05-14
Remove pre/post scripts

binh-dam-ibigroup:
This PR restricts scripts that can run at install/run/publish time and limits which packages and versions can run scripts.

Note: only-allow was deprecated as of April 22, 2026 and has been replaced with a `devEngines` entry in package.json. pnpm version 11 is now required for development.


daniel-heppner-ibigroup (approved):
this is great! Thanks for modernizing things

alec-georgoff (approved):

PR #947, opened 2026-05-19
feat(itinerary-body): Add leg prop to PlaceName

grant-humphries:
Add `leg` as a required prop to the `PlaceName` component, which is accepted as a prop by `ItineraryBody`

Including `leg` provides additional context about the place being rendered and enables more flexibility in how `PlaceName` determines its output

For example, when an itinerary includes a same-stop transfer, an identical value is passed to `PlaceName` under its `place` prop for the transfer walking leg and the subsequent transit leg. Passing the `leg` provides a way to differentiate what should be displayed for each case

### DIFF
--- a/packages/itinerary-body/src/defaults/place-name.tsx
+++ b/packages/itinerary-body/src/defaults/place-name.tsx
@@ -10,11 +10,17 @@ import BasicPlaceName from "./basic-place-name";
 export default function PlaceName({
   config,
   interline,
+  leg,
   place
 }: PlaceNameProps): ReactElement {
   return (
     <>
-      <BasicPlaceName config={config} interline={interline} place={place} />
+      <BasicPlaceName
+        config={config}
+        interline={interline}
+        leg={leg}
+        place={place}
+      />
 
       {/* TODO: take another pass on this when working the Transit Leg */}
       {/* Place subheading: Transit stop */}
--- a/packages/itinerary-body/src/ItineraryBody/place-row.tsx
+++ b/packages/itinerary-body/src/ItineraryBody/place-row.tsx
@@ -26,7 +26,7 @@ function getLegPlaceName(
   const interline = !!(!isDestination && leg.interlineWithPreviousLeg);
   const place = isDestination ? { ...leg.to } : { ...leg.from };
   const placeName = (
-    <PlaceName config={config} interline={interline} place={place} />
+    <PlaceName config={config} interline={interline} leg={leg} place={place} />
   );
 
   return {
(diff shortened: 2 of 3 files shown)


packages/itinerary-body/src/defaults/place-name.tsx
```diff
@@ -10,11 +10,17 @@ import BasicPlaceName from "./basic-place-name";
 export default function PlaceName({
   config,
   interline,
-  place
+  place,
+  leg
```
binh-dam-ibigroup: Nit - Sort props where you added that new `leg` prop.

binh-dam-ibigroup (approved):

packages/itinerary-body/src/defaults/place-name.tsx
grant-humphries: sorted ✅

daniel-heppner-ibigroup (commented):
Is it possible to derive the `interline` variable from the `leg`? And for an interlined leg's joining stop, is it the preceding or following leg that you're expecting? Might be good to add some comments to clarify that.

miles-grant-ibigroup (changes requested):
Clean change! I agree with @daniel-heppner-ibigroup  though. Is there no way to isolate only the parts of `leg` that you need?

grant-humphries:
@daniel-heppner-ibigroup 

> Is it possible to derive the interline variable from the leg? 

Based on the logic below, it doesn’t look like we can derive this purely from the `leg`. Whether a place is considered an interline depends in part on whether it’s a “destination” (i.e., the final stop in the itinerary), and that context isn’t available directly on the leg object:

Screenshot 2026-05-26 100408

That said, I haven’t fully traced where `isDestination` is coming from yet. If it can be derived from leg, were you thinking we should remove the `interline` prop? I’ve been trying to avoid introducing breaking changes so far

> And for an interlined leg's joining stop, is it the preceding or following leg that you're expecting? 

From what I’m seeing, an interlining stop still has its own itinerary leg. For example, travel from the boarding stop to the interlining stop is one leg, and from the interlining stop to the alighting stop is another. With the current changes, the interlining `PlaceName` instance receives the latter leg. Does that align with your understanding?

Happy to add comments for clarity, but I just need a little more guidance on what you’d like called out

grant-humphries:
@miles-grant-ibigroup

> Is there no way to isolate only the parts of leg that you need?

We're using `PlaceName` in an unconventional way and our pending design uses the
following attributes from `leg`:

- `agencyName`
- `duration`
- `distance`
- `startTime`
- `mode`
- `to`
- `from`
- `transitLeg`

Given that, passing the full leg object seems cleaner than splitting these into
multiple props, though some could be coalesced down to a single value (e.g., we
use `to`, `from` & `transitLeg` to determine same-stop transfers).

Could you share more about your concerns with passing the full `leg`? From a
rendering perspective, `PlaceName` sits near the leaf nodes of the DOM tree, and
its parent (`PlaceRow`) already receives `leg`, so this shouldn’t introduce
additional rendering overhead

miles-grant-ibigroup (approved):
That's a reasonable explanation, thanks!

PR #948, opened 2026-05-21
feat - Add aria-disabled to location field

nikkisato:
Specifically on Mac OS Safari 26.3 w/ VoiceOver and IOS Devices on 26.3 with Voiceover 

we are getting an issue with the Location Field Icon itself 

```When there are no search suggestions, the toggle element is not identified as dimmed or unavailable.```

image

752a5288-a3dc-44dc-895d-01ecc84078ca

By adding an aria-disabled this would help 

Also fixing the Package.json file script for update-snapshots and test-snapshots they were using npx while the Github CI was using pnpx so was asked if i could make the change

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,8 +124,8 @@
     "typescript": "lerna run tsc",
     "unit": "jest",
     "test-storybook": "test-storybook --url http://localhost:5555",
-    "update-snapshots": "pnpm build-storybook; npx concurrently -k -s first -n \"SB,TEST\" \"npx http-server storybook-static --port 5555 --silent\" \"npx wait-on tcp:5555 && pnpm test-storybook --url http://localhost:5555 -u\"",
-    "test-snapshots": "npx concurrently -k -s first -n \"SB,TEST\" \"npx http-server storybook-static --port 5555 --silent\" \"npx wait-on tcp:5555 && pnpm test-storybook --url http://localhost:5555\"",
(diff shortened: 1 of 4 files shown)


nikkisato:
This would be an accessibility fix so @amy-corson Tagging you to hopefully get this reviewed! 

@miles-grant-ibigroup As the other reviewer

nikkisato:
Please also let me know If I need to do a specific word for the commit message to kick off something in the background too

miles-grant-ibigroup:
> Please also let me know If I need to do a specific word for the commit message to kick off something in the background too

Oh good catch thanks! You'll probably need to add a `fix: ` commit to trigger a new release!

daniel-heppner-ibigroup (approved):

amy-corson-ibigroup:
Hi @nikkisato thanks for the changes! Sorry for the delay in getting back to you, just for future reference, amy-corson is technically me but I don't really use it!

So just to confirm, you're thinking if there's no active search in the location field, this button should be `aria-disabled`? I'm asking because the condition you're using to check in this PR is `menuItemCount === 0` but I'm testing in storybook and I'm not seeing a case where this is true. If there's no active search, there's always the "use current location" button or the "user has disabled geolocation" notice. So `menuItemCount` pretty much always is at least `1`. Is that okay?

image

nikkisato:
So this was a case where if the user or developer doesn't have current location on 

https://trimet.org/home/search/

https://github.com/user-attachments/assets/b93035bd-5ae9-42eb-b021-a1388bff52f9

Attached a screenrecording

nikkisato:
@miles-grant-ibigroup can you assign Amy corson please thank you!

amy-corson-ibigroup (approved):
Thanks for the clarification! This makes sense, I appreciate the change

nikkisato:
@miles-grant-ibigroup This looks ready to be merged!

PR #949, opened 2026-05-28
fix(core-utils): Fix the bug in walk reluctance

amy-corson-ibigroup:
We shouldn't need a `falseValue` for the walk reluctance checkbox, it should just use the routing engine default.


miles-grant-ibigroup (approved):
Clean!

alec-georgoff (approved):

PR #950, opened 2026-05-29
Create timetable package

alec-georgoff:
This is the initial code for a new `timetable` package in OTP-UI. Most of the styling and accessibility work is not yet completed; the goal here is to get a "pre-release" package published for consumption in OTP-RR. Additional improvements will continue to be made over time.


packages/timetable/src/index.tsx
```diff
@@ -0,0 +1,326 @@
...
+  align-items: center;
+  padding: 1rem;
+  width: ${COLUMN_WIDTH};
+`;
+
+const RowContainer = styled.div`
+  display: flex;
+`;
+
+const TimeTableRow = (props: TimeTableRowProps): ReactElement => {
```
alec-georgoff: I will work on converting this to proper semantic HTML `table` elements in parallel with this review!

packages/timetable/src/index.tsx
```diff
@@ -0,0 +1,326 @@
...
+  name: string;
+  tripsForDate: Trip[];
+}
+
+interface Trip {
+  blockId: string;
+  stoptimesForDate: Stoptime[];
+}
+
+interface Stoptime {
```
alec-georgoff, line 67: The eventual goal is to use this as an opportunity to improve our type handling (potentially via auto-generated types and graphQL fragments) but these types are just a stopgap to get this component moving until then.

packages/timetable/src/index.tsx
amy-corson-ibigroup, line 1: Can you sort (alphabetize) the css in this file

packages/timetable/src/stories/Timetable.story.tsx
amy-corson-ibigroup, line 1: There are "controls" in storybook that already allow you to toggle blockId and dwell stops, so honestly I think you only need one story here. I'm getting everything with just the default story.

packages/timetable/src/index.tsx
```diff
@@ -0,0 +1,336 @@
+import React, { ChangeEvent, ReactElement, useMemo, useState } from "react";
+import styled from "styled-components";
+import toposort from "toposort";
+
+const COLUMN_WIDTH = "85px";
```
amy-corson-ibigroup, line 6: Pretty sure that using a `table` will remove the need for this! If it doesn't, I think it's okay to inline it

packages/timetable/src/index.tsx
```diff
@@ -0,0 +1,336 @@
...
+        const departureStopTime: Stoptime = {
+          ...st,
+          scheduledArrival: st.scheduledDeparture,
+          stop: {
+            ...st.stop,
+            gtfsId: dwellStopId
+          }
+        };
+        updatedStopTimes.push(arrivalStopTime);
+        updatedStopTimes.push(departureStopTime);
```
amy-corson-ibigroup:
```suggestion
updatedStopTimes.push(arrivalStopTime, departureStopTime);
```

amy-corson-ibigroup (commented):
Still working my way through this sorting logic, but in the interest of time, first round of comments

packages/timetable/src/index.tsx
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    const patternStopsFromSorted: PatternStop[] = sorted.map(stopId => {
+      return {
+        id: stopId,
+        name: stopIdToNameMap.get(stopId) ?? ""
+      };
+    });
+
+    return { patternStops: patternStopsFromSorted, tripStopSets: sets };
```
amy-corson-ibigroup, line 254: maybe I'm wrong, but establishing and renaming these variables was confusing to me. I'd rather `patternStops` and `tripStopSets` be named from the jump, but I'll defer to 2nd reviewer

packages/timetable/src/index.tsx
```diff
@@ -0,0 +1,336 @@
...
+    <>
+      <button type="button" onClick={() => setExpanded(!expanded)}>
+        {expanded ? "Show Timepoints Only" : "Show All Stops"}
+      </button>
+      <select id="direction-select" onChange={handleDirectionSelection}>
+        <option value={0}>One Direction</option>
+        <option value={1}>Other Direction</option>
+      </select>
+      <RowContainer>
+        {showBlockId ? <CellContainer>Block ID</CellContainer> : <div />}
```
amy-corson-ibigroup: The font weight here should mirror the other table column headers

amy-corson-ibigroup (approved):
Approving because I don't see anything blocking! But I will keep an eye on this PR as you add the `<table>`, test with different GTFS data, etc.

I've sent you a few clarifying questions, but I think those are more to develop my understanding of some of these sorting methods than actual issues with the code.

Thanks for the changes!

packages/timetable/src/index.tsx
alec-georgoff: Done!

packages/timetable/src/index.tsx
alec-georgoff: I updated to `table` but still had to end up using this just to make things look consistent. I'd like to keep it as a const since it's used in a couple of places.

packages/timetable/src/index.tsx
```diff
@@ -0,0 +1,335 @@
...
+            const rowValues: string[] = showBlockId ? [t.blockId] : [];
+            filteredPatternStops.forEach(patternStop => {
+              const stopDetail = t.stops.get(patternStop.id);
+              rowValues.push(
+                stopDetail
+                  ? stopDetail.time.toLocaleTimeString("en-us", {
+                      timeZone,
+                      hour12: false,
+                      hour: "2-digit",
+                      minute: "2-digit" // use react intl for formatting
```
alec-georgoff: Will update this to use the `Intl` approach currently used elsewhere

packages/timetable/src/stories/Timetable.story.tsx
alec-georgoff: Oh cool, I didn't notice those! Removed the extra stories.

packages/timetable/src/index.tsx
alec-georgoff: This is an annoying TypeScript thing, I think. I tried just using `patternStops` and `tripStopSets` initially but then the linter complained that those variables were already declared in the upper scope. If there's an easier way around this I'd be happy to implement!

binh-dam-ibigroup:
There is a file with conflicts, you'll have to rerun `pnpm install`.

packages/timetable/src/index.tsx
```diff
@@ -0,0 +1,329 @@
...
+import { useIntl } from "react-intl";
+import styled from "styled-components";
+import toposort from "toposort";
+
+const COLUMN_WIDTH = "85px";
+
+const Table = styled.table`
+  display: block;
+  overflow: auto;
+  table-layout: fixed;
```
binh-dam-ibigroup: Remove this attribute, it has no effect.

Image

packages/timetable/src/index.tsx
```diff
@@ -0,0 +1,329 @@
...
+
+const TH = styled.th`
+  min-width: ${COLUMN_WIDTH};
+  padding: 1rem;
+`;
+
+const TR = styled.tr`
+  text-align: center;
+  vertical-align: middle;
+  width: ${COLUMN_WIDTH};
```
binh-dam-ibigroup: Remove `width`, it does not apply to `<tr>` elements, I think.

packages/timetable/src/index.tsx
```diff
@@ -0,0 +1,329 @@
...
+`;
+
+const TD = styled.td`
+  padding: 1rem 0;
```
binh-dam-ibigroup: Nit - Inconsistent padding with `TH`

packages/timetable/src/index.tsx
```diff
@@ -0,0 +1,329 @@
...
+`;
+
+export interface PatternStop {
```
binh-dam-ibigroup: I think only `TimetableProps` needs to be exported.

packages/timetable/src/index.tsx
```diff
@@ -0,0 +1,329 @@
...
+export interface TimeTableProps {
+  route: Route;
+  includeDwellStops?: boolean;
+  showBlockId?: boolean;
+  errorOnStopSorting?: boolean;
+  timeZone?: string;
+}
+
+const TimeTable = (props: TimeTableProps): ReactElement => {
+  const { route, includeDwellStops, showBlockId, errorOnStopSorting } = props;
```
binh-dam-ibigroup: Sort props

packages/timetable/src/index.tsx
```diff
@@ -0,0 +1,329 @@
...
+
+  const { patterns } = route;
+
+  const intl = useIntl();
+
+  const [directionId, setDirectionId] = useState<number>(0);
```
binh-dam-ibigroup: Convert this into a prop.

packages/timetable/src/index.tsx
```diff
@@ -0,0 +1,329 @@
...
+  const [expanded, setExpanded] = useState(false);
```
binh-dam-ibigroup: Convert this into a prop.

packages/timetable/src/index.tsx
```diff
@@ -0,0 +1,329 @@
...
+  };
+
+  return (
+    <>
+      <button type="button" onClick={() => setExpanded(!expanded)}>
+        {expanded ? "Show Timepoints Only" : "Show All Stops"}
+      </button>
+      <select id="direction-select" onChange={handleDirectionSelection}>
+        <option value={0}>One Direction</option>
+        <option value={1}>Other Direction</option>
```
binh-dam-ibigroup: I'd recommend turning these into props for timepoints only/all stops, and a direction id prop. It gives implementers more flexibility.

packages/timetable/src/index.tsx
```diff
@@ -0,0 +1,329 @@
...
+
+  // Used to extract stop names more efficiently later
+  const stopIdToNameMap = new Map<string, string>();
+
+  const [allTrips, timepointStopIds] = useMemo(() => {
+    const trips = patterns
+      .filter(p => p.directionId === directionId)
+      .flatMap(p => p.tripsForDate);
+
+    const timepoints = new Set<string>();
```
binh-dam-ibigroup: Extract a method for computing timepoints.

packages/timetable/src/index.tsx
```diff
@@ -0,0 +1,329 @@
...
+        patterns
+          .filter(p => p.directionId === directionId)
+          .forEach(p => {
+            p.tripsForDate.forEach(trip =>
+              trip.stoptimesForDate.forEach(st => {
+                naiveStops.add(st.stop.gtfsId);
+              })
+            );
+          });
+        sorted = [...naiveStops];
```
binh-dam-ibigroup: Extract a method for the naive approach.

packages/timetable/src/index.tsx
```diff
@@ -0,0 +1,329 @@
...
+        stopIdToNameMap.set(st.stop.gtfsId, st.stop.name);
+        stopIds.push(st.stop.gtfsId);
+        set.add(st.stop.gtfsId);
+      });
+      sets.push(set);
+      stopIds.forEach((stopId, index) => {
+        if (index !== stopIds.length - 1)
+          stopGraph.push([stopId, stopIds[index + 1]]);
+      });
+    });
```
binh-dam-ibigroup: Extract a method to generate the graph.

packages/timetable/src/index.tsx
```diff
@@ -0,0 +1,329 @@
...
+          stop: {
+            ...st.stop,
+            gtfsId: dwellStopId
+          }
+        };
+        updatedStopTimes.push(arrivalStopTime, departureStopTime);
+        if (st.timepoint) timepoints.add(dwellStopId);
+      });
+      withDwellStops.push({
+        blockId: trip.blockId,
```
binh-dam-ibigroup: Extract a method to compute the trips including dwell stops.

binh-dam-ibigroup (changes requested):
Nice initial implementation. I would recommend replacing the expand button and the direction with props, and for our future sake, splitting the long methods.

packages/timetable/src/index.tsx
alec-georgoff: Actually, I think we can just use the default export for `Timetable`; the props don't need to be exported anywhere (unless I'm misunderstanding how we handle props for a package)

packages/timetable/src/index.tsx
alec-georgoff: [Done here](https://github.com/opentripplanner/otp-ui/pull/950/changes/36ac945e7c6f06a8ae3affa0625d2f8a06c59389)

packages/timetable/src/index.tsx
alec-georgoff: [Done here](https://github.com/opentripplanner/otp-ui/pull/950/changes/2b7fea79d034869f6a04f30edf611b8de198ea3d)

alec-georgoff:
Addressed most of the smaller suggestions [here](https://github.com/opentripplanner/otp-ui/pull/950/changes/4960ca1fc63a7d26f380d754f79d87dd4bca42ff)

pnpm-lock.yaml
```diff
@@ -15948,6 +15967,14 @@ snapshots:
 
   es6-promise@4.2.8: {}
 
+  esbuild-register@3.6.0(esbuild@0.25.5):
```
alec-georgoff: @binh-dam-ibigroup do you know where this addition comes from? This was included in the merge conflict

pnpm-lock.yaml
```diff
@@ -17912,6 +17939,32 @@ snapshots:
 
   jest-serializer@24.9.0: {}
 
+  jest-snapshot@30.0.4:
```
alec-georgoff: @binh-dam-ibigroup do you know where this addition comes from? This was included in the merge conflict

binh-dam-ibigroup: How did you resolve the conflict? If you accepted an older commit from master that how it might have ended up there. I would remove jest-snapshot@30.0.4.

pnpm-lock.yaml
binh-dam-ibigroup: Same thing here, I don't know how you resolved conflicts. Prob safe to remove.

packages/timetable/src/stories/Timetable.story.tsx
```diff
@@ -1,7 +1,7 @@
 // eslint-disable-next-line prettier/prettier
 import type { Meta, StoryObj } from "@storybook/react-vite"
 
-import { TimeTable } from ".."
+import TimeTable from ".."
```
binh-dam-ibigroup, line 4: Yes!

packages/timetable/src/index.tsx
```diff
@@ -268,21 +311,8 @@ const TimeTable = (props: TimeTableProps): ReactElement => {
...
   }, [allTrips, comparator]);
 
-  const handleDirectionSelection = (e: ChangeEvent<HTMLSelectElement>) => {
-    const val = e.target.value;
-    const numVal = Number(val);
-    setDirectionId(numVal);
-  };
-
   return (
     <>
```
binh-dam-ibigroup: You can remove the surrounding fragment tag, it is no longer needed.

packages/timetable/src/stories/Timetable.story.tsx
```diff
@@ -15,6 +15,8 @@ type Story = StoryObj<typeof meta>;
 
 export const Default: Story = {
     args: {
-        route: twinCitiesRouteMock.data.route
+        directionId: 0,
```
binh-dam-ibigroup, line 18: Add an entry for the prop `includeDwellStops`.

packages/timetable/src/index.tsx
binh-dam-ibigroup: Instead of hard-coding the format, let react-intl pick up the ambient time settings, and 
`intl.formatTime(stopDetail.time)` should be enough.

binh-dam-ibigroup (approved):
One tweak to the time rendering and one tweak to the story, and it should be good to go.

packages/timetable/src/stories/Timetable.story.tsx
alec-georgoff: [Amy pointed out](https://github.com/opentripplanner/otp-ui/pull/950#discussion_r3344327318) that we can just rely on the controls within Storybook for altering props! Reduces the amount of story code we have to write.

packages/timetable/src/index.tsx
alec-georgoff: I'm good with doing that for now. I think potentially a future addition could be a prop that handles time formatting...I know that some agencies use bold to denote PM times, while others use 24 hour time, and still others use AM and PM. Might be useful to let consumers have some flexibility here.

pnpm-lock.yaml
alec-georgoff: This and the other change were both from my branch, not from master...so not really sure how they got introduced. I will remove them

packages/timetable/src/index.tsx
```diff
@@ -184,9 +184,11 @@ interface TimeTableProps {
    * values for `arrival_time` and `departure_time`
    */
   includeDwellStops?: boolean;
+  /** A react-intl object to use for time formatting */
+  intl?: IntlShape;
```
binh-dam-ibigroup, line 188: You don't need to set intl as a prop. Intl is context based, so the implementer provides that.

PR #951, opened 2026-06-01
Fix Pelias Options

binh-dam-ibigroup:
This PR forwards the options (specifically, custom HTTP headers) defined in config to Pelias. I suppose that feature broke when we added the abort signal.
Should we attempt to refactor the autocomplete and search functions? (e.g. extract the common body)?


josh-willis-arcadis (approved):
LGTM

amy-corson-ibigroup (approved):

PR #952, opened 2026-06-02
PRJ77-1791: Add dynamic layer that activates to display custom itinerary map-line outlines

lightsage88:
This pull-request sets up `TransitiveOverlay` so that if the passed in `transitiveData` includes within its `route` property's elements, a property called `route_outline_contrast_color`, then this property will be aliased as `contrastColor`. The presence of `contrastColor` will cause a new `<Layer />` instance to be displayed in packages/transitive-overlay/src/index.tsx which will provide an outline to the given itinerary line on the map.


miles-grant-ibigroup (commented):
Hm we might want to go back to the drawing board on this one... I'm not sure I am comfortable with adding a proprietary route attribute. Is it really impossible for you to create a method inside of transitive-overlay/core-utils that generates the stroke color based on the route color?

lightsage88:
Hey miles, I really do think it's impossible yes. Mainly because to get what I need otherwise, I would have to either:
A. Make a request to a trimet endpoint from within otp-ui
B. Send over an even larger proprietary list within a prop which you'll want even less

miles-grant-ibigroup:
I know you are currently storing the colors on the trimet site, but is it possible for you to generate those colors within otp-ui is what I'm asking

lightsage88:
We would need to add some constants that catch the most contrastable color of our map's background lines given the way our map is set up now. They would have to be:
* `LIGHT_BACKGROUND` = `#dbd9d8`
* `DARK_BACKGROUND` = `#1a364e`

Presently the TransitiveOverlay gets the `route_color` within the `transitiveData` prop that is passed in.

In the TriMet application, we:
* Make a call to the API for all routeColors
* Isolate the unique ones
* We run `calculateContrastColors`
* `calculateContrastColors` uses utilities from the following dependencies: `@ctrl/tinycolor` and `wcag-contrast-utils` in order to compare the unique colors against our `LIGHT_BACKGROUND` and `DARK_BACKGROUND` colors

Now we _do_, currently without any changes, already get the `route_color` within the transitiveData argument that is passed into `TransitiveOverlay`. To avoid passing in a new property within transitiveData, `route_outline_contrast_color`, we could create a facsimile of `calculateContrastColors` with the aforementioned dependencies added into the mix (`@ctrl/tinycolor` and `wcag-contrast-utils`). This way, we can run these processes within otp-ui, and not have to add anything new to the `transitiveData` prop being passed into TransitiveOverlay.

packages/transitive-overlay/src/index.tsx
```diff
@@ -177,6 +179,32 @@ const loadImages = (map: MapRef, images: MapImage[]) => {
...
+export const SELECTED_OUTLINE_WIDTH_MULTIPLIER = 1.4;
+
+/** Shared line-width for focused route contrast outlines */
+export const FOCUSED_ROUTE_OUTLINE_WIDTH: ExpressionSpecification = [
+  "interpolate",
+  ["linear"],
+  ["zoom"],
+  ...ROUTE_LINE_WIDTH_BY_ZOOM.flatMap(([zoom, width]) => [
+    zoom,
+    width * SELECTED_OUTLINE_WIDTH_MULTIPLIER,
```
binh-dam-ibigroup: Could you provide a story, please, so we can see what these parameters do?

lightsage88: You mean like a storybook thing, right?

binh-dam-ibigroup: Yes, please

lightsage88: [screen-capture - 2026-06-09T150104.389.webm](https://github.com/user-attachments/assets/140e4f9c-70a1-41af-a55a-f1deca55744e)

miles-grant-ibigroup:
> Now we _do_, currently without any changes, already get the `route_color` within the transitiveData argument that is passed into `TransitiveOverlay`. To avoid passing in a new property within transitiveData, `route_outline_contrast_color`, we could create a facsimile of `calculateContrastColors` with the aforementioned dependencies added into the mix (`@ctrl/tinycolor` and `wcag-contrast-utils`). This way, we can run these processes within otp-ui, and not have to add anything new to the `transitiveData` prop being passed into TransitiveOverlay.

Could you give this a try? If you put the new method in core-utils I think this would be a great addition

lightsage88:
I have just pushed up something that affects the transitive-overlay and also updates the core-utils (which transitive-overlay pulls from) so that we get the contrast colors and line with we seek on itinerary lines.

packages/core-utils/src/contrast-colors.ts
```diff
@@ -0,0 +1,110 @@
...
+ * - Expand 3-digit forms
+ * - Validate allowed chars
+ * Returns normalized '#rrggbb' or null if invalid.
+ */
+const normalizeHex = (input: string | null | undefined): string | null => {
+  if (!input) return null;
+  const color = new TinyColor(input);
+  return color.isValid ? color.toHexString().toLowerCase() : null;
+};
+
```
miles-grant-ibigroup: Do we need this method? Can we inline this

packages/core-utils/src/contrast-colors.ts
```diff
@@ -0,0 +1,110 @@
...
+ * Calculates WCAG 2.1 contrast ratio between two hex colors.
+ * Defensive try/catch: return 0 if library throws.
+ */
+const getContrast = (foreground: string, background: string): number => {
+  try {
+    return contrastRatio(foreground, background);
+  } catch {
+    return 0;
+  }
+};
```
miles-grant-ibigroup: Do we need try/catch for this? Does `|| 0` not work?

packages/core-utils/src/index.ts
```diff
@@ -27,4 +29,5 @@ const core = {
 };
 
 export { SafeSuspense };
+export { default as calculateContrastColors } from "./contrast-colors";
```
miles-grant-ibigroup: Can you avoid `default` and instead just export the method indivdually?

packages/core-utils/src/colors.ts
```diff
@@ -0,0 +1,110 @@
...
+  hexColor: string
+): {
+  light: string | null;
+  dark: string | null;
+} => {
+  return {
+    light: adjustColorForContrast(hexColor, LIGHT_BACKGROUND),
+    dark: adjustColorForContrast(hexColor, DARK_BACKGROUND)
+  };
+};
```
miles-grant-ibigroup, line 49: Can this functionality be merged into `adjustColorForContrast`?

packages/transitive-overlay/src/index.tsx
```diff
@@ -185,7 +213,9 @@ const TransitiveCanvasOverlay = ({
   transitiveData
 }: Props): JSX.Element => {
   const { current: map } = useMap();
-
+  const mapStyleName = map?.getStyle()?.name;
+  const isDark = mapStyleName === "TriMet Dark";
+  const isLight = mapStyleName === "TriMet";
```
miles-grant-ibigroup: We can't merge this into OSS. Is there some other flag on the style object you can use?

packages/transitive-overlay/src/index.tsx
```diff
@@ -353,11 +383,19 @@ const TransitiveCanvasOverlay = ({
...
-          )
+          ...(transitiveData.patterns || []).flatMap((pattern: TransitivePattern) => {
+            const feature = patternToRouteFeature(pattern, transitiveData.routes);
+            const routeColor = feature.properties.color as string;
+            let contrastColor: string | null = null;
+            if (isDark) {
+              contrastColor = calculateContrastColors(routeColor).dark;
+            } else if (isLight) {
+              contrastColor = calculateContrastColors(routeColor).light;
+            }
```
miles-grant-ibigroup: Instead of 2 variables, can you instead just have one `isDark` variable that you pass into `calculateContrastColors`? That way you can avoid having the calculate the incorrect one at all

packages/transitive-overlay/src/util.ts
```diff
@@ -516,5 +519,5 @@ const drawArc = (straight: any) => {
   ).geometry;
 };
 
-export { drawArc };
+export { drawArc, calculateContrastColors };
```
miles-grant-ibigroup: Why is this being exported?

packages/core-utils/package.json
```diff
@@ -13,6 +13,7 @@
   "private": false,
   "dependencies": {
     "@conveyal/lonlat": "^1.4.1",
+    "@ctrl/tinycolor": "4.2.0",
```
miles-grant-ibigroup: We already use chroma, which has the same features that you are using tinycolor for.  Can you use that instead?

packages/core-utils/src/contrast-colors.ts
```diff
@@ -0,0 +1,110 @@
...
+  threshold: number = WCAG_AA_NON_TEXT_CONTRAST
+): string | null => {
+  const fullHex = normalizeHex(hexColor);
+  const bgHex = normalizeHex(backgroundColor);
+  if (!fullHex || !bgHex) return null;
+
+  const initialContrast = getContrast(fullHex, bgHex);
+  if (initialContrast >= threshold) {
+    return null;
+  }
```
miles-grant-ibigroup: Can inline this

packages/core-utils/src/contrast-colors.ts
```diff
@@ -0,0 +1,110 @@
...
+ * Returns:
+ * - null if original color already meets contrast
+ * - adjusted hex if modification needed
+ * - fallback black/white if conversion fails
+ *
+ * Strategy:
+ * - Decide direction (darken vs lighten) based on background brightness
+ * - Step lightness using tinycolor lighten/darken until contrast passes or limit reached
+ */
+const adjustColorForContrast = (
```
miles-grant-ibigroup, line 26: What if instead of ajdusting for contrast this method instead always returns a safe value?

packages/core-utils/src/contrast-colors.ts
```diff
@@ -0,0 +1,110 @@
...
+    iterations += 1;
+    working = isLightBackground
+      ? working.darken(LIGHTNESS_STEP_PERCENT)
+      : working.lighten(LIGHTNESS_STEP_PERCENT);
+
+    const adjustedHex = working.toHexString().toLowerCase();
+    const ratio = getContrast(adjustedHex, bgHex);
+    if (ratio >= threshold) {
+      return adjustedHex;
+    }
```
miles-grant-ibigroup: You can probably use recursion here?

miles-grant-ibigroup: If this method always returns a slightly stepped up value you can just have the method keep calling itself until it passes and just returns the value it was fed

packages/core-utils/src/contrast-colors.ts
```diff
@@ -0,0 +1,110 @@
...
+    const lightness = working.toHsl().l;
+    if (
+      (isLightBackground && lightness <= 0) ||
+      (!isLightBackground && lightness >= 1)
+    ) {
+      break;
+    }
+  }
+  // Fallback: return extreme color for best contrast
+  return isLightBackground ? DARK_BACKGROUND : LIGHT_BACKGROUND;
```
miles-grant-ibigroup: You can inline these colors

miles-grant-ibigroup (changes requested):
I love this functionality, and I'm excited to see this merged in! But the implementation needs some cleanup

packages/transitive-overlay/src/index.tsx
lightsage88: We get this value from _mapStyle.name_

Could we perhaps just do a string search to see if _mapStyle.name_ at any moment has 'Dark' or not? We could probably the string to compare first. Then going forward we just tell people with theming things to ensure that their mapStyle `name` property has to have `dark` in it for the goods to work. Then that way we don't have 'TriMet' in the OSS codebase.

packages/transitive-overlay/src/index.tsx
lightsage88: This kind of ties into what I was saying in the comment response above. We could assume lightness and then have one isDark, and make the logic flow from that point.

packages/transitive-overlay/src/util.ts
lightsage88: There was some intense fixups I had to do to my set up locally that involved alot of surgery. I overlooked this one where I'm importing a method just to export it. I'll have it fixed in the next push.

packages/core-utils/src/contrast-colors.ts
lightsage88: Well, sometimes we don't need a color contrast. There are many times where the itinerary line when presented against the map background color has a sufficient enough level of contrast according to the WCAG requirements we're following that it would be overkill to make th e method deliver an unnecessary color value. In these situations, which are especially common, returning null safely preserves our legally required contrast level without adding extra business to the screen.

packages/core-utils/src/contrast-colors.ts
lightsage88: It's got a way of ending itself because of the while-loop. But I have a different approach coming in my next push. I had just lifted this utility out of the TriMet planning app's files, and the issue hadn't really come up there.

packages/core-utils/src/contrast-colors.ts
lightsage88: In my forthcoming refactor of adjustColorForContrast, they are declared within the scope of the method.

packages/core-utils/src/contrast-colors.ts
lightsage88: Addressed

packages/core-utils/src/contrast-colors.ts
lightsage88: While I think that since we're feeding strings that are pre-massaged to ensure that they are in the format of a hexstring and that they are lowercased, we'd be reasonably safe from needing to handle an error situation with something this overbuilt. However, I'm not sure how `contrastRatio` handles errors and with this implementation we're at least being defensive.

I could try this, which works:
```
const getContrast = (foreground: string, background: string): number =>
 contrastRatio(foreground, background);
```

packages/core-utils/package.json
lightsage88: Yes. The adjustments will be in the next push!

packages/transitive-overlay/src/index.tsx
miles-grant-ibigroup: That's a fine compromise!

miles-grant-ibigroup (commented):
Let me know when your fixes are all done and this is ready for review again!

packages/core-utils/src/contrast-colors.ts
miles-grant-ibigroup: If you're doing this could you not just use `contrastRatio` right away?

packages/core-utils/src/contrast-colors.ts
miles-grant-ibigroup: I agree with you, I am referring more to the implementation. Right now you have 2 methods inside of your method, but what if you merged `adjust` and `adjustForBackground` and also `adjustColorForContrast`? That way you could do proper recursion. The main method should either return the value it was given if it's appropriate, or shift it if it is not.

packages/transitive-overlay/src/index.tsx
```diff
@@ -185,8 +214,12 @@ const TransitiveCanvasOverlay = ({
   transitiveData
 }: Props): JSX.Element => {
   const { current: map } = useMap();
-
-  const mapImages: MapImage[] = [];
+const mapStyleName = map?.getStyle()?.name.toLowerCase();
+  const containsDark = (str) =>
+  ["dark", "dk", "drk"].some(term =>
```
miles-grant-ibigroup: "dk" is a bit vague. Could we maybe just check for "dark" substring? That way we can also get rid of this entire method and just do the substring check directly on line 222

packages/transitive-overlay/src/index.tsx
```diff
@@ -353,11 +386,19 @@ const TransitiveCanvasOverlay = ({
...
             ),
-          ...(
-            transitiveData.patterns || []
-          ).flatMap((pattern: TransitivePattern) =>
-            patternToRouteFeature(pattern, transitiveData.routes)
-          )
+          ...(transitiveData.patterns || []).flatMap((pattern: TransitivePattern) => {
+            const feature = patternToRouteFeature(pattern, transitiveData.routes);
+            const routeColor = feature.properties.color as string;
+            let contrastColor: string | null = null;
```
miles-grant-ibigroup: Set `constrastColor` to the light color here, then you only need 1 if statement for the dark case

packages/transitive-overlay/lerna-package.json
```diff
@@ -0,0 +1,44 @@
+{
```
miles-grant-ibigroup: Please uncheck this file

packages/types/src/index.ts
```diff
@@ -610,6 +610,7 @@ export type TransitiveRoute = {
   route_color?: string;
   route_id: string;
   route_long_name: string;
+  route_outline_contrast_color?: string;
```
miles-grant-ibigroup: Please remove this

miles-grant-ibigroup (changes requested):
It's looking a lot better! But I think we can avoid the nested methods and also do some other cleanup. Also please remove all the various line ending and space/tab changes

packages/types/src/index.ts
lightsage88: Done!

packages/transitive-overlay/lerna-package.json
lightsage88: Done!

packages/transitive-overlay/src/index.tsx
lightsage88: Done!

packages/transitive-overlay/src/index.tsx
lightsage88: Done!

packages/core-utils/src/contrast-colors.ts
lightsage88: Done

packages/core-utils/src/contrast-colors.ts
lightsage88: Done

packages/core-utils/package.json
lightsage88: Done

packages/transitive-overlay/src/index.tsx
lightsage88: Done

packages/transitive-overlay/src/index.tsx
lightsage88: Done

packages/transitive-overlay/src/util.ts
lightsage88: Done

packages/transitive-overlay/src/index.tsx
```diff
@@ -185,8 +214,8 @@ const TransitiveCanvasOverlay = ({
   transitiveData
 }: Props): JSX.Element => {
   const { current: map } = useMap();
-
-  const mapImages: MapImage[] = [];
+  const mapStyleName = map?.getStyle()?.name.toLowerCase();
+  const isDark = mapStyleName?.includes("dark");  const mapImages: MapImage[] = [];
```
miles-grant-ibigroup: You can inline `mapStyleName`, since it's not used again

miles-grant-ibigroup: also please put `const mapImages` on a new line

packages/transitive-overlay/src/index.tsx
```diff
@@ -433,6 +468,20 @@ const TransitiveCanvasOverlay = ({
...
+      <Layer 
+        filter={["all", routeFilter, ["has", "contrastColor"]]}
+        type="line"
+        id="route-contrast-outlines"
+        layout={{
+          "line-cap": "round",
+          "line-join": "round"
+        }}
+        paint={{
+          "line-color": ["get", "contrastColor"],
```
miles-grant-ibigroup, line 489: Instead of all the `isDark` shenanigans in index.tsx, why not just store both values and do a match inside this get query to return contrastColorLight or contrastColorDark depending on if the name of the map style includes dark?

packages/core-utils/src/contrast-colors.ts
```diff
@@ -0,0 +1,84 @@
...
+      .hex()
+      .toLowerCase();
+
+    if (getContrast(fullHex, bgHex) >= WCAG_AA_NON_TEXT_CONTRAST) {
+      return null;
+    }
+
+    const isLightBackground = chroma(bgHex).get("hsl.l") > 0.5;
+
+    const adjust = (working: chroma.Color, iterationsLeft: number): string => {
```
miles-grant-ibigroup: Please remove the nested methods. You can achieve what you're trying to do without a while loop and without nested methods. You can make the iteration count optional and if unset, set it to MAX_ITERATIONS for the next loop. You can inline `adjustForBackground`. Finally, you can have `adjustColorForContrast` call itself, and tweak both the light and dark values as needed. You might also need to return `original`

miles-grant-ibigroup (changes requested):
We're getting close! Just a few last tweaks and we will be good to go.

packages/core-utils/src/contrast-colors.ts
miles-grant-ibigroup: Can we explore removing this method?

.storybook/main.js
```diff
@@ -51,7 +51,7 @@ export default {
           plugins: []
         }
       },
-      
+
```
miles-grant-ibigroup, line 54: Please revert this change

packages/core-utils/src/contrast-colors.ts
```diff
@@ -0,0 +1,79 @@
+import chroma from "chroma-js";
+import { contrastRatio } from "wcag-contrast-utils";
+
+const WCAG_AA_NON_TEXT_CONTRAST = 3.5; // 1.4.11 Non-text Contrast: https://www.w3.org/TR/WCAG21/#non-text-contrast
+const LIGHTNESS_STEP_PERCENT = 1; // 1% lightness adjustment per iteration
```
miles-grant-ibigroup: Remove this const

miles-grant-ibigroup (changes requested):
The new method is really great, thanks for that cleanup! I have a few final small nitpicks, and I'd also like it if we could clean up this `isDark` stuff and move that logic into the maplibre query? But I'd be ok with that as a TODO

packages/transitive-overlay/src/index.tsx
```diff
@@ -177,6 +180,32 @@ const loadImages = (map: MapRef, images: MapImage[]) => {
...
 
+
+// Route line widths (in px) by zoom level. Format is [zoom, thickness]
+export const ROUTE_LINE_WIDTH_BY_ZOOM: [number, number][] = [
+  [8, 2],
+  [10, 4],
+  [15, 6.5],
+  [19, 20],
+];
+
```
miles-grant-ibigroup: Can you convert this to an object

packages/transitive-overlay/src/index.tsx
miles-grant-ibigroup: Please address this

packages/transitive-overlay/src/index.tsx
miles-grant-ibigroup: Also, I'm not sure this is working perfectly? This is making routes with contrast show up much thicker than routes without contrast

packages/transitive-overlay/src/index.tsx
lightsage88: MapLibre GL expressions have no operator to introspect the active map style name on the fly. (there's no equivalent of `map-style-name` in the available expressions from maplibre). As `map` gets instantiated and the overlay is rendering, the `getStyle` method is undefined/null for a moment and then we end up with nothing. I tried several ways of pulling it off, but we wound up with no contrast outline lines.

The isDark variable approach is the only viable option, and storing two color properties to later select via `[["literal", isDark]]` seems to be a roundabout way of doing the same thing but with extra failure conditions.

packages/transitive-overlay/src/index.tsx
lightsage88: Addressed!

packages/core-utils/src/contrast-colors.ts
lightsage88: Addressed!

.storybook/main.js
lightsage88: I pasted the .storybook/main.js file from master. If it still shows up could it be a GH bug or something?

packages/core-utils/src/contrast-colors.ts
lightsage88: Addressed!

packages/transitive-overlay/src/index.tsx
lightsage88: Addressed

packages/transitive-overlay/src/index.tsx
```diff
@@ -353,11 +382,17 @@ const TransitiveCanvasOverlay = ({
...
+          ...(transitiveData.patterns || []).flatMap((pattern: TransitivePattern) => {
+            const feature = patternToRouteFeature(pattern, transitiveData.routes);
+            const routeColor = feature.properties.color as string;
+            let contrastColor: string | null = adjustColorForContrast(routeColor, LIGHT_BACKGROUND);
+            if (isDark) {
+              contrastColor = adjustColorForContrast(routeColor, DARK_BACKGROUND);
+            }
+            return contrastColor
+              ? { ...feature, properties: { ...feature.properties, contrastColor } }
+              : feature;
```
miles-grant-ibigroup: `contrastColor` will never be null. This can be inlined

packages/transitive-overlay/src/index.tsx
```diff
@@ -353,11 +382,17 @@ const TransitiveCanvasOverlay = ({
...
-            transitiveData.patterns || []
-          ).flatMap((pattern: TransitivePattern) =>
-            patternToRouteFeature(pattern, transitiveData.routes)
-          )
+          ...(transitiveData.patterns || []).flatMap((pattern: TransitivePattern) => {
+            const feature = patternToRouteFeature(pattern, transitiveData.routes);
+            const routeColor = feature.properties.color as string;
+            let contrastColor: string | null = adjustColorForContrast(routeColor, LIGHT_BACKGROUND);
+            if (isDark) {
+              contrastColor = adjustColorForContrast(routeColor, DARK_BACKGROUND);
```
miles-grant-ibigroup: What if instead of passing in `LIGHT_BACKGROUND` or `DARK_BACKGROUND` we just pass in `isDark` and let the method pick the color?

miles-grant-ibigroup (changes requested):
So close! Last round of changes I think. I like the new line widths much better

packages/transitive-overlay/src/index.tsx
lightsage88: It can be `null` 
image


and when it is paired with an implementation like this:
```
const routeColor = feature.properties.color as string;

return {
  ...feature,
  properties: {
    ...feature.properties,
    contrastColor: adjustColorForContrast(
      routeColor,
      isDark
    )
  }
};
```
then we end up with black lines around itinerary map lines that shouldn't have contrast outline lines
image

For this reason, I think we should leave this piece as is.

packages/transitive-overlay/src/index.tsx
lightsage88: Addressed!

packages/transitive-overlay/src/index.tsx
miles-grant-ibigroup: Can you adjust the places where the method is used to treat undefined/null as the same?

lightsage88: Addressed, I think.

packages/transitive-overlay/src/index.tsx
```diff
@@ -353,11 +382,13 @@ const TransitiveCanvasOverlay = ({
...
             ),
-          ...(
-            transitiveData.patterns || []
-          ).flatMap((pattern: TransitivePattern) =>
-            patternToRouteFeature(pattern, transitiveData.routes)
-          )
+          ...(transitiveData.patterns || []).flatMap((pattern: TransitivePattern) => {
+            const feature = patternToRouteFeature(pattern, transitiveData.routes);
+            const routeColor = feature.properties.color as string;
+            const contrastColor = adjustColorForContrast(routeColor, isDark);
```
miles-grant-ibigroup: Now that the isDark check is only used in 1 place, can we inline it?

miles-grant-ibigroup (changes requested):
Once this is done I am ready to approve!

miles-grant-ibigroup (approved):
Thank you for all your work on this!

lightsage88:
> Thank you for all your work on this!

Hey, @miles-grant-ibigroup, I accidentally re-requested your review again after you had just approved!

miles-grant-ibigroup (approved):
It's so good I approve it a second time!

.storybook/main.js
daniel-heppner-ibigroup: It's a whitespace change, maybe from your autoformatter?

packages/core-utils/src/colors.ts
```diff
@@ -0,0 +1,68 @@
...
+ * Returns:
+ * - null if the color already meets contrast against the given background
+ * - adjusted hex if modification was needed
+ * - fallback color if lightness limits or max iterations are reached
+ *
+ * Strategy:
+ * - Decide direction (darken vs lighten) based on background brightness
+ * - Recursively step lightness until contrast passes or limits are reached
+ */
+const adjustColorForContrast = (
```
daniel-heppner-ibigroup, line 26: At a high level I'd like to know why we are using this iterative process to find a color with enough contrast instead of just using chroma's luminance method which allows you to set the luminance on a color directly to achieve the desired contrast. Does it not maintain the hue/saturation in a way that you desired?

packages/core-utils/src/colors.ts
daniel-heppner-ibigroup, line 1: I think this file is too specific, can we just make a colors.ts file so we can add other color utils to it in the future if needed?

package.json
```diff
@@ -285,6 +285,7 @@
     "workerDirectory": "public"
   },
   "dependencies": {
-    "@storybook/icons": "^1.6.0"
+    "@storybook/icons": "^1.6.0",
+    "wcag-contrast-utils": "1.0.2"
```
daniel-heppner-ibigroup: Don't add dependencies to the repo level, only sub packages.

packages/transitive-overlay/src/index.tsx
```diff
@@ -1,5 +1,8 @@
+// eslint-disable-next-line prettier/prettier
+import type { ExpressionSpecification } from "maplibre-gl";
 import { FilterSpecification, SymbolLayerSpecification } from "maplibre-gl";
 import { util } from "@opentripplanner/base-map";
+import adjustColorForContrast from "@opentripplanner/core-utils/src/contrast-colors";
```
daniel-heppner-ibigroup: Import the index from opentripplanner/core-utils directly, then extract the colors module from it. Take a look at how core-utils is imported in other packages to get an idea of how this is done.

packages/core-utils/src/contrast-colors.ts
```diff
@@ -0,0 +1,68 @@
+import chroma from "chroma-js";
+import { contrastRatio } from "wcag-contrast-utils";
```
daniel-heppner-ibigroup: Chroma provides a contrast ratio function, we didn't need to add another library for this.

packages/transitive-overlay/src/RouteLineWidths.story.tsx
daniel-heppner-ibigroup, line 1: Looking at the story, are the route outlines supposed to just be black? Is this intended? 

Image

daniel-heppner-ibigroup (changes requested):

packages/core-utils/src/colors.ts
lightsage88: I think it's because different line colors on the map have a different WCAG contrast ratio. For ones where the ratio is fine, we don't need to have another contrast outline as part of it. For when the ratio doesn't meet the WCAG standard, we need to get that contrast outline color and layer involved.

By processing the colors iteratively, we avoid creating unnecessary outline layers.

packages/core-utils/src/contrast-colors.ts
lightsage88: Addressed, _wcag_contrast_utils_ is out of the picture.

package.json
lightsage88: Addressed

packages/transitive-overlay/src/index.tsx
lightsage88: Addressed!

packages/transitive-overlay/src/RouteLineWidths.story.tsx
lightsage88: Addressed! I refactored the story file to get colors using the same pathology as transitive-overlay.

packages/core-utils/src/colors.ts
lightsage88: Addressed!

packages/core-utils/src/colors.ts
miles-grant-ibigroup: I'm not sure I understand what you mean here. I think @daniel-heppner-ibigroup is just asking if you can use luminance methods to set the color luminance to a threshold that's acceptable for you rather than manually iterating

lightsage88: Addressed! I've refactored this so that we remove the iterative nature of the method.

packages/core-utils/src/colors.ts
```diff
@@ -0,0 +1,60 @@
...
+    if (chroma.contrast(colorHex, bgHex) >= CONTRAST_TARGET) return null;
+  } catch {
+    // If the contrast check throws, continue adjusting
+  }
+
+  const bgLuminance = chroma(bgHex).luminance();
+  const isLightBackground = bgLuminance > 0.5;
+
+  const targetLuminance = isLightBackground
+    ? (bgLuminance + 0.05) / CONTRAST_TARGET - 0.05 // darken the color
```
miles-grant-ibigroup: Why not just use the `lighten` method? https://gka.github.io/chroma.js/

miles-grant-ibigroup (changes requested):
I think this method has gotten more complicated than it needs to be...

packages/core-utils/src/colors.ts
lightsage88: My understanding is that `brighten` (not `lighten`) and `darken` move the color difference by an arbitrary fixed step which would require iteration on our end to converge on a target ratio. With `luminance` we can set the exact quantity the WCAG formula is built around so the target can be solved algebraically and applied all in one call.

packages/core-utils/src/colors.ts
```diff
@@ -0,0 +1,60 @@
+import chroma from "chroma-js";
+
+export const LIGHT_BACKGROUND = "#dbd9d8";
+export const DARK_BACKGROUND = "#1a364e";
+
+const CONTRAST_TARGET = 3.5;
```
miles-grant-ibigroup: Move this into the method, or inline

packages/core-utils/src/colors.ts
```diff
@@ -0,0 +1,60 @@
...
+    .toLowerCase();
+  const bgHex = chroma(backgroundColor)
+    .hex()
+    .toLowerCase();
+
+  try {
+    if (chroma.contrast(colorHex, bgHex) >= CONTRAST_TARGET) return null;
+  } catch {
+    // If the contrast check throws, continue adjusting
+  }
```
miles-grant-ibigroup: This is a confusing use of try/catch. Is there no other way to implement this

packages/core-utils/src/colors.ts
```diff
@@ -0,0 +1,60 @@
...
+ * Solving for the color's required luminance:
+ * - Light bg: L_color = (L_bg + 0.05) / CONTRAST_TARGET - 0.05
+ * - Dark  bg: L_color = CONTRAST_TARGET * (L_bg + 0.05) - 0.05
+ */
+const adjustColorForContrast = (
+  hexColor: string,
+  isDark = false
+): string | null => {
+  const backgroundColor = isDark ? DARK_BACKGROUND : LIGHT_BACKGROUND;
+  if (!chroma.valid(hexColor) || !chroma.valid(backgroundColor)) return null;
```
miles-grant-ibigroup: Can you throw a more helpful error

packages/core-utils/src/colors.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  const bgLuminance = chroma(bgHex).luminance();
+  const isLightBackground = bgLuminance > 0.5;
+
+  const targetLuminance = isLightBackground
+    ? (bgLuminance + 0.05) / CONTRAST_TARGET - 0.05 // darken the color
+    : CONTRAST_TARGET * (bgLuminance + 0.05) - 0.05; // lighten the color
+
+  if (targetLuminance < 0 || targetLuminance > 1) {
+    return isLightBackground ? DARK_BACKGROUND : LIGHT_BACKGROUND;
```
miles-grant-ibigroup: Why did you define `backgroundColor` just to recreate the logic here

packages/core-utils/src/colors.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  return chroma(colorHex)
+    .luminance(targetLuminance)
+    .hex()
+    .toLowerCase();
+};
+
+export { adjustColorForContrast };
+export default adjustColorForContrast;
```
miles-grant-ibigroup: Why are there 2 exports

packages/core-utils/src/colors.ts
```diff
@@ -0,0 +1,60 @@
...
+  } catch {
+    // If the contrast check throws, continue adjusting
+  }
+
+  const bgLuminance = chroma(bgHex).luminance();
+  const isLightBackground = bgLuminance > 0.5;
+
+  const targetLuminance = isLightBackground
+    ? (bgLuminance + 0.05) / CONTRAST_TARGET - 0.05 // darken the color
+    : CONTRAST_TARGET * (bgLuminance + 0.05) - 0.05; // lighten the color
```
miles-grant-ibigroup: Is this ternary more complicated than it should be

miles-grant-ibigroup (changes requested):

packages/core-utils/src/colors.ts
lightsage88: Addressed!

packages/core-utils/src/colors.ts
lightsage88: Addressed!

packages/core-utils/src/colors.ts
lightsage88: Quite right. I've simplified things a good deal here. Addressed!

packages/core-utils/src/colors.ts
lightsage88: Eliminated! I forgot to exclude it after some earlier versions of this seemed to required it. Addressed

packages/core-utils/src/colors.ts
lightsage88: Yes. It will now say:
```
Invalid color: "${hexColor}"
```
Addressed!

packages/core-utils/src/colors.ts
lightsage88: Addressed

packages/core-utils/src/colors.ts
```diff
@@ -0,0 +1,60 @@
...
+  if (!chroma.valid(hexColor)) throw new Error(`Invalid color: "${hexColor}"`);
+
+  const colorHex = chroma(hexColor)
+    .hex()
+    .toLowerCase();
+  const bgHex = chroma(backgroundColor)
+    .hex()
+    .toLowerCase();
+
+  if (chroma.contrast(colorHex, bgHex) >= contrastTarget) return null;
```
miles-grant-ibigroup: Can you add a comment about why you're returning null here

packages/core-utils/src/colors.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  const bgLuminance = chroma(bgHex).luminance();
+  const isLightBackground = bgLuminance > 0.5;
```
miles-grant-ibigroup: do you need this variable?

packages/core-utils/src/colors.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  const factor = isLightBackground ? 1 / contrastTarget : contrastTarget;
+  const targetLuminance = (bgLuminance + 0.05) * factor - 0.05;
+
+  if (targetLuminance < 0 || targetLuminance > 1) {
+    return fallbackColor;
+  }
+
+  return chroma(colorHex)
+    .luminance(targetLuminance)
```
miles-grant-ibigroup, line 46: This is really confusing. I see that this works but it's really hard to follow the logic. Couldn't this all be much simpler? I feel like you could avoid the backgroundColor and foreGroundcolor calculations alltogether maybe? 

It just feels like you're creating a ton of variables and branches when the actual decision the method is trying to make is very simple

miles-grant-ibigroup (changes requested):
Sorry for the endless review process, but some of the logic has been very difficult to follow

packages/core-utils/src/colors.ts
lightsage88: Addressed!

packages/core-utils/src/colors.ts
lightsage88: Nope, the value it held has been placed inline. Addressed!

packages/core-utils/src/colors.ts
lightsage88: I simplified it some. I hope it's enough! Addressed.

packages/transitive-overlay/src/index.tsx
```diff
@@ -183,6 +188,32 @@ const loadImages = (map: MapRef, images: MapImage[]) => {
...
+
+/** Multiplier for selected route contrast outline relative to focused route width */
+export const SELECTED_OUTLINE_WIDTH_MULTIPLIER = 1.2;
+
+/** Shared line-width for focused route contrast outlines */
+export const FOCUSED_ROUTE_OUTLINE_WIDTH: ExpressionSpecification = [
+  "interpolate",
+  ["linear"],
+  ["zoom"],
+  ...Object.entries(ROUTE_LINE_WIDTH_BY_ZOOM).flatMap(([zoom, width]) => [
```
miles-grant-ibigroup: Why not just store the values in the way maplibre expects? Or just inline them?

packages/transitive-overlay/src/index.tsx
```diff
@@ -364,11 +394,13 @@ const TransitiveCanvasOverlay = ({
...
-            transitiveData.patterns || []
-          ).flatMap((pattern: TransitivePattern) =>
-            patternToRouteFeature(pattern, transitiveData.routes)
-          )
+          ...(transitiveData.patterns || []).flatMap((pattern: TransitivePattern) => {
+            const feature = patternToRouteFeature(pattern, transitiveData.routes);
+            const routeColor = feature.properties.color as string;
+            const contrastColor = adjustColorForContrast(routeColor, map?.getStyle()?.name?.toLowerCase()?.includes("dark"));
+            if (!contrastColor) return feature;
+            return { ...feature, properties: { ...feature.properties, contrastColor } };
```
miles-grant-ibigroup, line 401: This can probably be inlined

miles-grant-ibigroup (changes requested):
It's better, but the logic still feels weirdly disjointed from the actual problem the method is trying to solve.

packages/transitive-overlay/src/index.tsx
lightsage88: Addressed!

packages/transitive-overlay/src/index.tsx
lightsage88: Addressed!

packages/core-utils/src/colors.ts
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * WCAG contrast formula: contrast = (L_lighter + 0.05) / (L_darker + 0.05)
+ * Solving for the color's required luminance:
+ * - Light bg: L_color = (L_bg + 0.05) / contrastTarget - 0.05
+ * - Dark  bg: L_color = contrastTarget * (L_bg + 0.05) - 0.05
+ *
+ * We use chroma's luminance(value) setter — not brighten()/darken() — because
+ * the contrast formula is defined in terms of relative luminance, so we can
+ * solve for the exact target algebraically in one step. brighten/darken adjust
+ * Lab L* by a fixed arbitrary step and would require iteration to hit a ratio.
```
miles-grant-ibigroup, line 24: Can you possibly make this comment more readable

packages/core-utils/src/colors.ts
```diff
@@ -0,0 +1,51 @@
...
+  // Color already meets contrast; no adjustment needed
+  if (chroma.contrast(hexColor, bgColor) >= contrastTarget) return null;
+
+  const bgLuminance = chroma(bgColor).luminance();
+  const factor = bgLuminance > 0.5 ? 1 / contrastTarget : contrastTarget;
+  const targetLuminance = (bgLuminance + 0.05) * factor - 0.05;
+
+  if (targetLuminance < 0 || targetLuminance > 1) {
+    return isDark ? LIGHT_BACKGROUND : DARK_BACKGROUND;
+  }
```
miles-grant-ibigroup, line 43: The logic here is still very hard to follow...

packages/transitive-overlay/src/index.tsx
miles-grant-ibigroup: can you just inline `contrastColor || null`?

packages/transitive-overlay/src/index.tsx
```diff
@@ -455,8 +500,7 @@ const TransitiveCanvasOverlay = ({
         }}
         paint={{
           "line-color": ["get", "color"],
-          // Apply a thinner line (width = 6) for bus routes (route_type = 3), set width to 10 otherwise.
-          "line-width": ["match", ["get", "routeType"], 3, 6, 10],
```
miles-grant-ibigroup, line 459: Can you please restore this distinction

miles-grant-ibigroup (changes requested):
We're getting there... Some of the logic is very difficult to follow

grant-humphries:
@miles-grant-ibigroup our dev team at TriMet has talked this over and we want to go in a different direction from the approach that's implemented here to get away from hard-coding any colors or contrast logic into this project, so I'm going to close out this PR

PR #953, opened 2026-06-02
deps: update playwright

daniel-heppner-ibigroup:
Updates playwright to try to fix the node bug preventing us from releasing otp-ui

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   "devEngines": {
     "packageManager": {
       "name": "pnpm",
-      "version": ">=11.0.0",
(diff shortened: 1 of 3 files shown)


daniel-heppner-ibigroup:
I changed the pnpm version to a specific version. If you use corepack to manage your package manager install then this will just work, but I'm not sure what will happen otherwise. Let me know!

binh-dam-ibigroup (approved):
Nice fix, thank you!

alec-georgoff (approved):

PR #955, opened 2026-06-09
feat(location-field): Option to associate custom label to input

kaylee42:
This PR adds the option to pass an id to the `LocationField` component to allow associating a label to the input. In cases where this id is passed, the `aria-label` that is usually automatically applied using the `placeholder` text is not added in order to prevent violations of [2.5.3 Label in Name](https://www.w3.org/WAI/WCAG21/Understanding/label-in-name.html).

This change is fully backwards compatible, and anyone not opting in to using the `idForInputLabel` prop will experience no change in behavior. This change is proposed in support of the TriMet code base, where we had longer placeholder text which was cut off when text spacing is adjusted in violation of [1.4.12 Text Spacing](https://www.w3.org/WAI/WCAG22/Understanding/text-spacing.html); this also supports meeting [3.3.2 Labels or Instructions](https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html) as placeholders are not sufficient to act as labels for this SC.

As a side note, it looks to me like you're using the Airbnb eslint preset rules, which is why I had to disable `jsx-a11y/label-has-associated-control` even though the label was correctly associated (y'all have several other places in the codebase with this same rule disabled with a correct association). The Airbnb presets require labels be associated both explicitly AND implicitly, which is why this flags even on a correct explicit association. IMO this is overkill and I override their rule; I prefer for explicit labels though you can also set it to either. Anyway this is just a heads up cause I think the Airbnb setting is unfortunate in a way that encourages people to disable the otherwise good linting rule.

### DIFF
--- a/packages/location-field/src/index.tsx
+++ b/packages/location-field/src/index.tsx
@@ -298,6 +298,7 @@ const LocationField = ({
   ],
   getCurrentPosition,
   hideExistingValue = false,
+  idForInputLabel,
   initialSearchResults = null,
   inputPlaceholder = null,
   isRequired = false,
@@ -1062,8 +1063,11 @@ const LocationField = ({
   let menuItemCount = indexedOptionLookup.length;
 
   /** the text input element * */
-  // Use this text for aria-label below.
-  const defaultPlaceholder = inputPlaceholder || locationType;
+  // Use this text for aria-label below if no user-provided label.
+  const defaultPlaceholder = idForInputLabel
+    ? inputPlaceholder || undefined
+    : inputPlaceholder || locationType;
+  const ariaLabel = idForInputLabel ? undefined : defaultPlaceholder;
   const placeholder =
     currentPosition && currentPosition.fetching
       ? intl.formatMessage({ id: "otpUi.LocationField.fetchingLocation" })
@@ -1081,10 +1085,11 @@ const LocationField = ({
       aria-expanded={isExpanded}
       aria-haspopup="listbox"
       aria-invalid={!isValid}
-      aria-label={defaultPlaceholder}
+      aria-label={ariaLabel}
       aria-required={isRequired}
       autoFocus={autoFocus}
       className={formControlClassname}
+      id={idForInputLabel}
       onChange={onTextInputChange}
       onClick={handleTextInputClick}
       onBlur={onFieldBlur}
--- a/packages/location-field/src/index.tsx
+++ b/packages/location-field/src/index.tsx
@@ -298,7 +298,7 @@ const LocationField = ({
   ],
   getCurrentPosition,
   hideExistingValue = false,
-  inputIdForCustomLabel,
+  customLabelInputId,
   initialSearchResults = null,
   inputPlaceholder = null,
   isRequired = false,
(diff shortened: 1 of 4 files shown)


packages/location-field/src/index.tsx
```diff
@@ -1062,8 +1063,11 @@ const LocationField = ({
   let menuItemCount = indexedOptionLookup.length;
 
   /** the text input element * */
-  // Use this text for aria-label below.
-  const defaultPlaceholder = inputPlaceholder || locationType;
+  // Use this text for aria-label below if no user-provided label.
+  const defaultPlaceholder = idForInputLabel
+    ? inputPlaceholder || undefined
+    : inputPlaceholder || locationType;
```
amy-corson-ibigroup: Putting `inputPlaceholder` on both sides of the ternary is a bit redundant. Perhaps instead: 

```
  const defaultPlaceholder =
    inputPlaceholder || idForInputLabel ? undefined : locationType;
```

packages/location-field/src/types.ts
```diff
@@ -129,6 +129,10 @@ export interface LocationFieldProps {
    * initial render.
    */
   hideExistingValue?: boolean;
+  /**
+   * Id for input so that custom label can be applied to location field
+   */
+  idForInputLabel?: string;
   /**
```
amy-corson-ibigroup, line 138: Can we make the language here a bit more explicit that if you add this prop you _must_ have a custom associated label? If you add this id and forget to add a label, the input just doesn't have a label at all.

amy-corson-ibigroup (approved):
I think this looks good, thank you for the changes! Approving, but please fix the ternary before merging, and let me know your thoughts about strengthening the languages in the types.

packages/location-field/src/index.tsx
kaylee42: Good catch, fixed!

packages/location-field/src/types.ts
kaylee42: Ah yes. I took two steps here: 
1) Added more explicit language to the comment, clarifying that if you don't provide a label yourself there won't be an accessible name
2) Changed the name of the prop itself to `inputIdForCustomLabel`, which I hope provides stronger cluing that there's an extra step you need to take yourself. 

LMK if that looks good now or if there's anything else that could help make this behavior clear!

kaylee42:
@amy-corson-ibigroup all requested updates made! This is my first contribution to OTP-UI, so please let me know if there are any other steps I should take before y'all merge this in (for instance, I think I've set my commit message up correctly to kick off the deploy, but would appreciate a confirmation of that). Thanks for your help!

packages/location-field/src/index.tsx
```diff
@@ -298,6 +298,7 @@ const LocationField = ({
   ],
   getCurrentPosition,
   hideExistingValue = false,
+  inputIdForCustomLabel,
```
miles-grant-ibigroup: Could we rename this to `customLabelInputId`? I know it's a nitpick, but it helps to have consistent phrasing especially when there's so many props!

miles-grant-ibigroup (approved):
One small nitpick and we're good to go! I can handle the actual merge since I have write permissions. 

In this repo we always do 2 approvals and then merge!

packages/location-field/src/index.tsx
kaylee42: @miles-grant-ibigroup done! I also rebased and updated snaps so that failing test should be fixed now

PR #958, opened 2026-06-24
Fix core-utils location parsing

miles-grant-ibigroup:
This bug stayed in impressively long! 

But we now make sure that if we are parsing a comma, it's actually a coordinate


daniel-heppner-ibigroup (approved):
I don't love that the tests are just using snapshots because it's harder to review them, but the others are all like that so okay

alec-georgoff (approved):

PR #960, opened 2026-07-01
strict mode: base map

daniel-heppner-ibigroup:
This PR enables strict mode for base map. There was another stacked PR situation that we had going at first, but it was too complicated to maintain that. this PR goes directly into master to keep things simple.


miles-grant-ibigroup (approved):
Great fixes!

alec-georgoff (approved):

PR #961, opened 2026-07-01
strict mode: transitive overlay

daniel-heppner-ibigroup:
Enables strict mode on the transitive overlay package.


miles-grant-ibigroup (approved):
Typescript is so strange

alec-georgoff (approved):

PR #962, opened 2026-07-01
Strict Mode: Transit Vehicle Overlay

daniel-heppner-ibigroup:
pretty straightforward one here

### DIFF
--- a/.changeset/legal-cooks-grow.md
+++ b/.changeset/legal-cooks-grow.md
@@ -1,0 +1,6 @@
+---
+"@opentripplanner/transit-vehicle-overlay": major
+"@opentripplanner/location-field": patch
+---
+
+Enable TS strict mopde in transit-vehicle-overlay package.
--- a/packages/transit-vehicle-overlay/src/index.tsx
+++ b/packages/transit-vehicle-overlay/src/index.tsx
@@ -27,7 +27,7 @@ type Props = {
    * Containing component in which route icons/numbers are rendered.
    * Can optionally support a vehicle prop.
    */
-  IconContainer?: FC;
+  IconContainer?: FC<VehicleComponentProps>;
 
   /**
    * Sets the padding between component and icon in IconContainer instances that support it.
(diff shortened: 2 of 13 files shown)


daniel-heppner-ibigroup:
gotta fix the type error here

packages/transit-vehicle-overlay/src/index.tsx
```diff
@@ -102,31 +102,35 @@ const TransitVehicleOverlay = ({
...
+  return validVehicles?.map(
+    vehicle =>
+      vehicle.lat &&
+      vehicle.lon && (
+        <MarkerWithPopup
+          key={vehicle.vehicleId}
+          autoOffset={iconPixels / 2 + iconPadding + 4}
+          popupProps={{
+            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+            // @ts-ignore why is this required? Is it some pnpm packaging issue?
```
miles-grant-ibigroup: Do we want to address this

miles-grant-ibigroup (approved):
Seems to be working well

packages/transit-vehicle-overlay/src/index.tsx
daniel-heppner-ibigroup: good point... after playing with this in the storybook I think this parameter actually doesn't do anything at all. the typescript error was right

daniel-heppner-ibigroup: I'm just gonna remove it

miles-grant-ibigroup: `MarkerWithProps` is defined in `base-map` and `popupProps` is definitely used...

miles-grant-ibigroup (changes requested):
Not comfortable removing `popupProps`

packages/transit-vehicle-overlay/src/index.tsx
daniel-heppner-ibigroup: I saw that, but it's just passed through the the maplibre component which doesn't seem to consume it. I played with some of the options we pass into it and it doesn't seem to change the output at all.

daniel-heppner-ibigroup:
update here: the padding option was not supported in our version of maplibre, however the newer versions do support it. I updated maplibre and restored the padding option. 

so I wonder if at some point we had been using a newer version, or maybe OTP-RR has a newer version? anyway, it's restored now.

miles-grant-ibigroup (approved):
Lovely

.changeset/legal-cooks-grow.md
```diff
@@ -0,0 +1,6 @@
+---
+"@opentripplanner/transit-vehicle-overlay": major
+"@opentripplanner/location-field": patch
+---
+
+Enable TS strict mopde in transit-vehicle-overlay package.
```
binh-dam-ibigroup, line 6: Fix the typo.

.changeset/legal-cooks-grow.md
```diff
@@ -0,0 +1,6 @@
+---
+"@opentripplanner/transit-vehicle-overlay": major
```
binh-dam-ibigroup, line 2: What was the breaking change? I don't see a corresponding commit.

packages/transit-vehicle-overlay/src/index.tsx
```diff
@@ -102,31 +102,33 @@ const TransitVehicleOverlay = ({
...
-      </StyledContainer>
-    </MarkerWithPopup>
-  ));
+  return validVehicles?.map(
+    vehicle =>
+      vehicle.lat &&
+      vehicle.lon && (
+        <MarkerWithPopup
+          key={vehicle.vehicleId}
+          autoOffset={iconPixels / 2 + iconPadding + 4}
```
binh-dam-ibigroup, line 111: Nit - sort props

binh-dam-ibigroup (commented):
Do we need a new major version for transit-vehicle-overlay?

PR #965, opened 2026-07-01
Strict mode: park and ride overlay

daniel-heppner-ibigroup:
Some of the logic is moved a little bit here but overall not much changed.

### DIFF
--- a/packages/park-and-ride-overlay/src/index.tsx
+++ b/packages/park-and-ride-overlay/src/index.tsx
@@ -2,8 +2,9 @@ import {
   Styled as BaseMapStyled,
   MarkerWithPopup
 } from "@opentripplanner/base-map";
-import FromToLocationPicker from "@opentripplanner/from-to-location-picker";
-import { Location } from "@opentripplanner/types";
+import FromToLocationPicker, {
+  FromToLocation
+} from "@opentripplanner/from-to-location-picker";
 import React from "react";
 import ParkAndRideMarker from "./park-and-ride-marker";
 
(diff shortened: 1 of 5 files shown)


miles-grant-ibigroup (approved):
Very upset about the null/undefined

packages/park-and-ride-overlay/src/index.tsx
```diff
@@ -2,30 +2,30 @@ import {
...
 };
 
-const ParkAndRideOverlay = (props: Props): JSX.Element => {
+const ParkAndRideOverlay = (props: Props): JSX.Element | null => {
   const { parkAndRideLocations, setLocation } = props;
   if (!parkAndRideLocations || parkAndRideLocations.length === 0) return null;
 
+  const setLocationOverride = setLocation
+    ? (l: FromToLocation) => {
+        setLocation(undefined);
```
alec-georgoff, line 24: Why aren't we able to just `setLocation(1)` without this intermediary step? If there's some reason why, I think a comment explaining would be helpful

daniel-heppner-ibigroup: I have no idea, but that was something that was here before this PR, I just moved the function around a bit. I'm guessing it's solving some bug in a consuming component.

daniel-heppner-ibigroup: I'd love to write a comment explaining why this is here but I'd be opening a can of worms. This doesn't change any behavior and it improves the health of the codebase, so I think it's okay to merge as is.

alec-georgoff (approved):

PR #966, opened 2026-07-01
Strict mode: map popup

daniel-heppner-ibigroup:
This one is a bit more complex than the others. A lot of the diff is simple formatting changes, but there is some slight logic changes as well. 

- The change to generateLocation was because I couldn't identify anywhere that x and y were present in the map popup entity. I think it might be a legacy thing? It didn't exist in the type.

### DIFF
--- a/packages/map-popup/src/index.tsx
+++ b/packages/map-popup/src/index.tsx
@@ -1,7 +1,10 @@
 import React, { useCallback } from "react";
 import FromToLocationPicker from "@opentripplanner/from-to-location-picker";
 
-import { RentalVehicle, VehicleRentalStation } from "@opentripplanner/types/otp2";
+import {
+  RentalVehicle,
+  VehicleRentalStation,
+} from "@opentripplanner/types/otp2";
 // eslint-disable-next-line prettier/prettier
 import type {
   Company,
@@ -27,7 +30,9 @@ import defaultEnglishMessages from "../i18n/en-US.yml";
 // the YAML loaders behave differently between webpack and our version of jest:
 // - the yaml loader for webpack returns a nested object,
 // - the yaml loader for jest returns messages with flattened ids.
-export const defaultMessages: { [key: string]: string } = flatten(defaultEnglishMessages);
+export const defaultMessages: { [key: string]: string } = flatten(
+  defaultEnglishMessages,
+);
 
 export type Feed = {
   feedId: string;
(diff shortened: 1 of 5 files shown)


packages/map-popup/src/index.tsx
```diff
@@ -36,162 +41,198 @@ export type Feed = {
...
-  // @ts-expect-error some of these values may be null, but that's ok
-  const { lon: entityLon, lat: entityLat, x, y } = entity
-
-  const lat = entityLat || y
-  const lon = entityLon || x
-  if (!lat || !lon) return null
-
+const generateLocation = (entity: MapPopupEntity, name?: string) => {
+  const { lon, lat } = entity;
+  if (!lat || !lon) return undefined;
```
miles-grant-ibigroup, line 46: I believe the type is wrong... We definitely get `x` and `y` sometimes...

packages/map-popup/src/util.ts
```diff
@@ -27,19 +31,19 @@ export function makeDefaultGetEntityName(
...
+    let stationName: string | undefined = entity.name ?? entity.id;
     // If the station name or id is a giant UUID (with more than 3 "-" characters)
     // best not to show that at all. The company name will still be shown.
     // Also ignore "Default Vehicle Type"
     if (
       (stationName.match(/-/g) || []).length > 3 ||
       stationName === "Default vehicle type"
     ) {
-      stationName = null;
+      stationName = undefined;
```
miles-grant-ibigroup, line 46: `null` is clearly such a different case here from `undefined`! `undefined` means it is blank from the server, `null` means we set it to null to hide an unwanted string

miles-grant-ibigroup (changes requested):
This diff is bizarre... 

Anyway the x/y stuff needs to stay I think... Something with certain vector tiles?

packages/map-popup/src/util.ts
daniel-heppner-ibigroup: it doesn't matter! there's no reason to overcomplicate the codebase for this differentiation that no consumer will care about. All that does is introduce the possibility of bugs if you forget to check for both undefined and null.

packages/map-popup/src/index.tsx
daniel-heppner-ibigroup: I just can't find a place where they are used... You're right that vector tiles provide and x and y, but that's not lat and lon, it's coordinates in the mapbox grid system. And they get converted to the lat/lon entity which we receive by maplibre. 
see: https://mapbox.github.io/vector-tile-spec/#/0001/01/03/format

stops vector tile:
image

lime scooter:
image

packages/map-popup/src/util.ts
miles-grant-ibigroup: Seeing null/undefined in a debugger makes things much easier to debug! But I see your side too.

packages/map-popup/src/index.tsx
miles-grant-ibigroup: Did you check when the map-popup is rendered by stops-overlay? Or it might be when it's rendered inside otp-rr directly I think

PR #967, opened 2026-07-01
Strict mode: otp2 tile overlay

daniel-heppner-ibigroup:
Pushing my undefined > null agenda big time here

### DIFF
--- a/packages/otp2-tile-overlay/src/index.tsx
+++ b/packages/otp2-tile-overlay/src/index.tsx
@@ -95,9 +95,10 @@ const OTP2TileLayerWithPopup = ({
   const [clickedEntity, setClickedEntity] = useState<any>(null);
 
   const defaultClickHandler = (event: MapLayerMouseEvent) => {
-    const sourceLayer = event.features?.[0].sourceLayer;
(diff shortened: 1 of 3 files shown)


miles-grant-ibigroup (approved):
Really hate the undefined stuff

packages/otp2-tile-overlay/src/index.tsx
```diff
@@ -85,14 +85,17 @@ const OTP2TileLayerWithPopup = ({
...
+  if (!map) {
+    return undefined;
+  }
 
   // TODO: handle this complex type: it can be a stop, a station, and some extra fields too
   const [clickedEntity, setClickedEntity] = useState<any>(null);
 
   const defaultClickHandler = (event: MapLayerMouseEvent) => {
-    const { sourceLayer } = event.features?.[0];
+    const sourceLayer = event.features?.[0].sourceLayer;
```
alec-georgoff: What happens if `features` is an empty array?

daniel-heppner-ibigroup: hm interesting. Looks like we need another TS flag to make this throw an error. Good call

alec-georgoff (approved):

PR #968, opened 2026-07-01
Strict mode: route viewer overlay

daniel-heppner-ibigroup:
This was one of the more complex changes I had to do. I spent a lot of time trying to understand the purpose of the old code but I'm not sure I got it perfect so worth double checking!

### DIFF
--- a/packages/route-viewer-overlay/src/index.tsx
+++ b/packages/route-viewer-overlay/src/index.tsx
@@ -149,13 +149,13 @@ const extractCoordinatesFromGeometry = (
  * @returns Extended bounds object
  */
 const extendBoundsWithGeometry = (
-  bounds: LngLatBounds | null,
+  bounds: LngLatBounds | undefined,
   geometry: GeoJSON.Geometry
-): LngLatBounds => {
+): LngLatBounds | undefined => {
   const coordinates = extractCoordinatesFromGeometry(geometry);
 
   if (coordinates.length === 0) {
-    return bounds || new LngLatBounds([0, 0], [0, 0]);
+    return bounds ?? new LngLatBounds([0, 0], [0, 0]);
   }
 
   let newBounds = bounds;
@@ -203,7 +203,8 @@ const removePointsInFlexZone = (stops: Stop[], points: [number, number][]) => {
 /**
  * Reducer helper for computing the bounds of a geometry.
  */
-const reduceBounds = (bnds, coord) => bnds.extend(coord);
+const reduceBounds = (bnds: LngLatBounds, coord: LngLatLike) =>
+  bnds.extend(coord);
 
 /**
  * An overlay that will display all polylines of the patterns of a route.
@@ -215,11 +216,13 @@ const RouteViewerOverlay = (props: Props): JSX.Element => {
   useEffect(() => {
     // if pattern geometry updated, update the map points
     let bounds: LngLatBounds | undefined;
-    let timeout: NodeJS.Timeout | undefined;
+    let timeout: number | undefined;
 
     if (isGeometryComplete(routeData)) {
-      const allPoints: LngLatLike[] = patterns.reduce((acc, ptn) => {
-        return acc.concat(polyline.decode(ptn.geometry.points));
+      const allPoints = patterns.reduce<[number, number][]>((acc, ptn) => {
+        return ptn.geometry?.points
+          ? acc.concat(polyline.decode(ptn.geometry?.points))
+          : acc;
       }, []);
 
       if (allPoints.length > 0) {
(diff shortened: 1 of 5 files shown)


packages/route-viewer-overlay/src/index.tsx
```diff
@@ -215,11 +216,13 @@ const RouteViewerOverlay = (props: Props): JSX.Element => {
   useEffect(() => {
     // if pattern geometry updated, update the map points
     let bounds: LngLatBounds | undefined;
-    let timeout: NodeJS.Timeout | undefined;
+    let timeout: number | undefined;
```
miles-grant-ibigroup, line 219: We're saving the value returned from `setTimeout` into here.

packages/route-viewer-overlay/src/index.tsx
```diff
@@ -265,16 +272,17 @@ const RouteViewerOverlay = (props: Props): JSX.Element => {
...
+    if (!objectExistsAndPopulated(pattern?.geometry) || !pattern?.stops) {
+      return [];
+    }
+    const pts = polyline.decode(pattern.geometry?.points);
+    const clippedPts = clipToPatternStops
+      ? removePointsInFlexZone(pattern?.stops, pts)
+      : pts;
+
+    return [clippedPts.map((pt: [number, number]) => [pt[1], pt[0]])];
+  });
```
miles-grant-ibigroup, line 285: Why `flatMap` and returning an array instead of just `map?`

packages/route-viewer-overlay/src/util.ts
```diff
@@ -4,8 +4,8 @@ import { LngLatBounds } from "maplibre-gl";
  * Method to test if an object not only exists but contains entries
  */
 // eslint-disable-next-line import/prefer-default-export
-export const objectExistsAndPopulated = (
-  o: LngLatBounds | Record<string, unknown>
-): boolean => {
+export function objectExistsAndPopulated<
+  T extends LngLatBounds | Record<string, unknown>
+>(o: T | undefined): o is T {
```
miles-grant-ibigroup, line 9: 🫪

miles-grant-ibigroup (changes requested):
Seems to be working well, I just have some small nits

packages/route-viewer-overlay/src/index.tsx
daniel-heppner-ibigroup: yes but it runs in the browser and the browser uses numbers for the timeout value. NodeJS.Timeout is for node environment.

daniel-heppner-ibigroup: ultimately, the type doesn't matter at runtime since it's stripped out anyway. But this is the more accurate representation. And we had number in some places and timeout in other places which _was_ causing type errors, so I needed to make it consistent.

packages/route-viewer-overlay/src/index.tsx
daniel-heppner-ibigroup: flatMap instead of filter and map. It just is one fewer loop through the segments, so it should be more efficient. Not sure about interpreter optimizations here though.

packages/route-viewer-overlay/src/util.ts
daniel-heppner-ibigroup: this took me a long time to figure out and i think it's still not technically the most strict return type but it is good enough

packages/route-viewer-overlay/src/index.tsx
miles-grant-ibigroup: I think this is a time/space issue but I don't think it's worth getting hung up on

miles-grant-ibigroup (approved):
It's beautiful

packages/route-viewer-overlay/src/index.tsx
daniel-heppner-ibigroup: i'm in my flatmap era <3

packages/route-viewer-overlay/src/index.tsx
```diff
@@ -149,13 +149,13 @@ const extractCoordinatesFromGeometry = (
  * @returns Extended bounds object
  */
 const extendBoundsWithGeometry = (
-  bounds: LngLatBounds | null,
+  bounds: LngLatBounds | undefined,
   geometry: GeoJSON.Geometry
-): LngLatBounds => {
+): LngLatBounds | undefined => {
```
alec-georgoff, line 154: Why are we updating to allow undefined return instead of the old "default" value? Also, if we're allowing undefined returns, I think the JSdoc comment should be updated to reflect that.

packages/route-viewer-overlay/src/index.tsx
```diff
@@ -265,16 +272,17 @@ const RouteViewerOverlay = (props: Props): JSX.Element => {
...
-      return clippedPts.map((pt: [number, number]) => [pt[1], pt[0]]);
-    });
+  const segments = patterns.flatMap(pattern => {
+    if (!objectExistsAndPopulated(pattern?.geometry) || !pattern?.stops) {
+      return [];
+    }
+    const pts = polyline.decode(pattern.geometry?.points);
+    const clippedPts = clipToPatternStops
+      ? removePointsInFlexZone(pattern?.stops, pts)
+      : pts;
```
alec-georgoff: I think we should remove the optional chaining on lines 279 and 281 since they're made redundant by the checks on 276. Initially I was worried about them throwing errors before I realized that the checks already occurred.

packages/route-viewer-overlay/src/index.tsx
daniel-heppner-ibigroup: It always could return undefined, we're just defining it correctly in the TS now. It only returns undefined if bounds is undefined.

packages/route-viewer-overlay/src/index.tsx
daniel-heppner-ibigroup: sure that works!

alec-georgoff (approved):

PR #972, opened 2026-07-10
fix(map-popup): Only render "view stop" link if setViewedStop exists

amy-corson-ibigroup:
If there's no `setViewedStop` handler, we shouldn't render the link! Simple as that. Should allow a little flexibility for anyone who doesn't want that link (for example - running the trip planner in iframe without nearby view) to just not pass the function.


josh-willis-arcadis (approved):
very clean!

alec-georgoff (approved):

PR #973, opened 2026-07-13
feat(building-blocks): Create a no-icon version of the alert

amy-corson-ibigroup:
Now passing the string "no-icon" to the `icon` prop will render an alert with no icon

with icon
image

without icon: 
image

### DIFF
--- a/packages/building-blocks/src/alert/Alert.tsx
+++ b/packages/building-blocks/src/alert/Alert.tsx
@@ -15,20 +15,20 @@ interface Props {
   collapsible?: boolean;
   children?: JSX.Element | string;
   /* Icon displayed on Alert - default is Bell. "no-icon" displays no icon */
-  Icon?: StyledIcon | "no-icon";
+  Icon?: StyledIcon | null;
 }
 
 const AlertContainer = styled.div<{
   backgroundColor?: string;
   collapsible: boolean;
   expandAlert: boolean;
-  noIcon: boolean;
+  icon: boolean;
 }>`
   background-color: ${props =>
     props.backgroundColor ? props.backgroundColor : blue[50]};
   display: grid;
   grid-template-columns: ${props =>
-    props.noIcon ? "auto auto" : "50px auto auto"};
+    props.icon ? "50px auto auto" : "auto auto"};
   grid-template-rows: minmax(25px, auto) auto;
   column-gap: 1em;
   padding: 1.5em;
@@ -45,7 +45,7 @@ const AlertContainer = styled.div<{
   }
 
   .content-row {
-    grid-column: ${props => (props.noIcon ? 1 : 2)};
+    grid-column: ${props => (props.icon ? 2 : 1)};
   }
 
   .alert-body {
@@ -72,7 +72,7 @@ const AlertContainer = styled.div<{
 
   @media (max-width: 550px) {
     grid-template-columns: 40px auto auto;
-    padding: 1.25em 1.25em;
+    padding: 1.25em;
   }
 `;
 
@@ -103,15 +103,14 @@ const Alert = ({
     id: "otpUi.buildingBlocks.alert.expand",
     defaultMessage: "Expand"
   });
-  const noIcon = Icon === "no-icon";
   return (
     <AlertContainer
       backgroundColor={backgroundColor}
       expandAlert={expandAlert}
       collapsible={!!collapsible}
-      noIcon={noIcon}
+      icon={!!Icon}
     >
-      {!noIcon && <Icon size={24} />}
+      {!!Icon && <Icon size={24} />}
       <span className="header content-row">{alertHeader}</span>
       <ButtonContainer>
         {collapsible && (
--- a/packages/building-blocks/src/alert/Alert.tsx
+++ b/packages/building-blocks/src/alert/Alert.tsx
@@ -14,22 +14,43 @@ interface Props {
   /* If true, adds a toggle that expands/collapses the alert content */
   collapsible?: boolean;
   children?: JSX.Element | string;
-  /* Icon displayed on Alert - default is Bell */
-  Icon?: StyledIcon;
+  /* Icon displayed on Alert - default is Bell. "None" displays no icon */
+  Icon?: StyledIcon | "no-icon";
 }
 
 const AlertContainer = styled.div<{
-  backgroundColor: string;
+  backgroundColor?: string;
   collapsible: boolean;
   expandAlert: boolean;
+  noIcon: boolean;
 }>`
   background-color: ${props =>
     props.backgroundColor ? props.backgroundColor : blue[50]};
   display: grid;
-  grid-template-columns: 50px auto auto;
+  grid-template-columns: ${props =>
+    props.noIcon ? "auto auto" : "50px auto auto"};
   grid-template-rows: minmax(25px, auto) auto;
   column-gap: 1em;
   padding: 1.5em;
+
+  .header {
+    align-self: center;
+    font-weight: 700;
+  }
+
+  .subheader {
+    align-self: center;
+    font-weight: 400;
+    margin-top: 0.3em;
+  }
+
+  .content-row {
+    grid-column: ${props => (props.noIcon ? 1 : 2)};
+  }
+
+  .alert-body {
+    grid-row: 3;
+  }
 
   svg {
     align-self: center;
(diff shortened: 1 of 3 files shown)


packages/building-blocks/src/alert/Alert.tsx
```diff
@@ -14,23 +14,44 @@ interface Props {
   /* If true, adds a toggle that expands/collapses the alert content */
   collapsible?: boolean;
   children?: JSX.Element | string;
-  /* Icon displayed on Alert - default is Bell */
-  Icon?: StyledIcon;
+  /* Icon displayed on Alert - default is Bell. "None" displays no icon */
+  Icon?: StyledIcon | "no-icon";
```
miles-grant-ibigroup: Can we just use null?

packages/building-blocks/src/alert/Alert.tsx
```diff
@@ -98,14 +103,16 @@ const Alert = ({
...
       backgroundColor={backgroundColor}
       expandAlert={expandAlert}
-      collapsible={collapsible}
+      collapsible={!!collapsible}
+      noIcon={noIcon}
     >
-      <Icon size={24} />
-      <AlertHeader>{alertHeader}</AlertHeader>
+      {!noIcon && <Icon size={24} />}
+      <span className="header content-row">{alertHeader}</span>
```
miles-grant-ibigroup, line 115: Can we reduce the padding when no icon is present?

packages/building-blocks/src/alert/Alert.tsx
```diff
@@ -98,14 +103,16 @@ const Alert = ({
...
     defaultMessage: "Expand"
   });
+  const noIcon = Icon === "no-icon";
   return (
     <AlertContainer
       backgroundColor={backgroundColor}
       expandAlert={expandAlert}
-      collapsible={collapsible}
+      collapsible={!!collapsible}
+      noIcon={noIcon}
```
miles-grant-ibigroup: Can we instead support overriding `Icon`? And then we get no icon for "free"

miles-grant-ibigroup (changes requested):
I see potential to make things simpler

packages/building-blocks/src/alert/Alert.tsx
amy-corson-ibigroup: Scares me because it does look worse in the stories, but you're right it looks better in the app!

packages/building-blocks/src/alert/Alert.tsx
```diff
@@ -14,22 +14,43 @@ interface Props {
   /* If true, adds a toggle that expands/collapses the alert content */
   collapsible?: boolean;
   children?: JSX.Element | string;
-  /* Icon displayed on Alert - default is Bell */
-  Icon?: StyledIcon;
+  /* Icon displayed on Alert - default is Bell. "no-icon" displays no icon */
```
miles-grant-ibigroup: We have to update this comment!

miles-grant-ibigroup (approved):
Love it!

packages/building-blocks/src/alert/Alert.tsx
```diff
@@ -14,22 +14,43 @@ interface Props {
...
   collapsible: boolean;
   expandAlert: boolean;
+  icon: boolean;
 }>`
   background-color: ${props =>
     props.backgroundColor ? props.backgroundColor : blue[50]};
   display: grid;
-  grid-template-columns: 50px auto auto;
+  grid-template-columns: ${props =>
+    props.icon ? "50px auto auto" : "auto auto"};
```
binh-dam-ibigroup, line 31: Is this me, or is this a flex layout in disguise?

packages/building-blocks/src/stories/alerts.story.tsx
```diff
@@ -70,6 +70,20 @@ export const BasicAlert = (): ReactElement => {
   );
 };
 
+export const BasicAlertNoIcon = (): ReactElement => {
+  return (
```
binh-dam-ibigroup: Nit - You can replace `return` with the `=> (<AlertContainer>...)` syntax.

packages/building-blocks/src/alert/Alert.tsx
```diff
@@ -102,10 +107,11 @@ const Alert = ({
     <AlertContainer
       backgroundColor={backgroundColor}
       expandAlert={expandAlert}
-      collapsible={collapsible}
+      collapsible={!!collapsible}
```
binh-dam-ibigroup: Do we need `!!collapsible`? (that variable is already `boolean | undefined`)

packages/building-blocks/src/alert/Alert.tsx
```diff
@@ -14,22 +14,43 @@ interface Props {
...
   background-color: ${props =>
     props.backgroundColor ? props.backgroundColor : blue[50]};
   display: grid;
-  grid-template-columns: 50px auto auto;
+  grid-template-columns: ${props =>
+    props.icon ? "50px auto auto" : "auto auto"};
   grid-template-rows: minmax(25px, auto) auto;
   column-gap: 1em;
-  padding: 1.5em;
+  padding: ${props => (props.icon ? "1.5em" : "1em")};
```
binh-dam-ibigroup, line 34: Why is the entire container padding depending on the presence of the icon?

packages/building-blocks/src/alert/Alert.tsx
```diff
@@ -51,7 +72,7 @@ const AlertContainer = styled.div<{
 
   @media (max-width: 550px) {
     grid-template-columns: 40px auto auto;
-    padding: 1.25em 1.25em;
```
binh-dam-ibigroup, line 54: On narrow viewports, the layout is really odd, and you should address that (see screenshot below).

Image

binh-dam-ibigroup (changes requested):
The layout without an icon is really strange on narrow viewports, but does that come from using a grid instead of a flex CSS layout?

packages/building-blocks/src/alert/Alert.tsx
amy-corson-ibigroup: The columns could probably be handled with flex, but the rows make things a little trickier... I think `grid` is a little easier to work with because we can straightforwardly assign the icon, header, body to rows/columns, but I'm open to trying to recreate this in flex if you'd prefer!

packages/building-blocks/src/alert/Alert.tsx
amy-corson-ibigroup: This was [something Miles suggested](https://github.com/opentripplanner/otp-ui/pull/973#discussion_r3572802382) and after applying it, I do think it looks a little better! Extra whitespace around the icon makes things look a little cleaner, but it's not as necessary with fewer elements in the alert.

packages/building-blocks/src/alert/Alert.tsx
amy-corson-ibigroup: Thank you so much for catching this! This was due to a media query that I didn't update, but it's been fixed in [1816a16](https://github.com/opentripplanner/otp-ui/pull/973/commits/1816a16430501b11ca717a8529e7c87dfe3ff4a8)

packages/building-blocks/src/alert/Alert.tsx
amy-corson-ibigroup: Good call, this was because the prop was not optional and so the `undefined` was throwing an error. Fixed in [1816a16](https://github.com/opentripplanner/otp-ui/pull/973/commits/1816a16430501b11ca717a8529e7c87dfe3ff4a8)

packages/building-blocks/src/alert/Alert.tsx
binh-dam-ibigroup: IMO, if it were flex, you wouldn't have to pass the props to that component. But no need to convert the layout in this PR.

packages/building-blocks/src/alert/Alert.tsx
binh-dam-ibigroup: Nit picking here too and something for another PR, I suspect that, ultimately, you don't need the `icon` prop to determine the padding, and that you can do it via CSS selectors.

binh-dam-ibigroup (approved):
Thanks for the fixes! Please update the snapshots before proceeding.

PR #974, opened 2026-07-13
Annotate canceled trips in itinerary body

alec-georgoff:
This update strikes through any transit trips that have a `realtimeState` of `CANCELED` and adds an annotation next to the place name:

Screenshot 2026-07-13 at 3 19 39 PM

### DIFF
--- a/packages/itinerary-body/i18n/en-US.yml
+++ b/packages/itinerary-body/i18n/en-US.yml
@@ -69,6 +69,7 @@ otpUi:
       escooter: E-scooter
       vehicle: Vehicle
   ItineraryBody:
+    canceled: Canceled
     common:
       durationShort: >-
         {approximatePrefix, select, true {About } other {}}{hours, plural, =0 {}
--- a/packages/itinerary-body/i18n/en-US.yml
+++ b/packages/itinerary-body/i18n/en-US.yml
@@ -70,6 +70,7 @@ otpUi:
       vehicle: Vehicle
   ItineraryBody:
     canceled: Canceled
+    canceledMessage: This trip has been canceled
     common:
       durationShort: >-
         {approximatePrefix, select, true {About } other {}}{hours, plural, =0 {}
--- a/packages/itinerary-body/i18n/en-US.yml
+++ b/packages/itinerary-body/i18n/en-US.yml
@@ -70,7 +70,7 @@ otpUi:
       vehicle: Vehicle
   ItineraryBody:
     canceled: Canceled
-    canceledMessage: This trip has been canceled
+    canceledMessage: The following trip has been canceled
     common:
       durationShort: >-
         {approximatePrefix, select, true {About } other {}}{hours, plural, =0 {}
--- a/packages/itinerary-body/i18n/fr.yml
+++ b/packages/itinerary-body/i18n/fr.yml
@@ -76,6 +76,7 @@ otpUi:
       escooter: Trottinette
       vehicle: Véhicule
   ItineraryBody:
+    canceled: Annulé
     common:
       durationShort: >-
         {approximatePrefix, select, true {Environ } other {}}{hours, plural, =0
--- a/packages/itinerary-body/i18n/fr.yml
+++ b/packages/itinerary-body/i18n/fr.yml
@@ -77,6 +77,7 @@ otpUi:
       vehicle: Véhicule
   ItineraryBody:
     canceled: Annulé
+    canceledMessage: This trip has been canceled
     common:
       durationShort: >-
         {approximatePrefix, select, true {Environ } other {}}{hours, plural, =0
--- a/packages/itinerary-body/i18n/fr.yml
+++ b/packages/itinerary-body/i18n/fr.yml
@@ -77,7 +77,7 @@ otpUi:
       vehicle: Véhicule
   ItineraryBody:
     canceled: Annulé
-    canceledMessage: This trip has been canceled
+    canceledMessage: The following trip has been canceled
     common:
       durationShort: >-
         {approximatePrefix, select, true {Environ } other {}}{hours, plural, =0
--- a/packages/itinerary-body/src/ItineraryBody/place-row.tsx
+++ b/packages/itinerary-body/src/ItineraryBody/place-row.tsx
@@ -16,6 +16,7 @@ import { defaultMessages } from "../util";
 
 const CanceledTripMessage = styled.span`
   color: ${colors.red[700]};
+  padding-top: 3px;
 `;
 
 function getLegPlaceName(
@@ -118,10 +119,15 @@ export default function PlaceRow({
     description: "Text describing the view-on-map button",
     id: "otpUi.ItineraryBody.viewOnMap"
   });
-  const canceledMessage = intl.formatMessage({
+  const canceledText = intl.formatMessage({
     defaultMessage: defaultMessages["otpUi.ItineraryBody.canceled"],
     description: "Text indicating a canceled trip",
     id: "otpUi.ItineraryBody.canceled"
+  });
+  const canceledInvisibleMessage = intl.formatMessage({
+    defaultMessage: defaultMessages["otpUi.ItineraryBody.canceledMessage"],
+    description: "Screen reader text indicating a canceled trip",
+    id: "otpUi.ItineraryBody.canceledMessage"
   });
 
   return (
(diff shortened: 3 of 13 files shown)


packages/itinerary-body/src/ItineraryBody/place-row.tsx
```diff
@@ -12,6 +13,10 @@ import AccessibilityRating from "./accessibility-rating";
 import { PlaceNameProps, PlaceRowProps } from "../types";
 import { defaultMessages } from "../util";
 
+const CanceledTripMessage = styled.span`
+  color: rgb(183, 38, 32);
```
amy-corson-ibigroup: Can you import the `red` from the color package in building-blocks? There should be other examples of it in itinerary body but lmk if you need help! Anything `[700]` or over should be accessible on white.

packages/itinerary-body/src/ItineraryBody/place-row.tsx
```diff
@@ -131,11 +142,18 @@ export default function PlaceRow({
...
       <S.PlaceHeader>
-        <S.PlaceName aria-hidden className="place-row-place-name">
+        <S.PlaceName
+          aria-hidden
+          className="place-row-place-name"
+          strikethrough={canceled}
+        >
           {placeName}
         </S.PlaceName>
+        {canceled && (
```
amy-corson-ibigroup, line 163: Can we move this message to be in-line with the place text, and potentially create a gap between the text and the place name? It's pretty far away in one set of stories and way too close in the other. 

Image

packages/itinerary-body/src/ItineraryBody/place-row.tsx
josh-willis-arcadis: we don't use rgb very often in the project. Any reason why this was used? Maybe we can use the red color from the building blocks package for consistency.

packages/itinerary-body/src/ItineraryBody/place-row.tsx
amy-corson-ibigroup: I think we could drop the text size down too? But maybe that wouldn't look right.

packages/itinerary-body/src/stories/OtpRrItineraryBody.story.tsx
```diff
@@ -131,6 +131,15 @@ export const WalkTransitWalkItinerary = (): ReactElement => (
   <OtpRRItineraryBodyWrapper itinerary={walkTransitWalkItinerary} />
 );
 
+const walkTransitWalkItineraryCanceled = { ...walkTransitWalkItinerary };
```
amy-corson-ibigroup: I'm seeing a bug in the stories where the standard "walk transit walk" story is also coming up as cancelled.

packages/itinerary-body/i18n/fr.yml
```diff
@@ -76,6 +76,7 @@ otpUi:
       escooter: Trottinette
       vehicle: Véhicule
   ItineraryBody:
+    canceled: Annulé
```
josh-willis-arcadis, line 79: I usually put the english word here and the weblate community will translate. Google translate says this correct though!

amy-corson-ibigroup (changes requested):
A few things on this one: 
1. The "canceled" is going to be read by screenreaders after the place name, which is going to make it confusing and inaccessible to blind users. Can you add an `aria-hidden` to this new component, and create invisible text at the top of the trip that says "this trip has been canceled"?
2. In order to keep a consistent design around realtime statuses, I am wondering if you can place the 'canceled" text in the time column like we do if a trip is late or early. If you do use this approach, there might already be built in code for the screenreader message: 

Image

packages/itinerary-body/src/stories/OtpRrItineraryBody.story.tsx
josh-willis-arcadis: same. A deeper clone is needed to avoid changing the legs in this itinerary for other stories.

packages/itinerary-body/src/stories/OtpUiItineraryBody.story.tsx
```diff
@@ -63,6 +63,18 @@ export const WalkTransitWalkItinerary = (): ReactElement => (
   />
 );
 
+const walkTransitWalkItineraryCanceled = { ...walkTransitWalkItinerary };
```
josh-willis-arcadis: same bug here

josh-willis-arcadis (approved):
same concerns as @amy-corson-ibigroup plus a nit or 2

packages/itinerary-body/src/stories/OtpRrItineraryBody.story.tsx
alec-georgoff: Good catch, fixed!

packages/itinerary-body/src/ItineraryBody/place-row.tsx
alec-georgoff: Oh yeah sorry about that, I saw the accessibility error and just copied the raw value of a different red in the UI using the inspector...your solution is much better!

packages/itinerary-body/src/ItineraryBody/place-row.tsx
alec-georgoff: I was able to get it aligned vertically; I think the text size should stay the same right now for visibility, but we can adjust if consumers aren't happy with it.

I added a left margin for the cases where it's way too close, but I'm struggling a bit with the opposite case where it's far away. I'll add a follow-up to improve that.

alec-georgoff:
> A few things on this one:
> 
> 1. The "canceled" is going to be read by screenreaders after the place name, which is going to make it confusing and inaccessible to blind users. Can you add an `aria-hidden` to this new component, and create invisible text at the top of the trip that says "this trip has been canceled"?
> 2. In order to keep a consistent design around realtime statuses, I am wondering if you can place the 'canceled" text in the time column like we do if a trip is late or early. If you do use this approach, there might already be built in code for the screenreader message:
> 
> Image

I added the screenreader accommodations! I like the idea about adding content to the time column, but I'm a little worried about it not being as noticeable...I'll add a note to look more into that once we get some feedback from consumers.

packages/itinerary-body/src/styled.tsx
```diff
@@ -502,6 +503,7 @@ export const PlaceName = styled.span`
   overflow: hidden;
   white-space: nowrap;
   text-overflow: ellipsis;
+  text-decoration: ${props => (props.strikethrough ? "line-through" : "")};
   flex: 1 1 auto;
```
amy-corson-ibigroup: This flex attribute is what's causing the gap between the place and the "canceled". Can you conditionally adjust based on the `strikethrough` prop?

packages/itinerary-body/i18n/en-US.yml
```diff
@@ -69,6 +69,8 @@ otpUi:
       escooter: E-scooter
       vehicle: Vehicle
   ItineraryBody:
+    canceled: Canceled
+    canceledMessage: This trip has been canceled
```
amy-corson-ibigroup: Can we change this to "The following trip has been canceled" to better contextualize for screenreaders?

amy-corson-ibigroup (commented):
Just a couple things I'd like to see fixed before merge!

packages/itinerary-body/src/styled.tsx
alec-georgoff: Oh excellent catch, thank you! Fixed

packages/itinerary-body/i18n/en-US.yml
alec-georgoff: Yup!

packages/itinerary-body/src/ItineraryBody/place-row.tsx
```diff
@@ -12,6 +14,12 @@ import AccessibilityRating from "./accessibility-rating";
 import { PlaceNameProps, PlaceRowProps } from "../types";
 import { defaultMessages } from "../util";
 
+const CanceledTripMessage = styled.span`
+  color: ${colors.red[700]};
+  margin-left: 5px;
```
amy-corson-ibigroup: nitpicky but I think this can be upped to 7px

amy-corson-ibigroup (approved):
Thanks for the changes!

packages/itinerary-body/src/ItineraryBody/place-row.tsx
alec-georgoff: Will adjust!

PR #975, opened 2026-07-15
Add packageManager to package.json for corepack

alec-georgoff:
It appears that corepack requires a `packageManager` field in `package.json` to lock the version of pnpm that gets used. This gets auto-generated upon running `pnpm install`, but I had been reverting the change before pushing. I'm pretty sure this is what causes the [deployment pipeline to fail](https://github.com/opentripplanner/otp-ui/actions/runs/29433093010/job/87412793261) with a "working tree must be clean" message; the pipeline runs `pnpm install` and then this gets added and there's an uncommitted change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -294,5 +294,6 @@
   "dependencies": {
     "@storybook/icons": "^1.6.0",
     "@styled-icons/styled-icon": "catalog:"
-  }
+  },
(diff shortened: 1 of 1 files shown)


josh-willis-arcadis (approved):
lets see what happens!

amy-corson-ibigroup (approved):

PR #976, opened 2026-07-15
Update location-field geocoder

binh-dam-ibigroup:
Patch release for the location-field package so it uses the latest geocoder, and attempts to fix pnpm.lock from previous failed releases.

### DIFF
--- a/packages/location-field/src/index.tsx
+++ b/packages/location-field/src/index.tsx
@@ -392,6 +392,7 @@ const LocationField = ({
         setAbortController([...abortControllers, newController]);
 
         getGeocoder(geocoderConfig)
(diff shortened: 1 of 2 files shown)


alec-georgoff (approved):

daniel-heppner-ibigroup (approved):
I wonder what's going on with the weird pnpm lock change.

binh-dam-ibigroup:
@daniel-heppner-ibigroup and @alec-georgoff Unfortunately, I'm still getting a "Working tree is not clean" error. Running `pnpm semantic-release` locally (defaults to dry run), I see new .npmrc files. Is that intended? Is that the result of strict mode on the packages?

PR #977, opened 2026-07-17
Migrate from `plan` to `planConnection`

miles-grant-ibigroup:
Should be a pretty seamless transition on the otp-ui side of things.

- Removes now-uneeded qualifier support
- Removes a lot of combinatoric code, which should make implementation easier
- Attempts to be as backwards compatible as possible, re-using old variable names where possible
- [x] `banned` is still not working
- [x] There are some TS issues to iron out


packages/core-utils/src/planQuery.graphql
```diff
@@ -1,40 +1,31 @@
...
-    walkSpeed: $walkSpeed
-    wheelchair: $wheelchair
+    modes: $modes
+    origin: $origin
+    preferences: {accessibility: {wheelchair: {enabled: $wheelchair}}, street: {bicycle: {reluctance: $bikeReluctance}, car: {reluctance: $carReluctance}, walk: {speed: $walkSpeed, reluctance: $walkReluctance}}, transit: {filters: $transitFilter, timetable: {includeRealTimeCancellations: $includeRealTimeCancellations}}}
+    via: $via
   ) {
-    itineraries {
+    edges{
+    node {
```
daniel-heppner-ibigroup, line 28: indentation. I think there's a way to auto format this (maybe paste into graphiql and format there?)

packages/core-utils/src/query-gen.ts
```diff
@@ -199,13 +88,14 @@ export function generateCombinations(params: OTPQueryParams): OTPQueryParams[] {
...
-    .filter(combo => isCombinationValid(combo, queryTransitSubmodes))
-    .map(combo => ({ ...params, modes: combo }));
+  const transitModes = completeModeList
+    .filter(mode => isTransit(mode.mode) && mode.mode !== "TRANSIT")
+
+  // @ts-expect-error types packagge fail
+  return completeModeList
+    .filter(mode => !!mode.input)
+    // @ts-expect-error types packagge fail
+    .map(mode => ({ ...params, modes: { ...mode.input, transit: { ...mode?.input?.transit, ...transitModes.length > 0 && { transit: transitModes } } } }));
```
daniel-heppner-ibigroup, line 98: this ts-expect-error is because the ModeInput type is an or between the type you're expecting here and the { direct: string[] } type. I'll write some more thoughts on how to fix this in the types package comment

packages/core-utils/src/query-gen.ts
```diff
@@ -199,13 +88,14 @@ export function generateCombinations(params: OTPQueryParams): OTPQueryParams[] {
...
-    .filter(mode => TRANSIT_SUBMODES.includes(mode.mode))
-    .map(mode => mode.mode);
-
-  return combinations(completeModeList)
-    .filter(combo => isCombinationValid(combo, queryTransitSubmodes))
-    .map(combo => ({ ...params, modes: combo }));
+  const transitModes = completeModeList
+    .filter(mode => isTransit(mode.mode) && mode.mode !== "TRANSIT")
+
+  // @ts-expect-error types packagge fail
```
daniel-heppner-ibigroup, line 94: This ts-expect-error is just because of the below mode?.input?.transit issue. You can temporarily work around that by adding `as string[]` after `mode?.input?.transit` which reveals there's another typescript error:

the `mode` property on OTPQueryParams is expected to be an array, but in your map you're returning just an object. Looking at GraphiQL, I think this is supposed to be just an object. So maybe correct the OTPQueryParams type to only be a single object instead of an array. 

I think the reason you defined it as an array is because the same type is used as an argument to this function (`params`), so I think asking this type to be both the input and output type of this function is maybe asking it to be too much at once. How about adding a second argument `modes` which accepts an array of `TransportMode`, then we can just replace the `modes` on `params` with the generated combinations from the second argument.

packages/types/src/index.ts
```diff
@@ -834,12 +834,20 @@ export type ModeSettingValues = Record<string, number | string | boolean>;
...
 /**
  * TransportModes correspond with the OTP GraphQL TransportMode.
- * Could be anything from walk, bike (qualifier: rent) to transit, tram, or bus.
+ * Could be anything from walk, bike to transit, tram, or bus.
  */
-export type TransportMode = {
-  mode: string;
-  qualifier?: string;
-};
+export type TransportMode = { mode: string; input?: ModeInput };
```
daniel-heppner-ibigroup, line 839: I don't see anything in GraphiQL That matches up with this. The thing above `ModeInput` looks like this 

Image

packages/types/src/index.ts
```diff
@@ -834,12 +834,20 @@ export type ModeSettingValues = Record<string, number | string | boolean>;
...
+
+export type ModeInput =
```
daniel-heppner-ibigroup, line 841: This `ModeInput` type appears to be having an identity crisis. 
The first half is the same as `PlanTransitModesInput`, good.
The second half (which is causing the error in query-gen) matches up with the direct in `PlanModesInput` (see screenshot) which is the parent of PlanTransitModesInput. So there's a weird hierarchy thing going on

I think the best thing to do here is to split this up and try to make these types match the actual GraphQL shape a bit better, then we can try to fix the logic.

packages/core-utils/src/query-gen.ts
miles-grant-ibigroup: What's happening here is I added `input` as a custom field to make the otp-rr config easier. However, this now creates a situation where sometimes this method is fed a "clean" mode and sometimes it has this input field. I honestly don't know how best to model this without making the code needlessly more complex. I think this entire chunk of the code is due for a rewrite anyway once we feel PlanConnection is working and stable...

packages/types/src/index.ts
miles-grant-ibigroup: See my comment above... It's a bit messy, but I couldn't figure out anything cleaner

packages/types/src/index.ts
miles-grant-ibigroup: It matches the GraphQL shape, except for my one custom field

PR #978, opened 2026-07-17
Changesets parallel dry run poc

daniel-heppner-ibigroup:
This PR sets up changesets in otp-ui, running in parallel to semantic release so we can figure out how we like it.
- It requires every PR to have a changeset file. I think this will be a temporary thing just to basically force us to experiment with it.
- If you version a package with a changeset, after merging the PR, changeset bot will open another PR that does the actual package versioning. It will also check for dependent packages to the one you versioned and bump those version numbers too!
- The changeset PR created by the bot will be blocked from being merged while we have semantic release still around. This is done with a failing github action check


.github/workflows/changeset-check.yml
```diff
@@ -0,0 +1,35 @@
+# Validate that a changeset exists for this PR
+# Possibly remove this in the future, as enforcing a changeset is not necessarily good
```
alec-georgoff, line 2: In my previous company, we had a quick command to just add an empty changeset to satisfy this (we required the changeset to present as a pre-push hook, too). So I'm fine with keeping the requirement and allowing empty changesets rather than relying on habit/tribal knowledge

alec-georgoff (approved):
Exciting!

.github/workflows/changeset-check.yml
daniel-heppner-ibigroup: I am just going off of the changesets docs which recommended not using this gate (right before explaining how to do it if needed)

But yeah I kinda agree with you, I think if you make it easy to add an empty one then it's okay. I think you can just run pnpm changeset empty.

miles-grant-ibigroup (commented):
How does one differentiate between fix and feat?

daniel-heppner-ibigroup:
> How does one differentiate between fix and feat?

Try running pnpm changeset. You'll see the flow. You can choose patch, minor, and major. 

image

miles-grant-ibigroup (approved):
Excited for this post-discussion!

PR #979, opened 2026-07-24
Closed Stop Alert Map Popup

alec-georgoff:
Adds an annotation indicating a closed stop when the related entity has a `closed` property set to true.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -294,6 +294,5 @@
   "dependencies": {
     "@storybook/icons": "^1.6.0",
     "@styled-icons/styled-icon": "catalog:"
-  },
-  "packageManager": "pnpm@10.12.1+sha1.4bd977ce548dc9fcd1ca0c62838c55646cea6847"
+  }
 }
--- a/packages/map-popup/src/index.tsx
+++ b/packages/map-popup/src/index.tsx
@@ -39,15 +39,11 @@ export type Feed = {
 };
 
 const ClosedAlertWrapper = styled.div`
-  display: flex;
   align-items: center;
   color: ${colors.red[700]};
-
-  svg {
-    height: 15px;
-    width: 15px;
-    margin-right: 5px;
-  }
+  display: flex;
+  font-size: 14px;
+  gap: 5px;
 `
 
 const generateLocation = (entity: MapPopupEntity, name: string) => {
@@ -184,7 +180,7 @@ export function MapPopup({
 
       {stopClosureInfoAvailable && entity.closed &&
         <ClosedAlertWrapper>
-          <ExclamationTriangle />
+          <ExclamationTriangle size={15} />
           <FormattedMessage
             defaultMessage={defaultMessages["otpUi.MapPopup.closedStop"]}
             description="Text alerting user that the stop is closed"
--- a/packages/map-popup/src/index.tsx
+++ b/packages/map-popup/src/index.tsx
@@ -11,11 +11,13 @@ import type {
   StopEventHandler,
 } from "@opentripplanner/types";
 
-import { FocusTrapWrapper } from "@opentripplanner/building-blocks";
+import colors, { FocusTrapWrapper } from "@opentripplanner/building-blocks";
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -294,6 +294,5 @@
   "dependencies": {
     "@storybook/icons": "^1.6.0",
     "@styled-icons/styled-icon": "catalog:"
-  },
-  "packageManager": "pnpm@10.12.1+sha1.4bd977ce548dc9fcd1ca0c62838c55646cea6847"
```
alec-georgoff, line 298: This was added previously as an attempt to fix a deployment issue with semantic release. It didn't work, so removing it here.

packages/map-popup/src/index.tsx
```diff
@@ -168,6 +182,15 @@ export function MapPopup({
       {/* render dock info if it is available */}
       {entityIsStationHub && <StationHubDetails availableSpaces={entity.availableSpaces?.total} availableVehicles={entity.availableVehicles?.total} />}
 
+      {stopClosureInfoAvailable && entity.closed &&
```
josh-willis-arcadis, line 181: why do we need both checks here? aren't they just doing the same thing? Will just `entity.closed &&` not suffice?

josh-willis-arcadis (changes requested):
overall looks good. one question about rendering logic.

packages/map-popup/src/index.tsx
alec-georgoff: This is just a Typescript thing we have to do, unfortunately. The type of `entity` resolves to `Stop | VehicleRentalStation | RentalVehicle`, so we don't actually know if the `closed` property is on that object without doing some kind of check beforehand:

```ts
const stopClosureInfoAvailable = "closed" in entity;
```

If you try to just do `entity.closed`, TypeScript will yell at you

josh-willis-arcadis: ahhh I see. makes sense.

josh-willis-arcadis (approved):

packages/map-popup/src/index.tsx
```diff
@@ -36,6 +38,17 @@ export type Feed = {
   };
 };
 
+const ClosedAlertWrapper = styled.div`
```
amy-corson-ibigroup, line 41: Can we up the `font-size` here to 14? Right now this is defaulting to 12 between two larger fonts and it looks a little strange.

packages/map-popup/src/index.tsx
```diff
@@ -36,6 +38,17 @@ export type Feed = {
...
+  display: flex;
+  align-items: center;
+  color: ${colors.red[700]};
+
+  svg {
+    height: 15px;
+    width: 15px;
+    margin-right: 5px;
```
amy-corson-ibigroup: Nitpicky, but remove the `margin` here and add a `gap` to the parent `flex` element.

packages/map-popup/src/index.tsx
```diff
@@ -36,6 +38,17 @@ export type Feed = {
...
 };
 
+const ClosedAlertWrapper = styled.div`
+  display: flex;
+  align-items: center;
+  color: ${colors.red[700]};
+
+  svg {
+    height: 15px;
+    width: 15px;
```
amy-corson-ibigroup: you can pass a `size={15}` directly to the icon instead of setting it twice here

amy-corson-ibigroup (commented):
Could we look at fixing the vertical spacing between these elements? It's looking a little inconsistent:

Image

I'm also wondering if we want to having a larger conversation about canceled/closed elements? To me they feels similar enough cases that maybe we could see it as an opportunity to create some consistent styling between them?

alec-georgoff:
@amy-corson-ibigroup I implemented your suggestions and scheduled some time to go over design stuff!

amy-corson-ibigroup (approved):
Looks good! Thanks for the changes

PR #980, opened 2026-07-27
Fix canceled trip screenreader text

alec-georgoff:
Place the screenreader text for canceled trips behind the `canceled` boolean flag, as it should be


binh-dam-ibigroup (approved):
Thank you for the quick fix!

amy-corson-ibigroup (approved):
Can't believe I missed this the first time around. Good catch!

PR #981, opened 2026-07-27
Pass map layer fonts to transitive

amy-corson-ibigroup:
Using custom vector tiles can sometimes break the transitive overlay. This is because our custom `Layer` code requests certain fonts that are not always available in the tiles, resulting in an error. To provide a way to handle this, the followings changes are made to the following packages:

Base map:
- Map `ref` is exposed to the consumer

Transitive Overlay:
- Overlay now accepts the `vectorTileFonts<string[]>` as a prop. If users know which fonts are supported in their map layers, the can pass all the fonts at once to the overlay.

### DIFF
--- a/packages/base-map/src/index.tsx
+++ b/packages/base-map/src/index.tsx
@@ -41,6 +41,7 @@ type Props = React.ComponentPropsWithoutRef<React.ElementType> & {
   center?: [number, number];
   /** A unique identifier for the map (useful when using MapProvider) */
   id?: string;
+  /** Ref to pass so that map can be accessed by the consumer  */
   innerRef?: LegacyRef<MapRef>;
   /** An object of props which should be passed down to MapLibre */
   mapLibreProps?: MapProps;
--- a/packages/base-map/src/index.tsx
+++ b/packages/base-map/src/index.tsx
@@ -2,8 +2,13 @@ import maplibregl, {
   MapLayerMouseEvent,
   MapLayerTouchEvent
 } from "maplibre-gl";
-import React, { useCallback, useState } from "react";
-import { Map, MapProps, ViewStateChangeEvent } from "react-map-gl/maplibre";
+import React, { LegacyRef, useCallback, useState } from "react";
+import {
+  Map,
+  MapProps,
+  MapRef,
+  ViewStateChangeEvent
+} from "react-map-gl/maplibre";
 
 import { useIntl } from "react-intl";
 
@@ -36,6 +41,7 @@ type Props = React.ComponentPropsWithoutRef<React.ElementType> & {
   center?: [number, number];
   /** A unique identifier for the map (useful when using MapProvider) */
   id?: string;
+  innerRef?: LegacyRef<MapRef>;
   /** An object of props which should be passed down to MapLibre */
   mapLibreProps?: MapProps;
   /** The maximum zoom level the map should allow */
@@ -68,6 +74,7 @@ const BaseMap = ({
   center,
   children,
   id,
+  innerRef,
   mapLibreProps,
   maxZoom,
   onClick,
@@ -167,6 +174,7 @@ const BaseMap = ({
       onTouchCancel={clearLongPressTimer}
       onTouchEnd={clearLongPressTimer}
       onZoom={handleViewportChange}
+      ref={innerRef}
       style={style}
     >
       {(toggleableLayers.length > 0 ||
--- a/packages/transitive-overlay/src/index.tsx
+++ b/packages/transitive-overlay/src/index.tsx
@@ -109,6 +109,9 @@ const defaultTextPaintParams = {
   "text-halo-width": 2
 };
 
+const DEFAULT_REGULAR_FONTS = ["Open Sans Regular", "Arial Unicode MS Regular"];
+const DEFAULT_BOLD_FONTS = ["Open Sans Bold", "Arial Unicode MS Bold"];
+
 /**
  * Common text settings.
  */
@@ -122,10 +125,7 @@ const commonTextLayoutParams = (
     "text-justify": "auto",
     "text-radial-offset": 1,
     "text-size": 15,
-    "text-font": vectorTileFonts?.regular || [
-      "Open Sans Regular",
-      "Arial Unicode MS Regular"
-    ]
+    "text-font": vectorTileFonts?.regular || DEFAULT_REGULAR_FONTS
   };
 };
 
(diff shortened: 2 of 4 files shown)


packages/transitive-overlay/src/route-layers.ts
```diff
@@ -93,7 +94,10 @@ export function getRouteLayerLayout(
     ] as ExpressionSpecification,
     "text-allow-overlap": true,
     "text-field": ["get", textField],
-    "text-font": ["Open Sans Bold", "Arial Unicode MS Bold"],
+    "text-font":
+      boldFonts?.length && boldFonts.length > 0
+        ? boldFonts
+        : ["Open Sans Bold", "Arial Unicode MS Bold"],
```
miles-grant-ibigroup: Could this default not be initialized on line 58?

packages/transitive-overlay/src/index.tsx
```diff
@@ -187,6 +205,7 @@ const TransitiveCanvasOverlay = ({
   activeLeg,
   accessLegColorOverride,
   boundsFitting = true,
+  vectorTileFonts,
```
miles-grant-ibigroup:
```suggestion
  fontOverrides,
```

How's this?

packages/transitive-overlay/src/index.tsx
```diff
@@ -156,6 +172,8 @@ type Props = {
   activeLeg?: Leg;
   accessLegColorOverride?: string;
   boundsFitting?: boolean;
+  /* If applicable, pass a list of fonts supported by the custom base map layer */
+  vectorTileFonts?: string[];
```
miles-grant-ibigroup: Why not make this an object with `bold` and `regular`? That way you can remove a lot of the sketchy bold detection code?

miles-grant-ibigroup (changes requested):
The `ref` stuff is clean, the way we pass font info in could use some cleanup. If we pass in an object and make it clear that they're overrides I think that'll go a long way

packages/transitive-overlay/src/route-layers.ts
amy-corson-ibigroup: good call!

packages/transitive-overlay/src/util.ts
```diff
@@ -527,5 +534,33 @@ const drawArc = (straight: LineString) => {
...
+  // Once we have a list of unique fonts, sort them into bold and regular buckets
+  fonts.forEach(f => {
+    const font = f.toLowerCase();
+    if (font.includes("italic")) {
+      sortedFonts.italic.push(f);
+    } else if (font.includes("bold")) {
+      sortedFonts.bold.push(f);
+    } else {
+      sortedFonts.regular.push(f);
+    }
```
miles-grant-ibigroup: can we `switch` on `f.toLowerCase()`?

packages/transitive-overlay/src/index.tsx
```diff
@@ -106,13 +112,22 @@ const defaultTextPaintParams = {
...
-  "symbol-placement": "point",
-  "text-allow-overlap": false,
-  "text-field": ["get", "name"],
-  "text-justify": "auto",
-  "text-radial-offset": 1,
-  "text-size": 15
+const commonTextLayoutParams = (
+  vectorTileFonts?: SortedFonts
+): SymbolLayerSpecification["layout"] => {
+  const regularFonts = vectorTileFonts?.regular;
```
miles-grant-ibigroup: Can we inline this? Since it's only used once

miles-grant-ibigroup (approved):
Nice and clean! And should greatly simplify the otp-rr PR (if not remove it entirely)

Just a few small nitpicks!

packages/transitive-overlay/src/util.ts
amy-corson-ibigroup: Smart!!

packages/base-map/src/index.tsx
```diff
@@ -36,6 +41,7 @@ type Props = React.ComponentPropsWithoutRef<React.ElementType> & {
   center?: [number, number];
   /** A unique identifier for the map (useful when using MapProvider) */
   id?: string;
+  innerRef?: LegacyRef<MapRef>;
```
alec-georgoff, line 45: nit: Let's add a jsdoc comment on this to match the other props

packages/transitive-overlay/src/index.tsx
```diff
@@ -106,13 +112,21 @@ const defaultTextPaintParams = {
...
+  vectorTileFonts?: SortedFonts
+): SymbolLayerSpecification["layout"] => {
+  return {
+    "symbol-placement": "point",
+    "text-allow-overlap": false,
+    "text-field": ["get", "name"],
+    "text-justify": "auto",
+    "text-radial-offset": 1,
+    "text-size": 15,
+    "text-font": vectorTileFonts?.regular || [
```
alec-georgoff: nit: Can we extricate the default fonts into their own `DEFAULT_FONTS` const variable? Would make them easier to update in the future if desired.

packages/transitive-overlay/src/index.tsx
```diff
@@ -401,6 +434,11 @@ const TransitiveCanvasOverlay = ({
 
   const { fromAnchor, toAnchor } = getFromToAnchors(transitiveData);
 
+  const boldFonts =
+    vectorTileFonts?.bold.length && vectorTileFonts?.bold.length > 0
+      ? vectorTileFonts.bold
+      : ["Open Sans Bold", "Arial Unicode MS Bold"];
```
alec-georgoff: nit: Could also make these into a variable for easy updating!

packages/transitive-overlay/src/util.ts
```diff
@@ -527,5 +534,36 @@ const drawArc = (straight: LineString) => {
...
+  const sortedFonts: SortedFonts = {
+    bold: [],
+    regular: [],
+    italic: []
+  };
+
+  map.getStyle()?.layers?.forEach((l: any) => {
+    const fontLayers = l.layout && l.layout["text-font"];
+    // There are often duplicates so first filter those out.
+    fontLayers?.forEach((f: string) => !fonts.includes(f) && fonts.push(f));
```
alec-georgoff: I'd recommend using a `Set<string>` on 538 for `fonts`! Then you can just do:

```ts
fontLayers?.forEach((f: string) => fonts.add(f))
```

The set will take care of the deduplication, and it's more efficient because you don't have to iterate through an array a bunch of times.

amy-corson-ibigroup: Smart!

alec-georgoff (approved):

PR #983, opened 2026-07-31
Update dependency array to incorporate closedStops

alec-georgoff:
The lack of `closedStops` in the dependency array for a `useEffect` was causing the click handler to hold onto an `undefined` value for `closedStops` instead of the updated value once the query is completed.


amy-corson-ibigroup (approved):

daniel-heppner-ibigroup (approved):
use effect sadness

PR #984, opened 2026-08-10
fix: Location Field Aria Disabled Fix Logic 

nikkisato:
This is a follow up ticket to fix my mistake on https://github.com/opentripplanner/otp-ui/pull/948

Amy made a comment that menuItemCount would always be === 1 
https://github.com/opentripplanner/otp-ui/pull/948#issuecomment-4604157098

thought my solution would fix this since it was on mac that we found this issue 

But after testing I think this solution works

I also added a desktop story of suppressNearby since that what we are using on our side 

i also copied the accessibility overrides for the storybook from desktop to mobile was getting some error

The goal was to make sure that there was 
- no regression

- on location unavailable (it will still appear saying current location not available) 
image

- on SuppressNearby there is no open box  (please note this is the before the fix i just wanted to show what i mean by open box) 
image

### DIFF
--- a/packages/location-field/src/index.tsx
+++ b/packages/location-field/src/index.tsx
@@ -1000,7 +1000,7 @@ const LocationField = ({
   let optionTitle;
   let positionUnavailable;
 
-  if (currentPosition && !currentPosition.error) {
+  if (currentPosition?.coords && !currentPosition.error) {
     // current position detected successfully
     locationSelected = useCurrentLocation;
     optionIcon = currentPositionIcon;
@@ -1029,10 +1029,12 @@ const LocationField = ({
 
   const optionId = `current-position`;
 
-  // Add to the selection handler lookup (for use in onKeyDown)
-  pushToIndexedOptions(locationSelected, optionId);
-
   if (!suppressNearby) {
+    if (!positionUnavailable) {
+      // Add to the selection handler lookup (for use in onKeyDown)
+      pushToIndexedOptions(locationSelected, optionId);
+    }
+
     // Create and add the option item to the menu items array
     menuItems.push(
       <Option
@@ -1059,9 +1061,8 @@ const LocationField = ({
     }
     statusMessages.push(message);
   }
-
   // Store the number of location-associated items for reference in the onKeyDown method
-  let menuItemCount = indexedOptionLookup.length;
+  const menuItemCount = indexedOptionLookup.length;
 
   /** the text input element * */
   // Use this text for aria-label below if no user-provided label.
@@ -1072,8 +1073,19 @@ const LocationField = ({
     currentPosition && currentPosition.fetching
       ? intl.formatMessage({ id: "otpUi.LocationField.fetchingLocation" })
       : defaultPlaceholder;
-  const hasNoEnabledOptions = menuItemCount === 0;
-  const isExpanded = isStatic || menuVisible;
+
+  // Disable the toggle when the only menu item is "Current location not available".
+  // Disabled rows are not added to indexedOptionLookup, so the lookup is empty here.
+  const toggleDisabled =
+    menuItemCount === 0 &&
+    (menuItems.length === 0 ||
+      (positionUnavailable && !suppressNearby && menuItems.length === 1));
+
+  const isMenuOpen = isStatic || menuVisible;
+  const isExpanded = isMenuOpen && !toggleDisabled;
+  const ariaDisabled = toggleDisabled ? "true" : undefined;
+  const showMenuList = isStatic || menuItems.length > 0;
+  const listBoxControls = showMenuList ? listBoxId : undefined;
 
   const textControl = (
     <S.Input
@@ -1081,7 +1093,7 @@ const LocationField = ({
         activeIndex !== null ? indexedOptionLookup[activeIndex]?.id : null
       }
       aria-autocomplete="list"
-      aria-controls={listBoxId}
+      aria-controls={listBoxControls}
       aria-expanded={isExpanded}
       aria-haspopup="listbox"
       aria-invalid={!isValid}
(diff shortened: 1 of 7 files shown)


nikkisato:
Tagging miles here @miles-grant-ibigroup for review

packages/location-field/src/index.tsx
```diff
@@ -1029,10 +1029,12 @@ const LocationField = ({
...
 
-  // Add to the selection handler lookup (for use in onKeyDown)
-  pushToIndexedOptions(locationSelected, optionId);
-
   if (!suppressNearby) {
+    if (!positionUnavailable) {
+      // Add to the selection handler lookup (for use in onKeyDown)
+      pushToIndexedOptions(locationSelected, optionId);
+    }
+
```
miles-grant-ibigroup, line 1037: Doesn't this have the byproduct of position warnings never appearing when suppressNearby is active?

miles-grant-ibigroup (changes requested):
I'm a little weary of some of the performance implications of all these extra checks, as well as introducing react props into the CSS render. However, I'm not entirely sure what the desired outcome of this PR is. Maybe @amy-corson-ibigroup could help? 

If there is no other way to be a11y friendly I guess we could live with the performance hit, but it's a measurable 20-50ms in my testing which I think deserves a closer look

packages/location-field/src/index.tsx
nikkisato: /**
   * If true, do not show nearbyStops or current location as options
   */
  suppressNearby?: boolean;
  /**
  
  for the warnings there was only one that i was aware and it's still working 
  
  
Screenshot 2026-08-13 at 9 58 25 AM


Yes I believe that would be true but i'm not sure how to get a storybook with suppressNearby and location unavaliable will take a look into that

amy-corson-ibigroup:
Hello! Just flagging that I see this and I will review this week

packages/location-field/src/stories/__snapshots__/Desktop.story.tsx.snap
```diff
@@ -642,12 +632,57 @@ exports[`LocationField/Desktop Context SlowGeocoder smoke-test 1`] = `
...
       id="from-listbox"
       role="listbox"
-      class="styled__MenuItemList-sc-xahsq2-5 eCrJvo"
+      class="styled__MenuItemList-sc-xahsq2-5 fdpjmo"
   >
   </ul>
 </div>
 `;
 
+exports[`LocationField/Desktop Context SupressNearby smoke-test 1`] = `
```
amy-corson-ibigroup: "Suppress" spelled incorrectly here as well

packages/location-field/src/stories/Desktop.story.tsx
```diff
@@ -296,3 +296,15 @@ export const WithCustomLabel = (): JSX.Element => (
     />
   </>
 );
+
+export const SupressNearby = (): JSX.Element => (
```
amy-corson-ibigroup: nit: Suppress is spelled incorrectly here

amy-corson-ibigroup:
Hello, I'm really sorry for the delay! We've had devs on vacation, I had to take unexpected time off yesterday, and I am now at a client onsite, so I haven't given this the attention it deserves.

It looks like we're applying `aria-disabled` to the toggle when the toggle can open a menu with the "location unavailable" message. This doesn't feel like an accessibility improvement because sighted users now have access to information that AT users do not:

image

Would you also elaborate on the bug you're seeing or the situation you're encountering that `suppressNearby` prop seeks to solve? My original understanding of https://github.com/opentripplanner/otp-ui/pull/948 was that we want to disable an irrelevant toggle for AT users (which I agree with), but it feels like with the suppression of certain `menuItems` from the results list, the scope has changed and I think I'm struggling to understand why.

nikkisato:
@amy-corson-ibigroup 

I want to step back and describe the two issues we were trying to fix, because I think my earlier approach blurred them together.

You're right about aria-disabled: applying it when the menu still shows "location unavailable" isn't an accessibility win. Sighted users see that message; disabling the toggle can block screen reader users from the same info. I'll remove that.

Two issues we were fixing (I blurred them before):

#1 ) Toggle activates but nothing happens — aria-expanded stayed true with no selectable options.
#2) Icon lacks unavailable state — sighted users saw the message; assistive technology got no clear signal.
Updated approach (aligned with #948):

aria-expanded="false" when there are no selectable options → fixes #1
Unavailable state via aria-expanded="false" + role="status"; toggle stays activatable → fixes #2
aria-disabled="true" only when the menu is completely empty (irrelevant toggle)

On suppressNearby: Not original #948 scope — came up because TriMet uses it. We found the selectable lookup still counted a suppressed current-location entry. aria-disabled only applies there when the menu is literally empty.

Proposed changes:
- Don't push disabled "current location unavailable" into the selectable lookup
- aria-expanded="false" when there are no selectable options
- aria-disabled="true" only when the menu is completely empty — not when the unavailable message is showing
- Keep the toggle enabled for Location Unavailable so assistive technology users can open the menu; unavailable text via existing role="status"
- Keep $menuOpen CSS so the message stays visible when aria-expanded is false

nikkisato:
@miles-grant-ibigroup @amy-corson-ibigroup  
Ready for review!

packages/location-field/src/index.tsx
```diff
@@ -1059,33 +1062,71 @@ const LocationField = ({
...
+  // no other results). Informational rows such as "Current location not available"
+  // still count as menu content, so the toggle stays enabled in that case.
+  const toggleDisabled = menuItems.length === 0;
+  const ariaDisabled = toggleDisabled ? "true" : undefined;
+
+  // Avoid rendering an empty listbox shell when there are no menu rows.
+  const showMenuList = isStatic || menuItems.length > 0;
+  const listBoxControls = showMenuList ? listBoxId : undefined;
+
+  const statusAnnouncementText = statusMessages.filter(Boolean).join(". ");
```
miles-grant-ibigroup, line 1084: This causes problems in some languages (especially RTL languages). However, I'd be ok to merge this as-is if we add a TODO on here and the other line where we manually concatenate a string!

miles-grant-ibigroup (changes requested):
One quick comment, but otherwise I think things look good from my end! But I haven't checked through what @amy-corson-ibigroup mentioned

PR #987, opened 2026-09-03
Text Only Itinerary

amy-corson-ibigroup:
Creates a function that takes an `Itinerary` and returns a text-only itinerary string.

### DIFF
--- a/packages/itinerary-body/src/textOnlyItinerary/index.ts
+++ b/packages/itinerary-body/src/textOnlyItinerary/index.ts
@@ -1,0 +1,114 @@
+import coreUtils from "@opentripplanner/core-utils";
+import { Itinerary } from "@opentripplanner/types";
+import { IntlShape } from "react-intl";
+import { humanizeDistanceStringImperial } from "@opentripplanner/humanize-distance";
+import { getSummaryMode } from "../defaults/access-leg-description";
+
+const { isTransitLeg, getLegRouteName } = coreUtils.itinerary;
+const { toHoursMinutesSeconds } = coreUtils.time;
+
+const convertLegToTextString = (
+  leg: any,
+  intl: IntlShape,
+  isFirstLeg: boolean,
+  isLastLeg: boolean
+): any => {
+  const transitLeg = isTransitLeg(leg);
+  const textStrings: any = [];
+
+  if (isFirstLeg && !transitLeg) {
+    textStrings.push(
+      intl.formatMessage(
+        { id: "otpUi.TextOnlyItinerary.departFrom" },
+        { timeMillis: leg.startTime, place: leg.from.name }
+      )
+    );
+  }
+
+  if (transitLeg) {
+    const routeName =
+      getLegRouteName(leg) ||
+      // Idk why I'm having to do this, when it should be handled by the coreutils func >:(
+      leg.routeLongName ||
+      leg.routeShortName ||
+      leg.route;
+
+    textStrings.push(
+      // TODO: The stopCode needs to be conditional look at approxPrefix
+      intl.formatMessage(
+        { id: "otpUi.TextOnlyItinerary.transitDepartFrom" },
+        {
+          place: leg.from.name,
+          stopId: leg.from.stop.code,
+          timeMillis: leg.startTime,
+          routeName,
+          headsign: leg.headsign
+        }
+      )
+    );
+    textStrings.push(
+      intl.formatMessage(
+        { id: "otpUi.TransitLegBody.rideDurationAndStops" },
+        {
+          duration: intl.formatMessage(
+            { id: "otpUi.ItineraryBody.common.durationShort" },
+            {
+              ...toHoursMinutesSeconds(leg.duration),
+              approximatePrefix: undefined
+            }
+          ),
+          numStops: (leg.intermediateStops?.length || 0) + 1
+        }
+      )
+    );
+    // TODO: The stopCode needs to be conditional look at approxPrefix
+    textStrings.push(
+      intl.formatMessage(
+        { id: "otpUi.TextOnlyItinerary.transitArriveAt" },
+        {
+          timeMillis: leg.startTime,
+          place: leg.to.name,
+          stopId: leg.to.stop.code
+        }
+      )
+    );
+  }
+  if (!transitLeg) {
+    textStrings.push(
+      intl.formatMessage(
+        { id: "otpUi.AccessLegBody.summaryAndDistance" },
+        {
+          distance: humanizeDistanceStringImperial(leg.distance, false, intl),
+          mode: getSummaryMode(leg, intl),
+          place: leg.to.name
+        }
+      )
+    );
+  }
+  if (isLastLeg) {
+    textStrings.push(
+      intl.formatMessage(
+        { id: "otpUi.TextOnlyItinerary.arriveAt" },
+        { timeMillis: leg.startTime, place: leg.to.name }
+      )
+    );
+  }
+
+  return textStrings;
+};
+
+function textOnlyItineraryString(intl: IntlShape, itinerary: Itinerary): any {
+  const { legs } = itinerary;
+
+  return legs.map((leg, i, array) => {
+    const lastLeg = i === array.length - 1;
+    const firstLeg = i === 0;
+    return convertLegToTextString(leg, intl, firstLeg, lastLeg).map(
+      (string: string) => `
+            ${string}
+        `
+    );
+  });
+}
+
+export default textOnlyItineraryString;
--- a/packages/itinerary-body/i18n/en-US.yml
+++ b/packages/itinerary-body/i18n/en-US.yml
@@ -94,6 +94,15 @@ otpUi:
       likelyAccessible: likely accessible
       unclear: unknown
     viewOnMap: View on map
+  TextOnlyItinerary:
+    arriveAt: Arrive at {place} at {timeMillis, time, short}
+    departFrom: Depart from {place} at {timeMillis, time, short}
+    rideDurationAndStops: "Ride {route} for {duration}{numStops, plural, =1 {} other { / # stops}}"
+    routeDescription: "{routeName} to {headsign}"
+    transitArriveAt: "Arrive at {place} (Stop Id: {stopId}) at {timeMillis, time, short}"
+    transitDepartFrom: >-
+      Depart from {place} (Stop Id: {stopId}) at {timeMillis, time, short} on
+      {routeName} towards {headsign}
   TransitLegBody:
     AlertsBody:
       alertLinkText: See alert on {agency} website
--- a/packages/itinerary-body/src/ItineraryBody/index.tsx
+++ b/packages/itinerary-body/src/ItineraryBody/index.tsx
@@ -1,5 +1,6 @@
 import React, { ReactElement } from "react";
 
+import { Leg } from "@opentripplanner/types";
 import AccessibilityRating from "./accessibility-rating";
 import PlaceRow from "./place-row";
 import * as S from "../styled";
@@ -52,15 +53,15 @@ const ItineraryBody = ({
     trigger a rerender of this component itinerary is also another criteria
     that is used to trigger a rerender but has more reuse than companies here
   */
-  const rows = [];
+  const rows: ReactElement[] = [];
   let followsTransit = false;
-  let lastLeg;
-  let nextLeg;
+  let lastLeg: Leg | undefined;
+  let nextLeg: Leg | undefined;
   itinerary.legs.forEach((leg, i) => {
     const isLastLeg = i === itinerary.legs.length - 1;
     nextLeg = isLastLeg ? undefined : itinerary.legs[i + 1];
 
-    function createPlaceRow(isDestination) {
+    function createPlaceRow(isDestination: boolean) {
       // Create a row containing this leg's start place and leg traversal details
       rows.push(
         <PlaceRow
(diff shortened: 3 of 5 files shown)


packages/itinerary-body/src/AccessLegBody/rented-vehicle-subheader.tsx
```diff
@@ -80,6 +80,31 @@ function VehicleType({ type }: { type: string }) {
   }
 }
 
+export function vehicleTypeString(type: string, intl: IntlShape): string {
```
miles-grant-ibigroup: Is this not already in itinerary-body util.ts?

packages/itinerary-body/src/textOnlyItinerary/index.ts
```diff
@@ -0,0 +1,241 @@
...
+  config?: any
+): string => {
+  const intl = useIntl();
+  const transitLeg = isTransitLeg(leg);
+  const textStrings: any = [];
+
+  const isLastLeg = index === array.length - 1;
+  const isFirstLeg = index === 0;
+
+  const { units = "imperial" } = config;
```
miles-grant-ibigroup, line 31: Isn't config mandatory in other itinerary-body places? Can we keep it like that?

packages/itinerary-body/src/textOnlyItinerary/index.ts
```diff
@@ -0,0 +1,241 @@
...
+  getLegRouteName,
+  isFlex,
+  isReservationRequired
+} = coreUtils.itinerary;
+const { toHoursMinutesSeconds, ensureAtLeastOneMinute } = coreUtils.time;
+
+const convertLegToTextString = (
+  leg: Leg,
+  index: number,
+  array: Leg[],
```
miles-grant-ibigroup:
```suggestion
  allLegs: Leg[],
```

packages/itinerary-body/src/textOnlyItinerary/index.ts
```diff
@@ -0,0 +1,241 @@
...
+    // Walk to next leg, unless stop Ids are the same...
+    if (to.stopId !== from.stopId) {
+      textStrings.push(
+        intl.formatMessage(
+          { id: "otpUi.AccessLegBody.summaryAndDistance" },
+          {
+            distance:
+              units === "metric"
+                ? humanizeDistanceStringMetric(leg.distance, intl)
+                : humanizeDistanceStringImperial(leg.distance, false, intl),
```
miles-grant-ibigroup: `humanizeDistanceString` handles this for you

packages/itinerary-body/src/textOnlyItinerary/index.ts
```diff
@@ -0,0 +1,241 @@
...
+            mode: getSummaryMode(leg, intl),
+            place: getPlaceName(leg.to, [], intl)
+          }
+        )
+      );
+      // ...In which case we transfer
+    } else if (!leg.rideHailingEstimate) {
```
miles-grant-ibigroup: Why are we using rideHailingEstimate to determine if it's a transfer? Why not check for walking leg and then duration

packages/itinerary-body/src/textOnlyItinerary/index.ts
```diff
@@ -0,0 +1,241 @@
...
+  // DO NOT REFORMAT THIS, the template literals are adding line breaks between instructions and legs.
+  return textStrings.map((string: string, i: number) =>
+    i === 0 && !isFirstLeg
+      ? `
+  ${string}
+    `
+      : `${string}
+    `
+  );
+};
```
miles-grant-ibigroup, line 228: Why not `textStrings.join("\n")`?

packages/itinerary-body/src/TransitLegBody/index.tsx
```diff
@@ -75,26 +75,42 @@ const maximumAlertCountToShowUncollapsed = 2;
 
 /**
  * Helper function that assembles values for flex pickup/dropoff messages.
+ * To return text strings and not React fragments, pass outputString and intl
  */
-function getFlexMessageValues(info: FlexBookingInfo) {
+export function getFlexMessageValues(
+  info: FlexBookingInfo,
+  outputString?: boolean,
```
miles-grant-ibigroup: Instead of this can we always export a string and then create a new React function that uses the method?

packages/itinerary-body/src/textOnlyItinerary/index.ts
```diff
@@ -0,0 +1,241 @@
...
+  const textStrings: any = [];
+
+  const isLastLeg = index === array.length - 1;
+  const isFirstLeg = index === 0;
+
+  const { units = "imperial" } = config;
+
+  const { from, mode, rentedBike, to, duration } = leg;
+  const { name: fromName, networks, vertexType } = from;
+  const modeType = mode === "SCOOTER" ? "VEHICLERENTAL" : vertexType;
```
miles-grant-ibigroup, line 35: Why are we doing this? This seems brittle

packages/itinerary-body/src/textOnlyItinerary/index.ts
```diff
@@ -0,0 +1,241 @@
...
+    coreUtils.itinerary.getCompaniesLabelFromNetworks(
+      networks || [],
+      config?.companies
+    );
+  const vehicleName = isRental && leg.rentedCar && fromName ? fromName : "";
+
+  const interline = !isLastLeg && array[index + 1].interlineWithPreviousLeg;
+
+  // Flex header
+  if (isFlex(leg) && isReservationRequired(leg) && leg.pickupBookingInfo) {
```
miles-grant-ibigroup, line 51: These `if` and `if else` combined make this very difficult to read. Is there any way we can make it clean if statements only? If it's too much refactoring for right now I'm happy to make it a TODO as well. I believe switching to only using `if` should make the logic simpler

miles-grant-ibigroup (changes requested):
Getting rid of the `if else` is thing that I think will be most complicated to solve, so I'm alright with making it a TODO for now. But the other things I think we need to resolve before merging! There are many opportunities to remove duplicated code.

packages/itinerary-body/src/AccessLegBody/rented-vehicle-subheader.tsx
amy-corson-ibigroup: It is, `vehicleType()` just above this, but it returns component fragments instead of strings! It felt cleaner to make a new function but I'm open to refactoring.

packages/itinerary-body/src/textOnlyItinerary/index.ts
amy-corson-ibigroup: Sounds good to me!

packages/itinerary-body/src/textOnlyItinerary/index.ts
amy-corson-ibigroup: WOW

packages/itinerary-body/src/textOnlyItinerary/index.ts
amy-corson-ibigroup: Good catch, this was a holdover from trying to parse together a TNC text only itin!

packages/itinerary-body/src/textOnlyItinerary/index.ts
amy-corson-ibigroup: I pulled this almost directly from the `RentedVehicleSubheader`... I think it's because otp2 treats scooters as bikerentals? or at least thats's what I'm seeing in that file

packages/itinerary-body/src/textOnlyItinerary/index.ts
amy-corson-ibigroup: I left one, and it's the one where we check for transfers! But I tried to clean it up a bit to be more clear.

packages/itinerary-body/src/AccessLegBody/rented-vehicle-subheader.tsx
amy-corson-ibigroup: Oh wait I'm going to use your feedback from this comment and do it for both https://github.com/opentripplanner/otp-ui/pull/987#discussion_r4008636894

amy-corson-ibigroup: Okay I don't know what I was thinking; there's no reason these need to be fragments, so I just rewrote both the original functions

packages/itinerary-body/src/textOnlyItinerary/index.ts
```diff
@@ -0,0 +1,235 @@
...
+  config: any
+): string => {
+  const intl = useIntl();
+  const transitLeg = isTransitLeg(leg);
+  const textStrings: any = [];
+
+  const isLastLeg = index === allLegs.length - 1;
+  const isFirstLeg = index === 0;
+
+  const { units = "imperial" } = config;
```
miles-grant-ibigroup: I mean making config with units mandatory so that we don't have to define a default here!

miles-grant-ibigroup (changes requested):
This is much cleaner thanks! Only 2 nitpicks left are removing the default config and the `join`ing of the strings.

packages/itinerary-body/src/textOnlyItinerary/index.ts
amy-corson-ibigroup: Whoops! Sorry for not addressing this. `join` is certainly cleaner, my concern with it is that it makes it so we can't add further line breaks between legs. I do think breaking some of these longer itineraries into text blocks that are visually different is good so I'm not sure! What are your thoughts?

miles-grant-ibigroup: Can't you just do `\n`?

packages/itinerary-body/src/textOnlyItinerary/index.ts
amy-corson-ibigroup: Gotcha, fixed in [050ae14](https://github.com/opentripplanner/otp-ui/pull/987/commits/050ae1414427cd976a5d7c366e6c82139774c5b8)

packages/itinerary-body/src/textOnlyItinerary/index.ts
```diff
@@ -0,0 +1,235 @@
...
+  // DO NOT REFORMAT THIS, the template literals are adding line breaks between instructions and legs.
+  return textStrings.map((string: string, i: number) =>
+    i === 0 && !isFirstLeg
+      ? `
+  ${string}
+    `
+      : `${string}
+    `
+  );
+};
```
miles-grant-ibigroup, line 217:
```suggestion
  return textStrings.join("\n");
```

Does this really not work?

miles-grant-ibigroup (changes requested):

packages/itinerary-body/src/textOnlyItinerary/index.ts
amy-corson-ibigroup: It wasn't breaking up the legs the way I wanted, but I fixed it in [445ad22](https://github.com/opentripplanner/otp-ui/pull/987/commits/445ad22d581b6da019609a54db84c4e281068feb)

packages/itinerary-body/src/textOnlyItinerary/index.ts
```diff
@@ -0,0 +1,233 @@
...
+  if (isLastLeg) {
+    textStrings.push(
+      intl.formatMessage(
+        { id: "otpUi.TextOnlyItinerary.arriveAt" },
+        { timeMillis: leg.endTime, place: leg.to.name }
+      )
+    );
+  }
+
+  // DO NOT REFORMAT THIS, the template literals are adding line breaks between instructions and legs.
```
miles-grant-ibigroup: We can probably remove this comment now!

packages/itinerary-body/src/textOnlyItinerary/index.ts
```diff
@@ -0,0 +1,233 @@
...
+  return textStrings.join("\n");
+};
+
+function textOnlyItineraryString(itinerary: Itinerary, config: any): string {
+  const { legs } = itinerary;
+  return (
+    legs
+      .map((l, i, a) => convertLegToTextString(l, i, a, config))
+      // Create a linebreak between legs
+      .join("\n \n")
```
miles-grant-ibigroup:
```suggestion
      .join("\n\n")
```
Do we need the empty space?

miles-grant-ibigroup (approved):
Thanks for all your work on this! Just 2 small final cleanups, and then I think this should be good to go! (Once the types are resolved)

packages/itinerary-body/src/textOnlyItinerary/index.ts
```diff
@@ -0,0 +1,231 @@
...
+      intl.formatMessage(
+        { id: "otpUi.ItineraryBody.flexPickupMessage" },
+        { ...getFlexMessageValues(leg.pickupBookingInfo, intl) }
+      )
+    );
+  }
+
+  // Rental Micromobility Header
+  if (isRental) {
+    if (networks || rentedBike) {
```
alec-georgoff: nit: Combine these two `if`s: `if (isRental && (networks || rentedBike))`

packages/itinerary-body/src/textOnlyItinerary/index.ts
```diff
@@ -0,0 +1,231 @@
...
+        { timeMillis: leg.startTime, place: leg.from.name }
+      )
+    );
+  }
+
+  // Transit leg description
+  if (transitLeg) {
+    const routeName =
+      getLegRouteName(leg) ||
+      // Idk why I'm having to do this, when it should be handled by the coreutils func >:(
```
alec-georgoff: Should we maybe convert this to a TODO?

alec-georgoff (approved):
Two small nits but overall looks great! Nice job

packages/itinerary-body/src/textOnlyItinerary/index.ts
amy-corson-ibigroup: Nice catch!

packages/itinerary-body/src/textOnlyItinerary/index.ts
amy-corson-ibigroup: Whoops 100%

PR #988, opened 2026-09-03
Timetable trip specific notes

alec-georgoff:
Adds a `showNotices` flag prop that enables the timetable to show a hoverable icon in a new "Notices" column. When hovered, the notice for a trip will display.

For icons to show up, the timetable data fed to the component must include notices on trips. See google/transit#638 for more information on the notices field.

Screenshot 2026-09-11 at 11 43 14 AM

### DIFF
--- a/packages/timetable/src/index.tsx
+++ b/packages/timetable/src/index.tsx
@@ -3,7 +3,40 @@ import { IntlShape } from "react-intl";
 import styled from "styled-components";
 import toposort from "toposort";
 
+import colors from "@opentripplanner/building-blocks";
+
 const COLUMN_WIDTH = "85px";
+
+const NoticeSymbol = styled.span`
+  cursor: default;
+  border: solid black;
+  padding: 3px;
+  size: 15;
+`;
+
+const Notice = styled.div`
+  .hidden-child {
+    display: none;
+  }
+
+  :hover .hidden-child {
+    background-color: ${colors.grey[200]};
+    display: flex;
+    flex-direction: column;
+    position: absolute;
+  }
+`;
+
+const GenerateNotice = (content: string[]) => (
+  <Notice>
+    <NoticeSymbol>{"\u2139"}</NoticeSymbol>
+    <div className="hidden-child">
+      {content.map(s => (
+        <span key={s}>{s}</span>
+      ))}
+    </div>
+  </Notice>
+);
 
 const Table = styled.table`
   display: block;
@@ -42,6 +75,7 @@ interface TimetableTrip {
    * day. Used for sorting trips by first stop time
    */
   firstStopTime: number;
+  gtfsId: string;
   /** A map of stop GTFS ID to stop detail */
   stops: Map<string, StopDetail>;
 }
@@ -63,6 +97,7 @@ interface Pattern {
 
 interface Trip {
   blockId: string;
+  gtfsId: string;
   stoptimesForDate: Stoptime[];
 }
 
@@ -111,6 +146,7 @@ const createDwellStops = (trips: Trip[], timepoints: Set<string>): Trip[] => {
     });
     withDwellStops.push({
       blockId: trip.blockId,
+      gtfsId: trip.gtfsId,
       stoptimesForDate: updatedStopTimes
     });
   });
@@ -194,6 +230,8 @@ interface TimeTableProps {
   showBlockId?: boolean;
   /** Time zone in which to display stop times if no intl object is provided */
   timeZone?: string;
+  /** Text notes that should be shown as a hoverable popup on each individual trip in the timetable */
+  tripNotes?: Map<string, string[]>;
 }
 
 const TimeTable = (props: TimeTableProps): JSX.Element => {
@@ -206,7 +244,8 @@ const TimeTable = (props: TimeTableProps): JSX.Element => {
     route,
     showBlockId,
     timepointsOnly,
-    timeZone
+    timeZone,
+    tripNotes
   } = props;
 
   const { patterns } = route;
@@ -302,6 +341,7 @@ const TimeTable = (props: TimeTableProps): JSX.Element => {
         return {
           blockId: t.blockId,
           firstStopTime: firstStop.serviceDay + firstStop.scheduledArrival,
+          gtfsId: t.gtfsId,
           stops: new Map(
             t.stoptimesForDate.map(st => {
               return [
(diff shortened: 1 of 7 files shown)


packages/timetable/src/index.tsx
```diff
@@ -312,37 +355,54 @@ const TimeTable = (props: TimeTableProps): JSX.Element => {
...
-                  scope="col"
-                  closed={closedStops && closedStops.has(s.id)}
-                >
-                  {s.name}
-                </TH>
-              );
-            })}
+          {leadingColumns.concat(filteredPatternStops).map((s, index) => {
+            return (
+              <TH
```
josh-willis-arcadis, line 405: Do notices need their own column? It seems like overkill for an icon with an onHover, especially if not every row will have a notice. If we do, maybe we can simplify this column a bit?

packages/timetable/src/stories/Timetable.story.tsx
```diff
@@ -12,15 +12,18 @@ const meta = {
...
                 "Grand St NE & 29th Ave NE (Direction: 1)": "2:14634",
                 "46th St & I-35W (Direction: 0)": "2:53545"
             }
         },
         directionId: {
-            control: { type: "radio" },
+            control: "radio",
             options: [0, 1]
+        },
+        showNotices: {
```
josh-willis-arcadis, line 25: loving the storybook controls for review!

packages/timetable/src/index.tsx
```diff
@@ -3,8 +3,41 @@ import { IntlShape } from "react-intl";
 import styled from "styled-components";
 import toposort from "toposort";
 
+import colors from "@opentripplanner/building-blocks";
+
 const COLUMN_WIDTH = "85px";
 
+const NoticeSymbol = styled.span`
+  cursor: default;
+  border: solid black;
```
josh-willis-arcadis, line 11: I've seen this symbol more with a circle border rather than a square

josh-willis-arcadis (changes requested):
looking great so far. a few comments on the border style and notice column.

packages/timetable/src/index.tsx
alec-georgoff: Yeah that's fair. I went ahead and removed the column header for the notices column, which allowed us to greatly reduce the width. I think it does still need its own column though, since the leading column can change based on the props. So we wouldn't be able to just easily attach it to, say, the block ID.

packages/timetable/src/index.tsx
alec-georgoff: Switched to a circle!

packages/timetable/src/index.tsx
josh-willis-arcadis: This looks better. I still feel it could be a bit more sleek, but I feel this is fine for now. I'll leave to the second reviewer for any more thoughts or comments.

josh-willis-arcadis (approved):
Looks good. Any more UI refinements can be added in the future.

packages/timetable/src/index.tsx
amy-corson-ibigroup: I think this column needs an invisible header for screenreaders. Right now it's reading as an empty column with no details as to what the content is.

packages/timetable/src/index.tsx
```diff
@@ -3,12 +3,61 @@ import { IntlShape } from "react-intl";
...
+
+const Notice = styled.div`
+  display: flex;
+
+  .trip-notice-hover-content {
+    display: none;
+  }
+
+  :hover .trip-notice-hover-content {
+    background-color: ${colors.grey[200]};
```
amy-corson-ibigroup, line 27: 200 is looking a little dark would you be open to changing it to 100?

amy-corson-ibigroup (changes requested):
I'd really recommend against using an `onHover` here, since hover effects are almost impossible to use for screenreaders and anyone using a keyboard to navigate. 

Instead, I'd recommend turning the notice icon into a button toggle, that displays the details, which can be dismissed with `esc` keypress, clicking outside the container, or an `x` close button inside the container.

packages/timetable/src/index.tsx
alec-georgoff: For sure! Updated to 100

packages/timetable/src/index.tsx
alec-georgoff: Added an `aria-label` attribute to the empty notices column header: [cd6c388](https://github.com/opentripplanner/otp-ui/pull/988/commits/cd6c3880b1594797e875e793f724bc9878e8f1db)

alec-georgoff:
> I'd really recommend against using an `onHover` here, since hover effects are almost impossible to use for screenreaders and anyone using a keyboard to navigate.
> 
> Instead, I'd recommend turning the notice icon into a button toggle, that displays the details, which can be dismissed with `esc` keypress, clicking outside the container, or an `x` close button inside the container.

Makes sense! I went with the 'x' close button approach (you can also click on the notice icon again to close an open notice). Let me know what you think; happy to keep iterating on this behavior

packages/timetable/src/notice.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  const [isOpen, setIsOpen] = useState(false);
+
+  return (
+    <NoticeContainer>
+      <NoticeSymbol
+        className="trip-notice-symbol"
+        onClick={() => setIsOpen(!isOpen)}
+        role="button"
+      >
```
amy-corson-ibigroup, line 80: Why not just make this a button?

amy-corson-ibigroup: Also this will need an `aria-label` that says something like "open notices" when notices are closed and "close notices" when they're open

packages/timetable/src/notice.tsx
```diff
@@ -0,0 +1,68 @@
...
+        {"\u2139"}
+      </NoticeSymbol>
+      {isOpen ? (
+        <NoticeContent>
```
amy-corson-ibigroup, line 84: Can we add a TODO here that when notices open, the focus needs to be set to the first element in this container?

packages/timetable/src/notice.tsx
amy-corson-ibigroup, line 1: I think this component needs an `onKeyDown` handler where `esc` keypress sets `isOpen` to false.

packages/timetable/src/notice.tsx
```diff
@@ -0,0 +1,68 @@
...
+          <ul>
+            {content.map(s => (
+              <li key={s}>{s}</li>
+            ))}
+          </ul>
+          <CloseIcon onClick={() => setIsOpen(false)}>{"\u2715"}</CloseIcon>
```
amy-corson-ibigroup: This should also be a button and needs an aria label!

packages/timetable/src/index.tsx
```diff
@@ -194,6 +216,10 @@ interface TimeTableProps {
   showBlockId?: boolean;
   /** Time zone in which to display stop times if no intl object is provided */
   timeZone?: string;
+  /** Enable notices to be shown as a hoverable popup on each individual trip in the timetable. Requires
```
amy-corson-ibigroup: Might need to rewrite this comment to account for the new non-hover functionality

packages/timetable/src/index.tsx
amy-corson-ibigroup: Unfortunately screenreaders don't always handle `aria-label` consistently on non-interactive elements. The safer thing to do is to use an invisible text component. You can use the following CSS to hide the text from sighted users
```
  clip: rect(0, 0, 0, 0);
  height: 0;
  overflow: hidden;
  width: 0;
```

amy-corson-ibigroup (changes requested):
Thank you for the changes! Just a few cleanup items.

packages/timetable/src/notice.tsx
alec-georgoff: Updated: [7f0524f](https://github.com/opentripplanner/otp-ui/pull/988/commits/7f0524f05b71bcb5accdfde14f906f066b0f49e3)

packages/timetable/src/notice.tsx
alec-georgoff: Yup! [dbc15f6](https://github.com/opentripplanner/otp-ui/pull/988/commits/dbc15f61c8db9f698a809bd1c84aa3e6b56b0267)

packages/timetable/src/notice.tsx
alec-georgoff: Added: [d8fac74](https://github.com/opentripplanner/otp-ui/pull/988/commits/d8fac749fac0bc391b98b407b083e8475deda98d)

packages/timetable/src/notice.tsx
alec-georgoff: Updated: [c0f8f1a](https://github.com/opentripplanner/otp-ui/pull/988/commits/c0f8f1ab633efa999c2ff49dc36063fe65bbe15a)

packages/timetable/src/index.tsx
alec-georgoff: Good callout! [a32a87d](https://github.com/opentripplanner/otp-ui/pull/988/commits/a32a87d21d77dbe84c0a437a3e1cd6fcfc2931ae)

packages/timetable/src/index.tsx
alec-georgoff: Updated here: [8b86a3b](https://github.com/opentripplanner/otp-ui/pull/988/commits/8b86a3b0f6ca2c67ce05a3d2ae1eea6ec9295480)

Thanks for the accessibility expertise!!

amy-corson-ibigroup (approved):
Thank you for the changes!!

PR #989, opened 2026-09-10
Avoid crash on non-string font names

miles-grant-ibigroup:
#981 was very cool, but causes a WSOD on map layers with fonts that are specified as arrays (I had no idea that was possible...)

This fix is the cleanest I could come up with, but if someone has a cleaner idea then I'm all ears!

### DIFF
--- a/packages/transitive-overlay/src/util.ts
+++ b/packages/transitive-overlay/src/util.ts
@@ -549,17 +549,20 @@ const getFontsFromVectorTiles = (map: MapRef): SortedFonts => {
   });
   // Once we have a list of unique fonts, sort them into bold and regular buckets.
   fonts.forEach(f => {
-    const fontLowerCase = f.toLowerCase();
-    switch (true) {
-      case fontLowerCase.includes("italic"):
-        sortedFonts.italic.push(f);
-        break;
-      case fontLowerCase.includes("bold"):
-        sortedFonts.bold.push(f);
-        break;
-      default:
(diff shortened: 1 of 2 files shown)


packages/transitive-overlay/src/util.ts
amy-corson-ibigroup, line 548: would you want to add the check here instead, so `fontLayers` winds up being only the fonts with none of these random arrays?

miles-grant-ibigroup: I think the random arrays have an effect on the fonts

amy-corson-ibigroup (approved):

packages/transitive-overlay/src/util.ts
```diff
@@ -549,17 +549,20 @@ const getFontsFromVectorTiles = (map: MapRef): SortedFonts => {
...
-        break;
-      case fontLowerCase.includes("bold"):
-        sortedFonts.bold.push(f);
-        break;
-      default:
-        sortedFonts.regular.push(f);
+    if (typeof f === "string") {
+      // eslint-disable-next-line default-case
+      switch (true) {
+        case f?.toLowerCase().includes("italic"):
```
daniel-heppner-ibigroup: this is an insane use of a switch how have i not seen this before

would it really not be better as just two ifs?

daniel-heppner-ibigroup (approved):

